PR #37, opened 2019-07-24
Paulina

paulina-t:
Files that I have modified:

- read_gaw.py
- DMS.ipynb
- variables.ini


jgliss:
@paulina-t thanks for this PR: I tested the code and found some issues:

- Code failed since column name of vmrdms in (new file?) dms.dat is `DMS` and not `dimethylsulfide` as defined in `VAR_NAMES_FILE` (conflicts with other files containing DMS). I fixed that and renamed the variable name in the column in `dms.dat` from DMS to dimethylsulfide.
- Tests fail in https://github.com/metno/pyaerocom/blob/paulina/pyaerocom/test/test_dms_gaw.py

Also, I have some questions / comments:

- the new data files (dms.dat, msa.dat, so4.dat and blackcarbon.dat) are from Excel tables, right?
- the dms.dat and ams137s00.lsce.as.fl.dimethylsulfide.nl.da.dat contain largely the same data, right? We have to make a decision, which of those to use as default (it should not read both by default, which it currently does).
- The file names should be renamed (e.g. from dms.dat it is not clear that it is from Amsterdam Island).

I will try to look at it before merging, but if you have any comments or remarks, please let me know.

Cheers

paulina-t:
Hi Jonas,

It is strange that you still see the old name of the variable, DMS instead of dimethylsulfide since I asked Dirk (I have no write permission) to change it in the DMS.dat, and it worked for me. But it is good that you fixed it.

All the new files are generated from 2 Excel files provided by Dirk. DMS is duplicated, we talked about this a couple of times, and I would recommend using the original ams137s00.lsce.as.fl.dimethylsulfide.nl.da.dat file as default because I get the impression that this one has passed a quality control since it contains some NAN values where the excel file provides valid observations. It is also the file that I used in the examples. However, there is no metadata available for the Excel files, so I would ask Dirk before making any decision.

I agree that we should rename the files. I do not know what the naming convention is, but maybe you can use the name of the original excel file and add the name of the variable. This should not generate any conflict when running the code.

jgliss:
@paulina-t : I merged this pull request into v080DEV but related tests in `pyaerocom/test/test_dms_gaw.py` still fail. Will add an issue and would appreciate if you could check and give a hint, what is going wrong. Cheers!

PR #41, opened 2019-09-05
Augustin

AugustinMortier:
Use of aliases (e.g: check if 'sconcpm10' exists when asks for 'concpm10')


jgliss:
Hi, can you reverse the latest commit 5c6055d ? I already did this and pushed it into `v080DEV`

jgliss:
You can go back one commit via:

```
$ git reset --hard HEAD~1
```

PR #49, opened 2019-10-09
Altered SAFRICA and EUROPE in region ini file, not overlapping anymore

hannasv:
Need to agree on rectangular boundaries for NAFRICA and ASIA overlaps.

### DIFF
--- a/pyaerocom/config.py
+++ b/pyaerocom/config.py
@@ -159,6 +159,8 @@ class Config(object):
     SENTINEL5P_NAME = 'Sentinel5P'
     AEOLUS_NAME = 'AeolusL2A'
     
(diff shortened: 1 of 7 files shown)


jgliss:
Thanks @hannasv, I think this should not go into the current v080 release (which I will finalise today) since we have used these old regions throughout the whole year and in all websites. And I do not have time before the course to update and rerun the online pages so I think v0.8.0 should stick to the old definitions.

I also plan some minor updates related to the colocation routines in due time (i.e. within October), so maybe we could aim for a new release of v0.8.1 at the end of October, that has the updated regions and colocation and potentially a first BETA version of the land sea masks.

Maybe update the PR to branch v081dev (which I just created)

jgliss:
Here is the link to the new branch: https://github.com/metno/pyaerocom/tree/v081dev

PR #52, opened 2019-10-18
Updated the boundaries for SAFRICA and EUROPE regions. 

hannasv:
They are not overlapping anymore. Need to discuss NAFRICA and Asia because they overlap.


jgliss:
@hannasv is this PR still supposed to be merged? If yes, can you check the conflicts in config.py?

hannasv:
@jgliss, I fixed the merge conflict. 
I think it should be included since it removes overlapping square regions in Europe and NAfrica and NAfrica and SAfrica.

PR #73, opened 2019-12-03
Created test for filterclass

hannasv:
Currently testing:
1) Attributes (correct name and region)
2) Test examples of gridded, ungridded and colocated data objects.  

We should probably discuss other tests next week or revision of these. 

To other tests fail. Both in the test colocated.


jgliss:
This was merged manually into current working branch v090dev. Closing this issue now.

PR #115, opened 2020-05-20
updated ghost reader to handle new GHOST format

hansbrenna:


pyaerocom/io/read_ghost.py
```diff
@@ -121,11 +121,15 @@ class ReadGhost(ReadUngriddedBase):
     
     DATA_ID = 'GHOST.daily'
```
jgliss, line 122:
```suggestion
    DATA_ID = 'GHOST.EEA.daily'
```

PR #116, opened 2020-05-20
fix to ghost reader due to changed metadata keys

hansbrenna:
I had to update two metadata keys and change the metadata key used to automatically convert units. I also updated the list of invalid flags.


jgliss:
@hansbrenna great! Can you fix the failing test?

pyaerocom/io/test/test_ghost_meta_keys.py

pyaerocom/io/read_ghost.py
```diff
@@ -178,7 +178,7 @@ class ReadGhost(ReadUngriddedBase):
     #: Default flags used to invalidate data points (these may be either from
     #: provided flag or qa variable, or both, currently only from qa variable)
     DEFAULT_FLAGS_INVALID = {'qa' : np.asarray([0, 1, 2, 3, 6, 8, 9, 10, 12, 
-                                                13, 14, 17, 18, 22, 25, 30, 40, 
+                                                13, 14, 15, 18, 22, 25, 30, 40,
```
jgliss, line 181: I guess this is for a reason (removing 17 and using 15 instead)? :)

hansbrenna: From Dene's e-mail today:The default codes to remove are now: [0, 1, 2, 3, 6, 8, 9, 10, 12, 13, 14, 15, 18, 19, 22, 25, 30, 40, 41, 42]

hansbrenna:
> @hansbrenna great! Can you fix the failing test?
> 
> pyaerocom/io/test/test_ghost_meta_keys.py

How do I do that?

jgliss:
> > @hansbrenna great! Can you fix the failing test?
> > pyaerocom/io/test/test_ghost_meta_keys.py
> 
> How do I do that?

Looking at the test module, I guess you need to update the metadata keys in the list `desired_keys`. You can run the test module (it has a __main__ at the bottom that triggers the test). It should not raise an AssertionError.

hansbrenna:
I've updated the test and pushed it. What do I do now?

jgliss:
> I've updated the test and pushed it. What do I do now?

Hmm, it should have rerun the tests, but from the CI output it seems that was cancelled for some reason. I think it is okay to merge and in case the test continues to fail, we can fix it.

jgliss:
@hansbrenna before merging, I think you can do an empty commit via `git commit --allow-empty` from your branch, which should retrigger the CI checks, and hopefully they will pass then.

PR #126, opened 2020-05-28
Check gridded domain

jgliss:
I provided an option to check lat/lon domain of  `GriddedData` to avoid extraction of timeseries outside of grid-domain (relevant for analyses of non-global domains).

### DIFF
--- a/pyaerocom/helpers.py
+++ b/pyaerocom/helpers.py
@@ -84,7 +84,7 @@ def delete_all_coords_cube(cube, inplace=True):
 
 def extract_latlon_dataarray(arr, lat, lon, lat_dimname=None,
                              lon_dimname=None, method='nearest',
-                             new_index_name=None):
+                             new_index_name=None, check_domain=True):
     """Extract individual lat / lon coordinates from `DataArray`
     
     Parameters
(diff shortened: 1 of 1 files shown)


jgliss:
@hansbrenna please don't merge, this should not go into master yet, need to change branch :)

jgliss:
@hansbrenna now you can review and merge :)

pyaerocom/helpers.py
```diff
@@ -140,6 +132,13 @@ def extract_latlon_dataarray(arr, lat, lon, lat_dimname=None,
         lat = [lat]
     if isinstance(lon, str):
         lon = [lon]
+    if check_domain:
+        arr_lat = arr[lat_dimname].data
+        arr_lon = arr[lon_dimname].data
+        lat0, lat1 = arr_lat.min(), arr_lat.max()
+        lon0, lon1 = arr_lon.min(), arr_lon.max()
+        lat = [x for x in lat if lat0 <= x <= lat1]
+        lon = [x for x in lon if lon0 <= x <= lon1]
```
ejgal, line 141: I think this should be changed so that only pairs of latitudes and longitudes that exists gets added to lat and lon.
Currently this might lead to lat and lon having different lengths which causes problems in line 146.
Something like this (not tested):
`lat, lon = [x,y for x,y in zip(lat, lon) if (lat0 <= x <= lat1) and (lon0 <= y <= lon1) ]`

jgliss: @ejgal That is correct! Thanks for looking into this. Can you fix and send a new PR into v090dev?

ejgal: Yup, will fix.

jgliss: @ejgal I moved this into a new issue #127 since this PR is merged.

Issue #123, opened 2020-05-27
AerocomEvaluation with EMEP and GHOST colocates data outside of EMEP grid


ejgal:
When running the aerocom evaluation with EMEP data and GHOST.EEA.daily there are some matches outside of the EMEP domain, see pictures.

Evaluation interface:
![emep_ghost_web](https://user-images.githubusercontent.com/23613579/83000311-761dc100-a00a-11ea-8b50-865fcb520447.png)

EMEP grid:
![emep_grid](https://user-images.githubusercontent.com/23613579/83000338-7b7b0b80-a00a-11ea-8cd5-2d459342c032.png)

@hansbrenna , @jgliss: Are there some known issues I should be aware of here before looking into this?

jgliss:
Interesting... I have no idea why this is the case... @hansbrenna?

jgliss:
La Reunion is a French Island ;)

ejgal:
Oh, does the colocation routines check for more than matches in lat/lon and time, like country?
I guess this could look like it has matched Plateau Caillou (one of the stations on La Reunion) to some continental polluted region. I'm a bit confused..

![pm10_Plateau_Caillou](https://user-images.githubusercontent.com/23613579/83009687-6789d680-a017-11ea-9683-141325080892.png)

jgliss:
The colocation goes by lat / lon, so please check with EMEPers what is going on there. If these points show up, then I suspect that these coordinates are actually available in the EMEP model data.

jgliss:
Maybe you used a global run of EMEP?

ejgal:
Seems not, after reading in the EMEP data the latitude range is from 30.05 to 81.95, see below.
The units of the EMEP latitudes are units=Unit('degrees_north'), is that fine?
I tried colocating directly using colocate_gridded_ungridded with same results.

Reading the EMEP and GHOST data:
```
from pyaerocom.colocation import colocate_gridded_ungridded
from pyaerocom.io import ReadEMEP
from pyaerocom.io import ReadGhost
rghost = ReadGhost('GHOST.EEA.daily')
ghost = rghost.read('conco3')
filepath = '/lustre/storeB/project/fou/kl/emep/ModelRuns/2019_REPORTING/EMEP01_L20EC_rv4_
33.2018/Base_day.nc'
remep = ReadEMEP(filepath=filepath)
emep = remep.read_var('conco3', ts_type='daily')
```
Printing the latitudes of Plateu Caillou and the EMEP
```
[...]

PR #128, opened 2020-06-02
Ensure extract_latlon_dataarray handles lat/lon as pair

ejgal:
As discussed [here](https://github.com/metno/pyaerocom/pull/126/files#r433793518) extract_latlon_dataarray should check if both the latitude and longitude is contained in the dataarray to ensure that lat and lon have the same length.


pyaerocom/helpers.py
```diff
@@ -137,21 +137,28 @@ def extract_latlon_dataarray(arr, lat, lon, lat_dimname=None,
...
-        lon = [x for x in lon if lon0 <= x <= lon1]
+        new_lat = []
+        new_lon = []
+        for x, y in zip(lat, lon):
+            if (lat0 <= x <= lat1) and (lon0 <= y <= lon1):
+                new_lat.append(x)
+                new_lon.append(y)
+        if len(new_lat) == 0 and len(new_lon) == 0:
+            raise DataCoverageError('Coordinates not found in dataarray')
+        lat, lon = new_lat, new_lon
```
ejgal, line 148: A bit more explicit than what we discussed, but I had problems with the list comprehension solution when there were no matches between lat,lon and the coordinates in the input dataarray.

ejgal:
Guess we can hold off on merging this until we figure out the problems in #123.

ejgal:
Changed the naming of the lat/lon ranges.

pyaerocom/test/test_colocation.py
```diff
@@ -8,45 +8,70 @@
...
+
     means = [np.nanmean(coldata.data.data[0]),
              np.nanmean(coldata.data.data[1])]
-    
+
     npt.assert_allclose(means, [obsmean, modmean], rtol=TEST_RTOL)
-    
+
+@testdata_unavail
+def test_colocate_gridded_ungridded_nonglobal(aeronetsunv3lev2_subset):
```
jgliss, line 50: Consider simplifying initiation of dummy cube and GriddedData object and also to move iit outside of this method, e.g. into a fixture in conftest, that can then also be used by other tests.

jgliss:
@ejgal Feel free to merge if you consider this done!

pyaerocom/test/test_colocation.py
ejgal: Yes, created an issue for this. I would like to think a bit about what would be useful in general before implementing it.
#131.

ejgal:
Used filtering instead of manually selecting stations outside of lat/lon range and merged latest changes in v090dev. Should be ready to go.

ejgal:
@jgliss, merged the whitespace changes, should be ready to go.

PR #135, opened 2020-06-08
Colocation regrid latlon flex

jgliss:
Add option to specify regrid resolution lat/lon specific in lowlevel colocation routines.


pyaerocom/colocation.py
```diff
@@ -169,9 +172,15 @@ class can handle different methods with different inputs)
...
-                                                   scheme=regrid_scheme)
+        if not isinstance(regrid_res_deg, dict):
+            if not isnumeric(regrid_res_deg):
+                raise ValueError('Invalid input for regrid_res_deg. Need integer '
+                                 'or dict specifying lat and lon res')
+            regrid_res_deg = dict(lat_res_deg=regrid_res_deg,
+                                  lon_res_deg=regrid_res_deg)
+
+        gridded_data_ref = gridded_data_ref.regrid(scheme=regrid_scheme,
+                                                   **regrid_res_deg)
```
ejgal, line 183: This logic is repeated in `colocate_gridded_ungridded` (except from operating on gridded_data instead of gridded_data_ref). Move it into a function?

jgliss: Sounds good, will do that!

jgliss:
@ejgal moved regridding into a separate helper as you suggested. Also added a test for it, can you briefly review the latest updates? Cheers

jgliss:
@ejgal please let me know when you also checked the most recent updates (it says you approved the changes, but I am not sure, if that applies to the latest commits that I made).

Issue #130, opened 2020-06-05
Change in how observations/models are named for web processing


hansbrenna:
In CAMS61 there's a similarity between the variable specific obs data issue and the diurnal processing issue: The way observation data sets are named makes some things complicated. The name in the web interface now is derived directly from the key in the obs_config/model_config dictionaries. I see the potential for simplification if it was possible to optionally specify the web interface name (called `web_interface_name` below) inside each obs_config sub dict. This parameter would then be used when saving json files instead of the key in the dictionary. I have only really though about this for the observations, but it could be added for the models as well.

With support for dict specifiec obs_id, the setup for variable specific obs looks something like this:
```
'EEA-JP-1-5'        :   dict(obs_id={'concpm10' : 'GHOST.EEA.daily',
                                'concpm25' : 'GHOST.EEA.daily',
                                'concno2' : 'GHOST.EEA.daily',
                                'conco3' : 'GHOST.EEA.hourly'},
                             obs_vars=['concpm10','concpm25','concno2','conco3'],
                             obs_vert_type='Surface',
                             ignore_station_names={'concno2':no2_ignore_station_names,
                                                   'conco3':o3_ignore_station_names,
                                                   'concpm10':pm10_ignore_station_names,
                                                   'concpm25':pm25_ignore_station_names},
                             obs_filters={'set_flags_nan':True,
                                              'latitude':[30,70],'longitude':[-25,45],
                                              'Joly_Peuch_classification_code':[1,5],
                                              'ESDAC_Meybeck_landform_classification':landforms_to_include,}),
```
With this support I could turn this into (which is cleaner maybe?):
```
'EEA-JP-1-5'        :   dict(obs_id='GHOST.EEA.daily',
                             web_interface_name = 'EEA-JP-1-5'
                             obs_vars=['concpm10','concpm25','concno2'],
                             obs_vert_type='Surface',
                             ignore_station_names={'concno2':no2_ignore_station_names,
                                                   'conco3':o3_ignore_station_names,
                                                   'concpm10':pm10_ignore_station_names,
                                                   'concpm25':pm25_ignore_station_names},
                             obs_filters={filters}),

'EEA-JP-1-5_o3'        :   dict(obs_id='GHOST.EEA.hourly',
                             web_interface_name = 'EEA-JP-1-5'
                             obs_vars=['conco3'],
                             obs_vert_type='Surface',
                             ignore_station_names={'concno2':no2_ignore_station_names,
[...]

PR #139, opened 2020-06-10
initial implementation of web_interface_name option 

hansbrenna:
Initial implementation of functionality discussed in #130. 

One issue which requires some thinking now is that this change in the processing breaks the table under the informations pane in the web interface due to the fact that obs_ids and web_interface_names/obs_config-ids are no longer linked unambiguously.

### DIFF
--- a/pyaerocom/web/aerocom_evaluation.py
+++ b/pyaerocom/web/aerocom_evaluation.py
@@ -522,7 +522,10 @@ class AerocomEvaluation(object):
                                               model_name):
         """Creates all json files for one ColocatedData object"""
         vert_code = self.get_vert_code(obs_name, coldata.meta['var_name'][0])
-        web_iface_name = self.obs_config[obs_name]['web_interface_name']
+        try:
+            web_iface_name = self.obs_config[obs_name]['web_interface_name']
+        except:
+            web_iface_name = obs_name
         if len(self.region_groups) > 0:
             raise NotImplementedError('Filtering of grouped regions is not ready yet...')
         return compute_json_files_from_colocateddata(
@@ -899,10 +902,15 @@ class AerocomEvaluation(object):
                     pass
             except KeyError:
                 self.obs_config[obs_name]['web_interface_name'] = obs_name
-            assert isinstance(self.obs_config[obs_name]['web_interface_name'],str)
+            if not isinstance(self.obs_config[obs_name]['web_interface_name'],str):
+                raise ValueError('Invalid value for web_iface_name in {}. Need str type'.format(obs_name))
             iface_names.append(self.obs_config[obs_name]['web_interface_name'])
         iface_names = set(iface_names)
         return iface_names
+
+    @property
+    def iface_names(self):
+       return self._check_and_get_iface_names()
 
     def run_evaluation(self, model_name=None, obs_name=None, var_name=None,
                        update_interface=True,
@@ -967,8 +975,7 @@ class AerocomEvaluation(object):
             self.only_colocation = only_colocation
         if only_json is not None:
             self.only_json = only_json
-
-        self.iface_names = self._check_and_get_iface_names()
+        #self.iface_names = self._check_and_get_iface_names()
         if self.clear_existing_json:
             self.clean_json_files()
 
--- a/pyaerocom/web/aerocom_evaluation.py
+++ b/pyaerocom/web/aerocom_evaluation.py
@@ -522,6 +522,7 @@ class AerocomEvaluation(object):
                                               model_name):
         """Creates all json files for one ColocatedData object"""
         vert_code = self.get_vert_code(obs_name, coldata.meta['var_name'][0])
+        web_iface_name = self.obs_config[obs_name]['web_interface_name']
         if len(self.region_groups) > 0:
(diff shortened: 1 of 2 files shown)


pyaerocom/web/aerocom_evaluation.py
```diff
@@ -522,6 +522,7 @@ def compute_json_files_from_colocateddata(self, coldata, obs_name,
                                               model_name):
         """Creates all json files for one ColocatedData object"""
         vert_code = self.get_vert_code(obs_name, coldata.meta['var_name'][0])
+        web_iface_name = self.obs_config[obs_name]['web_interface_name']
```
jgliss, line 525: Consider adding try/except block in case `web_iface_name` is not defined, e.g.:
```python
try:
   web_iface_name = self.obs_config[obs_name]['web_interface_name']
except: 
   web_iface_name = obs_name
```
?

pyaerocom/web/aerocom_evaluation.py
```diff
@@ -1069,6 +1090,7 @@ def _info_from_map_file(filename):
 
     def get_web_overview_table(self):
         """Computes overview table based on existing map files"""
+        iface_names = self.iface_names
```
jgliss, line 1093: If I see it correctly this will fail if it is called without `run_evaluation` being called before. How about a property decorator method for iface_names:

```python
@property
def iface_names(self):
   return self._check_and_get_iface_names()
```

pyaerocom/web/aerocom_evaluation.py
```diff
@@ -948,6 +967,8 @@ class default will be used (defined in config file)
             self.only_colocation = only_colocation
         if only_json is not None:
             self.only_json = only_json
+
+        self.iface_names = self._check_and_get_iface_names()
```
jgliss, line 971: Will not be needed if you make this a property method (see comment below)

pyaerocom/web/aerocom_evaluation.py
```diff
@@ -885,6 +888,22 @@ def find_obs_matches(self, name_or_pattern):
                            .format(name_or_pattern))
         return matches
 
+    def _check_and_get_iface_names(self):
```
jgliss, line 891: I am not a fan of handling it this way. If I remember correctly `obs_config` entries are always instances of `ObsConfigEval`, correct? I think it could be safer and cleaner to handle this through a class method `get_web_iface_name` in `AerocomEvaluation`, e.g.:

```python
def get_web_iface_name(self, obs_name):
   try:
      name = self.obs_config[obs_name]['web_iface_name']
   except KeyError:
      name = obs_name
   if not isinstance(name, str):
      raise ValueError('Invalid value for web_iface_name in {}. Need str type'.format(name))
   return name

If you prefer to do it this way, then I suggest to make class attr. `iface_names` a property method, so that it is available on demand and does not require to be called by another method.
```

pyaerocom/web/aerocom_evaluation.py
```diff
@@ -885,6 +888,22 @@ def find_obs_matches(self, name_or_pattern):
...
+            try:
+                if self.obs_config[obs_name]['web_interface_name'] == None:
+                    self.obs_config[obs_name]['web_interface_name'] = obs_name
+                else:
+                    pass
+            except KeyError:
+                self.obs_config[obs_name]['web_interface_name'] = obs_name
+            assert isinstance(self.obs_config[obs_name]['web_interface_name'],str)
+            iface_names.append(self.obs_config[obs_name]['web_interface_name'])
+        iface_names = set(iface_names)
```
jgliss, line 904: Just curious: why making this a set?

pyaerocom/web/aerocom_evaluation.py
hansbrenna: Good point. I've added this.

pyaerocom/web/aerocom_evaluation.py
hansbrenna: I have actually never used this construct before. Very useful.

pyaerocom/web/aerocom_evaluation.py
hansbrenna: I'll remove.

pyaerocom/web/aerocom_evaluation.py
hansbrenna: I'm not immediately sure what other code changes would need to be done for this to work. Let me think about it.

pyaerocom/web/aerocom_evaluation.py
hansbrenna: I wanted to make sure there were no duplicates in the list. I'm not sure it's necessary for the way I've implemented this, but it seemed necessary at the time.

hansbrenna:
@jgliss I'm going to merge this soon, and I'll think about the change to
`def get_web_iface_name(self, obs_name):`
you suggested above.

jgliss:
> @jgliss I'm going to merge this soon, and I'll think about the change to
> `def get_web_iface_name(self, obs_name):`
> you suggested above.
Okay, I think that is okay for now, but on the longer run I would prefer to have a method that provides  access of `web_iface_name` for individual obs entries.

Issue #132, opened 2020-06-05
We need tests for colocation_auto.py


jgliss:
Tests for module `colocation_auto.py` are missing and urgently needed as there is several parallel developments going on at the moment. @hansbrenna and @ejgal, please keep that in mind. This module is the heart of the web evaluation tools which are currently also edited heavily and which lack any tests at the moment. The latter is on me and I am sorry for that, I will try to work on these tests a little here and there, but I appreciate any help on this. 

Tests should, if possible, rely on the content of [testdata-minimal](https://pyaerocom.met.no/pyaerocom-suppl/), so that they are executed in the automatic CI. If needed *testdata-minimal* may be extended by small, representative, subsets of your data.

The setup of the test environment is defined in [conftest.py](https://github.com/metno/pyaerocom/tree/v090dev) 

Creating dummy data is also an option and may be possible to use for testing certain things.

Let me know if you have any questions!

jgliss:
Both for this issue and also for #133 we will probably need a place where temporary data is stored (e.g. this may be colocated data NetCDF files created by Colocator class or json files created for web tools). Tests should then also check these files by reading them and testing if format, values, etc. are correct.

I think one option to handle temporary output created by tests is to use the provided [temporary directories from pytest](https://docs.pytest.org/en/latest/tmpdir.html). I haven't used that yet, but it looks useful and simple to use.

avaldebe:
the current version of `pyaerocom.colocation_auto` has a 82% coverage

PR #145, opened 2020-06-17
Tests for colocation_auto

ejgal:
working on #132

### DIFF
--- a/pyaerocom/test/test_colocation_auto.py
+++ b/pyaerocom/test/test_colocation_auto.py
@@ -1,0 +1,157 @@
+import os
+import pytest
+
+from pyaerocom.conftest import TESTDATADIR
+from pyaerocom.conftest import does_not_raise_exception
+from pyaerocom.conftest import testdata_unavail
+
+from pyaerocom import Colocator, ColocatedData, GriddedData, UngriddedData
+from pyaerocom.io import ReadGridded
+from pyaerocom.exceptions import DataCoverageError
+from pyaerocom.io.aux_read_cubes import add_cubes
+
+@pytest.fixture(scope='function')
+def col_tm5_aero(model_id='TM5-met2010_CTRL-TEST', obs_id='AeronetSunV3L2Subset.daily',
+                 obs_vars='od550aer'):
+    coloc = Colocator(model_id=model_id, obs_id=obs_id, obs_vars=obs_vars)
+    coloc.raise_exceptions = True
+    coloc.reanalyse_existing = True
+    coloc.start = 2010
+    return coloc
+
+@pytest.fixture(scope='function')
+def col():
+    return Colocator(raise_exceptions=True, reanalyze_existing=True)
+
+
+def test_colocator(col):
+    assert isinstance(col, Colocator)
+    col.obs_vars = 'var'
+    assert isinstance(col.obs_vars, str)
+
+def test_colocator_init_basedir_coldata(tmpdir):
+    basedir = os.path.join(tmpdir, 'basedir')
+    col = Colocator(raise_exceptions=True, basedir_coldata=basedir)
+    assert os.path.isdir(basedir)
+
+def test_colocator__check_add_model_read_aux():
+    coloc = Colocator(raise_exceptions=True)
+    r = ReadGridded('TM5-met2010_CTRL-TEST')
+
+    assert not coloc._check_add_model_read_aux('od550aer', r)
+
+    coloc.model_read_aux = {
+        'od550aer' : dict(
+            vars_required=['od550aer', 'od550aer'],
+            fun=add_cubes)}
+    assert coloc._check_add_model_read_aux('od550aer', r)
+
+@testdata_unavail
+def test_colocator__coldata_savename(data_tm5):
+    col = Colocator(raise_exceptions=True)
+    col.obs_name = 'obs'
+    col.model_name = 'model'
+    col.ts_type = 'monthly'
+    col.filter_name = 'WORLD'
+    savename = col._coldata_savename(data_tm5)
+    assert isinstance(savename, str)
+    assert savename == 'od550aer_REF-obs_MOD-model_20100101_20101231_monthly_WORLD.nc'
+
+def test_colocator_update(tmpdir):
+    col = Colocator(raise_exceptions=True)
+    col.update(test="test")
+    assert col.test == 'test'
+
+    obs_id = 'test'
+    col.update(obs_id=obs_id)
+    assert col.obs_id == obs_id
+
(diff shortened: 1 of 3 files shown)


ejgal:
@jgliss: Not sure if the fixtures I define here are the best, let me know if you have a better idea.
All tests from line 136 and down are for functions that are not run in the class or from any other place in pyaerocom as far as I can see.
Should I create an issue on that and trim the class down a bit, or are there plans for using those for something?

pyaerocom/test/test_colocation_auto.py
```diff
@@ -0,0 +1,181 @@
...
+from pyaerocom import Colocator, ColocatedData, GriddedData, UngriddedData
+from pyaerocom.io import ReadGridded
+from pyaerocom.exceptions import DataCoverageError
+from pyaerocom.io.aux_read_cubes import add_cubes
+
+
+@pytest.fixture(scope='module')
+def init_obsdata_paths():
+    import pyaerocom as pya
+    pya.const.add_data_search_dir(TESTDATADIR.joinpath('obsdata'))
```
jgliss, line 17: What is the advantage of having this as a fixture, instead of just adding this search directory in conftest?

pyaerocom/test/test_colocation_auto.py
```diff
@@ -0,0 +1,181 @@
...
+
+@pytest.fixture(scope='function')
+def col_tm5_aero(model_id='TM5-met2010_CTRL-TEST', obs_id='AeronetSunV3L2Subset.daily',
+                 obs_vars='od550aer'):
+    coloc = Colocator(model_id=model_id, obs_id=obs_id, obs_vars=obs_vars)
+    coloc.raise_exceptions = True
+    coloc.reanalyse_existing = True
+    coloc.start = 2010
+    return coloc
```
jgliss, line 26: Using `scope='function'` seems to be an interesting concept that I have not used before. I trust you that this makes sense, as I do not have the time right now to really understand why you do it this way.

ejgal: This is done since some of the tests might change the properties of coloc.
Should be okay since the fixture is not doing anything resource intensive.

pyaerocom/test/test_colocation_auto.py
ejgal: None that I can think of, will move it!

pyaerocom/test/test_colocation_auto.py
```diff
@@ -0,0 +1,181 @@
...
+    # gridded_ungridded
+    col_tm5_aero.run()
+    coldata = col_tm5_aero.data[col_tm5_aero.model_id][col_tm5_aero.obs_vars[0]]
+    assert isinstance(coldata, ColocatedData)
+
+    # gridded_gridded
+    col_tm5_aero.obs_id = col_tm5_aero.model_id
+    col_tm5_aero.run()
+    coldata = col_tm5_aero.data[col_tm5_aero.model_id][col_tm5_aero.obs_vars[0]]
+    assert isinstance(coldata, ColocatedData)
```
jgliss, line 90: 2 tests in 1? parametrize?

pyaerocom/test/test_colocation_auto.py
```diff
@@ -0,0 +1,181 @@
...
+    with pytest.raises(NotImplementedError):
+        col()
+
+def test_colocator__find_var_matches_OLD(): # Never called, delete?
+    pass
+
+def test_colocator_output_dir(): # Never called, delete?
+    col = Colocator(raise_exceptions=True)
+    with pytest.raises(AttributeError):
+        col.output_dir('task')
```
jgliss, line 177: Seems like method `output_dir` can indeed be deleted from Colocator class. Does not work anyways :)

pyaerocom/test/test_colocation_auto.py
```diff
@@ -0,0 +1,181 @@
...
+    data = col.read_model_data('od550aer')
+    assert isinstance(data, GriddedData)
+
+def test_colocator_call(): # Delete __call__?
+    col = Colocator(raise_exceptions=True)
+    with pytest.raises(NotImplementedError):
+        col()
+
+def test_colocator__find_var_matches_OLD(): # Never called, delete?
+    pass
```
jgliss, line 172: Feel free to remove this from Colocator class

pyaerocom/test/test_colocation_auto.py
```diff
@@ -0,0 +1,181 @@
...
+    col = Colocator(raise_exceptions=True)
+    model_id = 'TM5-met2010_CTRL-TEST'
+    col.model_id = model_id
+    data = col.read_model_data('od550aer')
+    assert isinstance(data, GriddedData)
+
+def test_colocator_call(): # Delete __call__?
+    col = Colocator(raise_exceptions=True)
+    with pytest.raises(NotImplementedError):
+        col()
```
jgliss, line 169: Maybe we want to re-introduce __call__ ?

pyaerocom/test/test_colocation_auto.py
```diff
@@ -0,0 +1,181 @@
...
+    data = col.read_ungridded([obs_var])
+    assert isinstance(data, UngriddedData)
+
+    col.read_opts_ungridded = None
+    col.obs_vars = ['od550aer']
+    with does_not_raise_exception():
+        data = col.read_ungridded()
+    col.obs_vars = ['invalid']
+    with pytest.raises(DataCoverageError):
+        data = col.read_ungridded()
```
jgliss, line 157: I would like to keep this method in `Colocator` class as I use it in some scripts for quick checks I think

pyaerocom/test/test_colocation_auto.py
jgliss, line 169: Maybe we want to re-introduce method `__call__` ?

pyaerocom/test/test_colocation_auto.py
```diff
@@ -0,0 +1,181 @@
...
+
+def test_colocator_read_model_data(): # Never called, delete?
+    col = Colocator(raise_exceptions=True)
+    model_id = 'TM5-met2010_CTRL-TEST'
+    col.model_id = model_id
+    data = col.read_model_data('od550aer')
+    assert isinstance(data, GriddedData)
```
jgliss, line 164: Please keep, need this method in some of my scripts.

jgliss: I mean the actual method in `Colocator` of course, not the test method :) But having a test for this is great as well!

ejgal:
Parametrized test_colocator_run and changed the obsdata fixture to just be a path in TEST_DATA.
Also removed the unnecessary functions and kept the others.

Issue #154, opened 2020-06-19
Standard on dealing with code that is commented out


ejgal:
There are some blocks of commented out code makes it harder to read the code and understand the flow. Most of these have no explanation as to why they are kept around.

My suggested way to deal with this is to delete any code that is commented and does not have an explanation. (At least before merging into the main branches)
Thoughts on this @hansbrenna, @jgliss 

Examples:
```
            if ts_type is None:
                # if colocation frequency is not specified
                ts_type = ts_type_src

            #check if co-located data file already exists before reading observational
            #data sets
            # if self.save_coldata:
            #     savename = self._coldata_savename(model_data, start, stop,
            #                                       ts_type, var_name=model_var)

            #     file_exists = self._check_coldata_exists(model_data.data_id,
            #                                              savename)

# =============================================================================
#             if not model_data.ts_type in all_ts_types:
#                 raise TemporalResolutionError('Invalid temporal resolution {} '
#                                               'in model {}'.format(model_data.ts_type,
#                                                                    self.model_id))
# =============================================================================
            ignore_stats = None
            if self.ignore_station_names is not None:
```

```
        for var in varlist_aerocom(self.obs_vars):
            if _oreader.var_supported(var):
                obs_vars.append(var)

        #flist(np.intersect1d(self.obs_vars, obs_vars_supported))

        if len(obs_vars) == 0:
```

jgliss:
Fine with me to just remove!

ejgal:
Okay, will do that from now on.

Issue #152, opened 2020-06-18
Add functionality for constrained gridded reading in web processing and colocation_auto


jgliss:
Check what is needed to be able to make use of the functionality implemented in #150 for automised web processing (i.e. specification in model_config or obs_config entries in `AerocomEvaluation`).

PR #155, opened 2020-06-19
Enable constrained gridded reading for observations in Colocator

jgliss:
This is just a quick fix to make a first version work. In principle, this should also be possible to be applied to the model data, both in gridded/gridded and gridded/ungridded colocation. Can thus, only be  considered a limited fix of #152.
Includes also some cleanup for #154.

### DIFF
--- a/pyaerocom/colocation_auto.py
+++ b/pyaerocom/colocation_auto.py
@@ -379,10 +379,6 @@ class Colocator(ColocationSetup):
         """Get the lowest resolution ts_type of input ts_types"""
         return get_lowest_resolution(ts_type, *ts_types)
 
-    def output_dir(self, task_name):
-        """Output directory for colocated data"""
-        return self._output_dirs[task_name]
-
     def _check_add_model_read_aux(self, model_var, model_reader):
         if not isinstance(self.model_read_aux, dict):
             return False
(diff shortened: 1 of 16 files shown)


pyaerocom/colocation_auto.py
```diff
@@ -948,7 +909,7 @@ def _run_gridded_gridded(self, var_name=None):
         data_objs = {}
 
         for model_var, obs_var in var_matches.items():
-
+            obs_filters_var = obs_filters[obs_var] if obs_var in obs_filters else {}
             print_log.info('Running {} / {} ({}, {})'.format(self.model_id,
```
ejgal, line 913: `obs_filters` might not be defined here.

pyaerocom/colocation_auto.py
```diff
@@ -920,22 +893,10 @@ def _run_gridded_gridded(self, var_name=None):
         obs_reader = ReadGridded(self.obs_id)
 
         if 'obs_filters' in self:
-            remaining_filters = self._eval_obs_filters()
-            if bool(remaining_filters):
-                raise NotImplementedError('Cannot apply filters {} to gridded '
-                                          'observation data.'.format(remaining_filters))
+            obs_filters = self._eval_obs_filters()
```
ejgal, line 896: Could do: `obs_filters = self._eval_obs_filters() if 'obs_filters' in self else {}` to ensure `obs_filters` is set.

jgliss: Done now, had some conflicts in colocation_auto.py after merging v090dev, and had to fix manually. Hope everything still works.

jgliss:
@ejgal can you check once more?

PR #159, opened 2020-06-23
Move site data colocation into new helper method 

jgliss:
Move code that colocates timeseries from 2 StationData objects out of main loop of `colocate_gridded_ungridded` method and into a small helper method. Also renamed some stuff and added a test.

Mostly done because I am working on a method to colocate 2 `UngriddedData` objects (which will probably make use of the new helper), which I need for some stuff in the paper. Also to improve readability and enable isolated tests of this functionality.

### DIFF
--- a/pyaerocom/colocation.py
+++ b/pyaerocom/colocation.py
@@ -362,6 +362,79 @@ def colocate_gridded_gridded(gridded_data, gridded_data_ref, ts_type=None,
                                   **kwargs)
     return data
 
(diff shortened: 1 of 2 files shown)


jgliss:
> This looks like a reasonable change to make the flow clearer and I think it's a good move towards having smaller functions/methods.
> 
> I could not see any obvious errors or bugs.

Thanks, @hansbrenna

PR #160, opened 2020-06-24
Manual caching ungriddeddata

jgliss:
These changes will enable to save (pickle) and reload an `UngriddedData`  object directly, using interface provided by `CacheHandlerUngridded`.

### DIFF
--- a/pyaerocom/conftest.py
+++ b/pyaerocom/conftest.py
@@ -139,8 +139,13 @@ def aasetal_data():
     return reader.read()  # read all variables
 
 @pytest.fixture(scope='session')
-def aeronetsunv3lev2_subset():
-    r = ReadAeronetSunV3('AeronetSunV3L2Subset.daily')
+def aeronet_sun_subset_reader():
+    reader = ReadAeronetSunV3('AeronetSunV3L2Subset.daily')
+    return reader
+
+@pytest.fixture(scope='session')
+def aeronetsunv3lev2_subset(aeronet_sun_subset_reader):
+    r = aeronet_sun_subset_reader
     #return r.read(vars_to_retrieve=TEST_VARS)
     return r.read(vars_to_retrieve=TEST_VARS_AERONET)
 
@@ -154,6 +159,12 @@ def loaded_nasa_ames_example():
     from pyaerocom.io.ebas_nasa_ames import EbasNasaAmesFile
     #fp = TESTDATADIR.joinpath(TEST_PATHS['nasa_ames_sc550aer'])
     return EbasNasaAmesFile(NASA_AMES_FILEPATHS['scatc_jfj'])
+
+@pytest.fixture(scope='session')
+def tempdir(tmpdir_factory):
+    """Temporary directory for dumping data shared between tests"""
+    tmpdir= tmpdir_factory.mktemp('data')
+    return tmpdir
 
 @contextmanager
 def does_not_raise_exception():
--- a/pyaerocom/io/cachehandler_ungridded.py
+++ b/pyaerocom/io/cachehandler_ungridded.py
@@ -112,7 +112,7 @@ class CacheHandlerUngridded(object):
         Parameters
         ----------
         var_name : str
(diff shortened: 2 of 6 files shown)


pyaerocom/io/cachehandler_ungridded.py
```diff
@@ -98,18 +98,57 @@ def dataset_to_read(self):
...
-        """File name of cache file"""
+    def default_file_name(self, var_name):
+        """File name of cache file
+
+
+        Parameters
+        ----------
+        var_name : str
+            name of variable to be cached. Note: if inpu
+
```
ejgal, line 116: The note is cut off.

pyaerocom/conftest.py
```diff
@@ -155,6 +160,12 @@ def loaded_nasa_ames_example():
     #fp = TESTDATADIR.joinpath(TEST_PATHS['nasa_ames_sc550aer'])
     return EbasNasaAmesFile(NASA_AMES_FILEPATHS['scatc_jfj'])
 
+@pytest.fixture(scope='session')
+def tempdir(tmpdir_factory):
+    """Temporary directory for dumping data shared between tests"""
+    tmpdir= tmpdir_factory.mktemp('data')
+    return tmpdir
```
ejgal, line 167: Maybe this should have scope='module' to make sure that tests from different modules using this fixture will interfere?

pyaerocom/io/cachehandler_ungridded.py
```diff
@@ -247,46 +296,59 @@ def delete_all_cache_files(self):
...
-                                  'not exist in input UngriddedData object'
-                                  .format(var_name))
-
-        if len(data.contains_vars) > 1:
-            data = data.extract_var(var_name)
-
-        fp = self.file_path(var_name)
+                data = data.extract_var(var_name)
+
+        fp = self.file_path(var_or_file_name, cache_dir=cache_dir)
```
ejgal, line 351: This part is a bit confusing after introduction of `var_or_file_name`.
Shouldn't the checks for number of datasets, variables and var_name run even if we call the function with a filename?

pyaerocom/io/cachehandler_ungridded.py
```diff
@@ -305,15 +367,14 @@ def write(self, data, var_name=None):
         finally:
             out_handle.close()
             if not success:
-                os.remove(self.file_path)
-        const.logger.info('Successfully wrote {} data ({}) to disk!'
-                    .format(var_name, self.reader.data_id))
+                os.remove(fp)
+        const.logger.info('Wrote: {}'.format(fp))
```
ejgal, line 371: +1 for short and concise log messages!

pyaerocom/conftest.py
jgliss: Actually, the idea is to allow them to look into the same temp directory.

pyaerocom/io/cachehandler_ungridded.py
jgliss: I agree it is a little confusing, but it is only important to check that for the normal caching (that requires access to the database), since the data in the database may have been changed in the meantime. The new case where also a filename can be provided directly is not supposed to check anything and is sort of just a workaround for storing *any* instance of an `UngriddedData`  object. Rethinking the whole thing, probably it would've been easier to not use `CacheHandlerUngridded` at all for this tast, and just `pickle.dump` directly `UngriddedData` objects directly... but let's leave it for now.

pyaerocom/conftest.py
ejgal: From different test files?

pyaerocom/io/cachehandler_ungridded.py
jgliss: :)

pyaerocom/io/cachehandler_ungridded.py
ejgal: I see, thanks for the clarification!

PR #168, opened 2020-07-08
Add option to fix invalid wavelengths in some EBAS files and minor cleanup

jgliss:


jgliss:
@ejgal, test_readgridded is failing here, I think it is related to the TM5 changes in testdata. Can you check (locally this test passes, as I am still having the former version of the testdataset). Cheers!

pyaerocom/io/read_ebas.py
```diff
@@ -1215,7 +1286,7 @@ def _read_files(self, files, vars_to_retrieve, files_contain, constraints):
         This method is not supposed to be called directly but is used in
         :func:`read` and serves the purpose of parallel loading of data
         """
-        data_obj = UngriddedData()
+        data_obj = UngriddedData(num_points=1000000)
```
ejgal, line 1290: Why is this initiated with this amount of points?

jgliss: Random choice, that was adapted from ReadGHOST. Was hoping this would speed up things a little, but did not test.

PR #172, opened 2020-07-16
Fix #171 and add some tests for units_helpers.py

### LINKED ISSUE #171: Unit conversion of some variables with custom Unit fails
E.g. the following does not work.

```python
import pyaerocom as pya
pya.units_helpers.convert_unit(1, 'ug m-3', 'g m-3', 'concso2')
```

This is because 'concso2' is sometimes reported in units that cannot be handled by `cf_units` module (e.g. ugS m-3) and therefore, for potentially affected variables a custom conversion was implemented in `pya.units_helpers` (see `UCONV_MUL_FACS` therein).

Should be be fixable with a simple try / except in method `convert_unit`.

jgliss:


pyaerocom/units_helpers.py
```diff
@@ -129,10 +127,22 @@ def convert_unit(data, from_unit, to_unit, var_name=None):
     data
         data in new unit
     """
+    if isinstance(from_unit, str):
+        from_unit = Unit(from_unit)
+    if isinstance(to_unit, str):
+        to_unit = Unit(to_unit)
+    if from_unit == to_unit:
+        # nothing to do
+        return data
```
ejgal, line 136: Good idea! Do you know of this is using Unit().definition for checking equality?

jgliss: Not sure how exactly equality is implemented there but something like Unit('1/m') == Unit('m-1') will be True

PR #180, opened 2020-08-13
Minor fixes 

jgliss:
- Update conversion of ts_type to numpy freq in `GriddedData.start` and `GriddedData.stop`
- Change access of data ID in ungridded reader classes (before it was done via `DATA_ID` attr., which is deprecated and now always returns the default data ID (should probably also be renamed to `DATA_ID_DEFAULT` or similar, at some point, to avoid confusion). Now access of current data ID is provided via attr. `data_id`
- Rename all occurrences of unit mole mole-1 to mol mol-1 in variables.ini (to be CF compliant)


pyaerocom/data/variables.ini
```diff
@@ -4744,9 +4744,9 @@ comments_and_purpose = Interpretation of Ozone budget; Oxidizing power of atmosp
 [lossco]
 var_name = lossco
 description = Rate of CO oxidation
-standard_name = tendency_of_atmosphere_mole_concentration_of_carbon_monoxide_due_to_chemical_destruction
+standard_name = tendency_of_atmosphere_mol_concentration_of_carbon_monoxide_due_to_chemical_destruction
```
hansbrenna, line 4747: Is this change right?

jgliss: Ah, nope!

jgliss:
@hansbrenna I hope now all is correct.

PR #183, opened 2020-08-24
implicit_to_explicit_rate

ejgal:
Add functionality for dealing with variables that should be defined as rates but does not have a frequency as part of its units (f.e. deposition rates in EMEP model data).

### DIFF
--- a/pyaerocom/helpers.py
+++ b/pyaerocom/helpers.py
@@ -411,7 +411,7 @@ def seconds_in_periods(timestamps, ts_type):
         timestamps = [to_pandas_timestamp(timestamp) for timestamp in timestamps]
     # From here on timestamps should be a numpy array containing pandas Timestamps
 
-    seconds_in_day = 24*60*60
+    seconds_in_day = 86400
     if ts_type >= TsType('monthly'):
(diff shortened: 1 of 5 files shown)


jgliss:
Probably won't be able to do anything on this this week.

pyaerocom/helpers.py
```diff
@@ -386,6 +386,51 @@ def infer_time_resolution(time_stamps):
...
+    """
+
+    ts_type = TsType(ts_type)
+    if isinstance(timestamps, np.datetime64):
+        timestamps = np.array([timestamps])
+    if isinstance(timestamps, np.ndarray):
+        timestamps = [to_pandas_timestamp(timestamp) for timestamp in timestamps]
+    # From here on timestamps should be a numpy array containing pandas Timestamps
+
+    seconds_in_day = 24*60*60
```
jgliss, line 414:
```suggestion
    seconds_in_day = 86400
```

jgliss (approved):

pyaerocom/helpers.py
jgliss: All good :+1:

PR #184, opened 2020-08-27
Provide logic for automatic EBAS file column selection based on variable unit

jgliss:
Needed to read, e.g. vmro3 from a NASA Ames file that contains both vmro3 and conco3, stored in different columns. 

Added `vmror` to ebas_config.ini. More variables for other gas mixing ratios (e.g. NO2, SO2) need to be added as needed.


jgliss:
@hansbrenna this should enable you to read vmr data from EBAS.

hansbrenna (approved):
I dont' see any obvious errors. If you merge it I can test it.

PR #185, opened 2020-08-31
Make .ts_types ignore unknown file names

### LINKED ISSUE #175: Available ts_types in ReadEMEP sometimes includes None
Listing available ts_types with `.ts_types` can include None values if there are files in data_dir that doesn't follow the EMEP naming convention.

This is probably due to `_get_ts_types()` using os.listdir when it should be using `_get_paths()`.
`_get_paths()` only list filenames matching the EMEP naming convention.

ejgal:
Fixes #175


jgliss (approved):
Looks good :+1:

PR #188, opened 2020-09-25
Fix to_datestring_YYYYMMDD and add tests

ejgal:
There was an issue with to_datestring_YYYYMMDD not working if input was
string.


jgliss (approved):
Good to go :)

Issue #194, opened 2020-10-05
Sparse test coverage for ReadGridded


ejgal:
There are not that many tests for ReadGridded and most rely on a lustre connection.
Some of the lustre tests could be converted to use the testdata instead.

There is a little confusion about the expected behavior of some functionality (see #191), which could be clarified through some tests.

jgliss:
@ejgal great initiative, thanks! I think the variable resolving logic in ReadGridded could indeed see some revision, it has been developed over quite some time and I agree that some method names are misleading, and likely some methods (like `read`) are buggy/outdated now. On the longer term, I think we should think about redesigning this issue of resolving variables. It is a thing in both ungridded and gridded reading and could be perhaps unified in a separate API that is used by ReadGridded, ReadEMEP, ReadUngridded, etc. Not for v0.10.0 but perhaps for v.1.0.0.

ejgal:
The API idea sounds interesting, will keep it in mind.
There are now at least some tests for ReadGridded not relying on lustre. Closing this for now.

PR #195, opened 2020-10-05
Add new tests for ReadGridded

### LINKED ISSUE #196: ReadGridded.get_var_info_from_files() fails if AUX_REQUIRES.keys() contain *
See `test_ReadGridded_get_var_info_from_files` in #195.
Caused by calling `check_compute_var(conc*`) which expects a variable name with no wildcards.

ejgal:
* Add tests for ReadGridded (#194)
* Fix #196


pyaerocom/io/test/test_readgridded.py
```diff
@@ -57,34 +58,120 @@ def test_ReadGridded_class_empty():
...
+    assert sorted(r.ts_types) == ['daily', 'monthly']
+
+@testdata_unavail
+def test_ReadGridded_read_var(reader_tm5):
+    r = reader_tm5
+    data = r.read_var('od550aer')
+    assert isinstance(data, GriddedData)
+    ds = data.to_xarray().load()
+    mean = float(ds.mean(dim='time').mean(dim='lat').mean(dim='lon'))
+    npt.assert_almost_equal(mean, 0.09631748)
```
jgliss, line 84: Why converting to xarray and doing mean over individual dimensions? You could probably also do data.mean() directly (which will be area weighted, though, by default, but should suffice).

pyaerocom/io/test/test_readgridded.py
```diff
@@ -57,34 +58,120 @@ def test_ReadGridded_class_empty():
...
+        assert r.has_var(var)  # calling has_var
+    assert sorted(r.vars_provided) == sorted(expected)
+    # assert r.has_var('conco3')
+
+
+def test_ReadGridded_prefer_longer():
+    r = ReadGridded(data_dir=path_tm5)
+    daily = r.read_var('abs550aer', ts_type='monthly', flex_ts_type=True,
+                       prefer_longer=True)
+    assert daily.ts_type == 'daily'
```
jgliss, line 126: Is there also a monthly field? Consider using parametrize with different settings?

jgliss (approved):
Good to go, but feel free to check my comments and add some updates.

pyaerocom/io/test/test_readgridded.py
ejgal: Yes, abs550aer has daily and monthly files in the testdata.
Running this with ts_type='yearly' and flex_ts_type=True results in `pyaerocom.exceptions.DataQueryError: Could not find file match for query`. Shouldn't flex_ts_type option ensure that it chooses the monthly or daily files?

pyaerocom/io/test/test_readgridded.py
ejgal: Forgot that GriddedData has that functionality! The weird way of taking means is caused by some 4D data I worked with recently :)

pyaerocom/io/test/test_readgridded.py
ejgal: Seems that the current behavior is that setting `ts_type` leads to `flex_ts_type` not affecting the behavior.

jgliss: Hmm, okay, can you make an issue for that?

jgliss:
@ejgal some tests are failing here. Can you check? Let me know if I should revise once more.

ejgal:
Seems that there was an error fetching the testdata. Reran the jobs without changing anything and they passed.

pyaerocom/io/test/test_readgridded.py
ejgal: Yep!

PR #198, opened 2020-10-07
Gliss paper modifications

jgliss:
Stuff I did during the revision of Gliss et al., 2020 paper. Some of it was done in a rush...


pyaerocom/colocateddata.py
```diff
@@ -1150,9 +1176,11 @@ def apply_latlon_filter(self, lat_range=None, lon_range=None,
...
             region_id = reg.name
 
-        if any(x is None for x in (lon_range, lat_range)):
-            raise ValueError('Need either lon_range or lat_range or valid '
-                             'region_id')
+# =============================================================================
+#         if any(x is None for x in (lon_range, lat_range)):
+#             raise ValueError('Need either lon_range or lat_range or valid '
+#                              'region_id')
+# =============================================================================
```
ejgal, line 1183:
```suggestion
```

pyaerocom/colocateddata.py
```diff
@@ -662,14 +680,22 @@ def calc_statistics(self, constrain_val_range=False,
...
-            weights = self.area_weights[0].flatten()
-        else:
-            weights = None
+        if use_area_weights and not 'weights' in kwargs and self.has_latlon_dims:
+# =============================================================================
+#                 raise DataDimensionError('Cannot calculate statistics with '
+#                                          'area weights. Data needs to have '
+#                                          'latitude / longitude dimension.')
+# =============================================================================
+
```
ejgal, line 689:
```suggestion
```

pyaerocom/colocateddata.py
```diff
@@ -350,13 +360,21 @@ def max(self):
...
         dims = self.data.dims
         if not 2 < len(dims) < 5:
             logger.info('Invalid number of dimensions. Must be 3 or 4')
             return False
         try:
-            if dims.index('data_source') == 0 and dims.index('time') == 1:
+            if dims.index('data_source') == 0:
+                if 'time' in dims and dims.index('time') == 1:
+                    return True
                 return True
```
ejgal, line 378: This will return true even if there is no index called time.

```suggestion
            if dims.index('data_source') == 0:
                if 'time' in dims and dims.index('time') == 1:
                    return True
```

ejgal: Or is that the intended use? Then the check for the time dimension here is redundant.

```suggestion
            if dims.index('data_source') == 0:
                return True
```

jgliss: I think this whole method should go out, I checked locally and it is not really needed I think. I'll remove it and push the update.

pyaerocom/colocation_auto.py
```diff
@@ -711,17 +766,15 @@ def _run_gridded_ungridded(self, var_name=None):
                     raise Exception(msg)
                 else:
                     continue
+
+            print(model_data.from_files)
+
```
ejgal, line 771: Left over from debugging?

```suggestion
```

pyaerocom/conftest.py
```diff
@@ -156,6 +161,12 @@ def loaded_nasa_ames_example():
     #fp = TESTDATADIR.joinpath(TEST_PATHS['nasa_ames_sc550aer'])
     return EbasNasaAmesFile(NASA_AMES_FILEPATHS['scatc_jfj'])
 
+@pytest.fixture(scope='session')
+def tempdir(tmpdir_factory):
+    """Temporary directory for dumping data shared between tests"""
+    tmpdir= tmpdir_factory.mktemp('data')
+    return tmpdir
```
ejgal, line 168: Is this meant to be persistent between different test files?
If not I think this should have scope='module'.

jgliss: Yes it is (I think the idea was to reload files created in a different module, such as colocated data objects).

pyaerocom/test/test_colocation_auto.py
```diff
@@ -112,12 +112,8 @@ def test_colocator_read_ungridded():
...
     col.obs_id = obs_id
     col.read_opts_ungridded = {'last_file' : 10}
-    with pytest.raises(DataCoverageError):
-        data = col.read_ungridded(obs_var) # Why is this raised?
-        # read_ungridded expects a list of one or more variables!
-        # Should check for str and convert to list?
-        # Should read_ungridded and read_model_data have same arguments?
-    data = col.read_ungridded([obs_var])
+
+    data = col.read_ungridded(obs_var)
```
ejgal, line 116: Thanks for cleaning that up!

ejgal (approved):
Lots of good clean up, refactoring and added functionality!
I'm not too familiar with the details of the plotting and calculation routines, but it looks reasonable.

jgliss:
> Lots of good clean up, refactoring and added functionality!
> I'm not too familiar with the details of the plotting and calculation routines, but it looks reasonable.

Thanks for reviewing!!

Issue #173, opened 2020-07-17
Add property years_avail to ReadEMEP


ejgal:
To match recent changes in colocation_auto (#170) property `years_avail` should be added to ReadEMEP.
Look to  ReadGridded for functionality.

PR #205, opened 2020-10-09
Add property years_avail to ReadEMEP

ejgal:
Add property to match changes in colocation routines (#173)

### DIFF
--- a/pyaerocom/io/read_emep.py
+++ b/pyaerocom/io/read_emep.py
@@ -7,6 +7,7 @@ Created on Mon Feb 10 13:20:04 2020
 """
 
 import xarray as xr
+import numpy as np
 import sys
 import os
 import glob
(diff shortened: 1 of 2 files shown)


ejgal:
Should give correct result, since all files in data_dir should be from the same year.
@jgliss, do you have a better idea for this?

pyaerocom/io/read_emep.py
```diff
@@ -121,6 +122,21 @@ def ts_types(self):
...
+        Years available in dataset
+        """
+        try:
+            paths = self._get_paths()
+            data = xr.open_dataset(paths[0])
+            years = data.time.dt.year.values
+            years = list(np.unique(years))
+            return sorted(years)
+        except Exception as e:
+            return []
```
jgliss, line 137: Seems okay and is probably not very critical for processing, is it?

jgliss (approved):
I think that's okay! For AeroCom reading it is more important to have `years_available` in the reader, as files are single year, and for computation of auxiliary variables the correct file matches have to be found. In case of EMEP, all variables are stored in a single netcdf file, and thereby share the same time dimension, correct?

ejgal:
Yes, that's the idea. This was done to work with _infer_start_stop from colocation_auto, so should only be accessed if start is not provided.

PR #207, opened 2020-10-14
Improve logic in surface layer extraction method in GriddedData

jgliss:


pyaerocom/griddeddata.py
```diff
@@ -1305,7 +1305,14 @@ def _infer_index_surface_level(self):
         except DimensionOrderError:
             self.reorder_dimensions_tseries()
         cname = self.dimcoord_names[-1]
+        coord = self[cname]
         from pyaerocom import vert_coords as vc
+        if 'positive' in coord.attributes:
+            if coord.attributes['positive'] == 'up':
```
hansbrenna, line 1311: I'm not sure I understand the logic here.

hansbrenna (commented):
I didn't completely understand the new logic you added

pyaerocom/griddeddata.py
hansbrenna: i think I got it now.

jgliss: This relies on the assumption that the entries of the 1D lev coordinate are representative, i.e. if it is pressure levels, then `positive` should be `down`, if it is altitudes above sea level then `positive` should be `up`. However, the order of the indices in the model output can be both, sometimes the first index corresponds to surface, but for pressure coordinates, often the first index in the array is ToA.

jgliss: But these things always confuse me as well...

jgliss: http://cfconventions.org/Data/cf-conventions/cf-conventions-1.8/cf-conventions.html#vertical-coordinate

hansbrenna: I see. They can be ordered both ways (in principle) so you want to first infer if the bottom level has the highest or lowest array value by using "up" or "down". Then you want to find the corresponding index for the highest/lowest value.

jgliss: Correct

PR #217, opened 2020-10-29
Tests for config.py module

jgliss:


hansbrenna (approved):
I think this is fine. I could see no obvious bugs and the tests are all passing. 

Taking the testdata access out into it's own class is a very good idea. Much easier to read and understand.

PR #223, opened 2020-11-02
Add aeronet read numobs

jgliss:
AERONET daily average files contain information about the number of observations for each data column. These information should be imported when reading the files, so that filtering by minimum no. of measurements per day can be applied.

It is a little cumbersome to implement at the moment, at least for auxiliary variables such as `od550aer`, which is computed from measurements of `od500aer` and `od440aer` (where the former channel contains NaNs) and using the provided `ang4487aer` data, and the respective numobs data would need to be combined during reading and then written into.


jgliss:
I still think this will be useful to have, so I keep this open. However, ultimately having the number of observations that go into a certain datapoint (e.g. a daily average) should be handled in pyaerocom in a more general way. That is, it would be useful also for other datasets (e.g. EBAS), and could also be stored, e.g. when temporal resampling is done (e.g. from daily to monthly).

jgriesfeller:
just a comment: in the aerocom-tools we had that number available, but it was never used

jgliss:
there is an issue for that and implementing this in the future should start from scratch, so I close this and delete the branch.

PR #227, opened 2020-11-04
Edit tutorials

jgliss:


jgliss:
@ejgal could you quickly check before I merge this?

PR #232, opened 2020-11-10
Dev mod mod coloc

hansbrenna:
Further development of model-model colocation.


jgliss:
@hansbrenna what about this branch?

jgliss:
@hansbrenna what about this PR? Will you be able to finish this for release of 0.12.0?

hansbrenna:
I started this when I thought I would need it for CAMS61. I ended up not needing it and I don't think I will need it in the near future. If this is a prioritized feature I should be able to implement it (I think)

jgliss:
Can you check if there is anything that you already did, that should be digested in main-dev? If not, feel free to close this PR and perhaps create an issue, if mod / mod co-location still does not work.

hansbrenna:
@avaldebe no.

PR #233, opened 2020-11-10
Dev gridded conc from vmr

hansbrenna:
Development of recursive variable calculation for gridded data


hansbrenna:
Implemented quick fix changes to allow direct vmr* -> conc* calculations using STP see #250.

Need to set the right standard temperature. Currently 293 K.

Issue #243, opened 2020-11-18
Calculating ammonia and nitric acid variables


ejgal:
Reading the variables Sum(HNO3+NO3)  and Sum(NH3+NH4) from EBAS works fine after defining them with units `ug N m-3` in variables.ini.

On the model side we have the following relevant variables:
Sum(HNO3+NO3)
* concno3c
* concno3f
* vmrhno3
concno3c and concno3f could be combined with add_cubes and AUX_REQUIRES in ReadGridded.

Sum(NH3+NH4)
* concnh4
* vmrnh3

Can mathutils.vmrx_to_concx be used for converting vmrnh3 and vmrhno3?

In the end we want everything in units of ug N m-3.

jgliss:
@hansbrenna can you comment on this? Eirik is checking units etc., and obsreading of the N related variables. Is there an easy way to convert vmrhno3 to conchno3? I remember you also implemented some more flexibility in ReadGridded for automatic conversion of vmrX to concX, given that P and T fields are provided. 

In the ideal case, this could be handled recursively via AUX_REQUIRES etc in ReadGridded.

ejgal:
I checked these variables for the following models (from here: https://gitlab.met.no/aerocom-evaluation/cams61_json_p2/-/blob/master/2018-rerun4/cfg_cams61_p2_2018-rerun4.json)
```
model_ids = ['EMEP.cams61.rerun',
'EMEP.cams50.u3all', 'EMEP.cams50.u3da', 'SILAM.cams61.rerun',
'MONARCH.cams61.rerun','EURAD-IM.cams61.rerun', 'MINNI.cams61.rerun',
'DEHM.cams61.rerun', 'MATCH.cams61.rerun', 'CHIMERE.cams61.rerun',
'LOTOSEUROS.cams61.rerun', 'CAMSRA.cams61.rerun']
```
Are these the right models @hansbrenna?

It showed the following variables are missing:
```
EMEP.cams50.u3all: ['ts', 'ps']
EMEP.cams50.u3da: ['ts', 'ps']
MONARCH.cams61.rerun: ['concno3c']
EURAD-IM.cams61.rerun: ['concno3c']
MINNI.cams61.rerun: ['concno3c']
MATCH.cams61.rerun: ['ts', 'ps', 'concno3c', 'concno3f', 'vmrhno3', 'concnh4', 'vmrnh3']
CHIMERE.cams61.rerun: ['ts']
[...]

PR #244, opened 2020-11-18
Reading of ammonia and nitric acid variables

ejgal:
Related to #243

### DIFF
--- a/pyaerocom/griddeddata.py
+++ b/pyaerocom/griddeddata.py
@@ -3,6 +3,7 @@
 from collections import OrderedDict as od
 
 import os
+from pyaerocom import units_helpers
 
 import iris
 from iris.analysis.cartography import area_weights
@@ -684,12 +685,18 @@ class GriddedData(object):
 
     def convert_unit(self, new_unit):
         """Convert unit of data to new unit"""
-# =============================================================================
-#         if self._size_GB > self._MAX_SIZE_GB:
-#             raise MemoryError('Cannot convert unit in {} since data is too '
-#                               'large ({} GB)'.format(self.name, self._size_GB))
-# =============================================================================
-        self.grid.convert_units(new_unit)
+        from_unit = self.units
+        var_name = self.var_name
+        import pyaerocom.units_helpers as uh
+        if self.var_name in uh.NON_SI_SUPPORTED.keys():
+            if uh.NON_SI_SUPPORTED[var_name] == new_unit:
+                _, fac = uh.unit_conversion_non_si_fac(var_name, from_unit)
+                cube = self.cube * fac
+                cube.units = new_unit
+                cube.var_name = var_name
+                self.cube = cube
+        else:
+            self.grid.convert_units(new_unit)
 
     def time_stamps(self):
         """Convert time stamps into list of numpy datetime64 objects
--- a/pyaerocom/units_helpers.py
+++ b/pyaerocom/units_helpers.py
@@ -17,6 +17,20 @@ from pyaerocom.helpers import seconds_in_periods
 
 VARS = const.VARS
 
+
+NON_SI_SUPPORTED = {
+    'concno3' : 'ug N m-3'
+}
+
+NON_SI_UCONV_MUL_FACS = pd.DataFrame([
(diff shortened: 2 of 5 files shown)


jgliss:
@ejgal thanks for looking into this. Can you change the base branch from `master` to `main-dev`? Cheers!

ejgal:
Ops, changed it now @jgliss.

pyaerocom/griddeddata.py
```diff
@@ -684,12 +685,18 @@ def _get_info_from_filenames(self):
...
+        from_unit = self.units
+        var_name = self.var_name
+        import pyaerocom.units_helpers as uh
+        if self.var_name in uh.NON_SI_SUPPORTED.keys():
+            if uh.NON_SI_SUPPORTED[var_name] == new_unit:
+                _, fac = uh.unit_conversion_non_si_fac(var_name, from_unit)
+                cube = self.cube * fac
+                cube.units = new_unit
+                cube.var_name = var_name
+                self.cube = cube
```
ejgal, line 697: This kind of works for allowing conversion to non SI units.
unit_conversion_non_si_fac is just a copy of unit_conversion_custom_fac..

ejgal: But not sure how it will fit into the flow we need @jgliss.
We could check NON_SI_SUPPORTED in readgridded, and convert the units there, but that is not something we always want to do..

pyaerocom/units_helpers.py
```diff
@@ -17,6 +17,20 @@
...
 VARS = const.VARS
 
+
+NON_SI_SUPPORTED = {
+    'concno3' : 'ug N m-3'
+}
+
+NON_SI_UCONV_MUL_FACS = pd.DataFrame([
+    ['concno3', 'ug/m3', 'ug N m-3', 3],
+], columns=['var_name', 'from', 'to', 'fac']).set_index(['var_name', 'from'])
```
jgliss, line 27: Can we harmonise this somehow with the already existing `UCONV_MUL_FACS`, that is, that `NON_SI_SUPPORTED` also includes those variables?

pyaerocom/units_helpers.py
```diff
@@ -63,7 +77,7 @@ def unit_conversion_fac_custom(var_name, from_unit):
     if from_unit in UALIASES:
         from_unit = UALIASES[from_unit]
     try:
-        info = UCONV_MUL_FACS.loc[(var_name, str(from_unit)), :]
+        info = df.loc[(var_name, str(from_unit)), :]
```
jgliss, line 80: I think `df` is not defined ...

pyaerocom/units_helpers.py
```diff
@@ -76,6 +90,25 @@ def unit_conversion_fac_custom(var_name, from_unit):
...
+                            'for dulplicate entries')
+    except KeyError:
+        raise UnitConversionError('Failed to convert unit {} (variable {}). '
+                                  'Reason: no custom conversion factor could '
+                                  'be inferred from table '
+                                  'pyaerocom.units_helpers.UCONV_MUL_FACS'
+                                  .format(from_unit, var_name))
+    return (info.to, info.fac)
+
+
```
jgliss, line 111: I am not sure I understand why `NON_SI_UCONV_MUL_FACS` is defined and why this function is needed? Shouldn't that just be added to `UCONV_MUL_FACS`?

jgliss (changes requested):
I think we need to revise this and harmonise with what was already in units_helpers.py.

pyaerocom/units_helpers.py
jgliss: Oh, forget my comment(s) @ejgal, I confused that we have to go in the other direction here, that is, from standard AeroCom to 'ug N m-3'... now I also understand what you were wondering in `ReadGridded`...

pyaerocom/units_helpers.py
jgliss: Again, I confused things here!

jgliss:
@ejgal, I merged your changes into #249, so you can probably close this one. See my comments in #249.

PR #246, opened 2020-11-19
Add new updated workflow

### LINKED ISSUE #245: Tests not running after recent changes to actions
Our test suite is not running anymore after set-env and att-path was deprecated due to security issues ([link](https://github.blog/changelog/2020-10-01-github-actions-deprecating-set-env-and-add-path-commands/).
I'm guessing one of those commands are used by [setup-conda](https://github.com/s-weigand/setup-conda)
or [checkout](https://github.com/actions/checkout).
We could try using the latest version of both as a possible quickfix, or look into other workarounds.

ejgal:
Update version of actions in workflow file. Fixes #245.

### DIFF
--- a/.github/workflows/CI.yml
+++ b/.github/workflows/CI.yml
@@ -10,14 +10,14 @@ jobs:
         python-version: [3.5, 3.6, 3.7, 3.8]
 
     steps:
-    - uses: actions/checkout@v2
+    - uses: actions/checkout@v2.3.4
     - name: Set up Python ${{ matrix.python-version }}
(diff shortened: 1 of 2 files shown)


ejgal:
I tried adding a new workflow file and bumping the versions which worked fine.
After I made the changes on the old workflow file and pushed it the checks are not appearing, but I think this will work after merging.

jgliss (approved):
All good (I think) =)

jgliss:
@ejgal thanks, I merge this now to see if it works!

PR #249, opened 2020-11-20
Check ebas concvars

jgliss:
@ejgal, I added some minor fix which I would like to merge into your branch.


jgliss:
@ejgal do you want to review or should I just merge this into your branch? :)

jgliss:
@ejgal, other idea: I continue forward from here and change the base branch to `main-dev`, since I have merged in your work from `read_ammonia_nitric`. I continue the work with reading the summed variables from the models, since we need that rather urgent. So you can probably close #244.

PR #254, opened 2020-11-25
Dev webeval multiobs

jgliss:
Just opening this to checkout if there is anything in dev-webeval-multiobs branch that should be incorporated. The last change in that branch was in May 2020...


jgliss:
I think this is too far departed from the current state, and the few changes that were applied here are probably obsolete, thus closing this.

PR #265, opened 2020-12-02
Fix failing tests

jgliss:


ejgal (commented):
The changes to test_read_ebas was made after changes to read_ebas in [this commit](https://github.com/metno/pyaerocom/commit/31c3f4177d3f98864425ba3452b1441eff0d4c00#). It seems the purpose was to disable checking if the ebas files actually exist due to performance issues. Maybe we should add the check back in, and change test_read_ebas accordingly (that is, back to the low number of stations)? @jgliss, let me know what you thank and I can fix it quickly.

Other changes looks ok :)

ejgal:
Ah, then the changes make sense. I thought this was just something you did while developing.

ReadEbas.read raises a datacoverage error if i try to read any of the files that are not present, which should be fine.

PR #266, opened 2020-12-03
Implement GHOSTv1.3

hansbrenna:
I had to update one standard QA flag in the GHOST reader. Added flag 131 after suggestion from Dene.


hansbrenna:
This should be very simple.

jgliss (approved):
Good to go!

PR #268, opened 2020-12-03
Second attempt to fix issue 263

### LINKED ISSUE #263: using resample_how='max' stops unit from being inferred when colocating
@jgliss Since yesterday, colocation fails when using `resample_how='max'. It prints the following warning:

`Cannot infer unit when aggregating using max. Please set unit in returned data object!`

I've tracked it to the commit b04bd27 with commit message: "Minor updates in GriddedData.resample_time (i.e. only keep unit for aggr mean, median, std)"

In griddeddata.py function _resample_time_xarray() you added:
```
        if how in ('mean', 'median', 'std'):
            data.units = self.units
        else:
            print_log.info(
                f'Cannot infer unit when aggregating using {how}. Please set '
                f'unit in returned data object!')
```

Can 'max' be safely added to the list of accepted aggregators?

jgliss:
Some temporal resampling aggregators do not preserve the unit of the data. E.g. if hourly precip data in units of mm hr-1 is resampled to daily using `sum` as an aggregator, the units will need to be changed to mm d-1. While this can be certainly automated (and should be at some point), for now, this PR introduces a list of "safe" aggregators in `TimeResampler` that preserve the units of the input data and checks if this criterion is fulfilled whenever `TimeResampler.resample` is called.

This has been implemented in `GriddedData.resample_time` in order to check if units in output data object are preserved. This should fix #263, but needs further looking into at some point.

### DIFF
--- a/pyaerocom/griddeddata.py
+++ b/pyaerocom/griddeddata.py
@@ -1546,7 +1546,6 @@ class GriddedData(object):
             resolution than this object
         """
         #from pyaerocom.tstype import TsType
-
         to = TsType(to_ts_type)
         current = TsType(self.ts_type)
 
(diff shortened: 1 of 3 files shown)


jgliss:
@hansbrenna I think this should fix the resampling issue, it is not the most general solution but now it will see at least - for stepwise resampling - if any of the entries in `how` applied to the data is *not* preserving the units, and then set the flag accordingly.

Can you review quickly?

hansbrenna:
It looks fine to me. I've tested it in my setup and it works.

jgliss:
Thanks!

PR #270, opened 2020-12-09
Fix lustre tests for release 0.10.0

ejgal:
This should fix all lustre tests except from some ghost tests where I cannot access the data at the moment.


ejgal:
@hansbrenna could you checkout this branch and try to run the ghost tests?

jgliss (approved):
Good to go =) Thanks! Maybe, on the long term, we should get rid of tests that depend on the data on lustre.

PR #272, opened 2020-12-15
Fixes for running EMEP report

ejgal:
Some features in ReadMscwCtm got lost in a merge at some point.
This PR adds some tests and fixes for the missing features.

### DIFF
--- a/pyaerocom/io/read_mscw_ctm.py
+++ b/pyaerocom/io/read_mscw_ctm.py
@@ -255,6 +255,16 @@ class ReadMscwCtm(object):
         GriddedData
         """
 
+        if isinstance(ts_type, dict):
+            try:
+                ts_type = ts_type[var_name]
+            except Exception:
+                const.print_log.info('Setting ts_type to None, since input '
+                                     'dict {} does not contain specification '
+                                     'variable to read {}'.format(ts_type,
+                                                                  var_name))
+                ts_type = None
+
         if self.filepath is None and self.data_dir is None:
             raise ValueError('filepath or data_dir must be set before reading.')
         elif self.data_dir is not None and ts_type is None:
--- a/pyaerocom/io/read_mscw_ctm.py
+++ b/pyaerocom/io/read_mscw_ctm.py
@@ -324,6 +324,8 @@ class ReadMscwCtm(object):
         new_unit = units
         if units == '' and prefix == 'AOD': #
             new_unit = '1'
+        elif units == '' and prefix == 'AbsCoef':
+            new_unit = '1/m'
         elif units == 'mgS/m2' or units == 'mgN/m2':
             raise NotImplementedError('Species specific units are not implemented.')
             new_unit = units
--- a/pyaerocom/io/read_mscw_ctm.py
+++ b/pyaerocom/io/read_mscw_ctm.py
@@ -309,7 +309,7 @@ class ReadMscwCtm(object):
             data.attrs['units'] = self.preprocess_units(data.units, EMEP_prefix)
             cube = data.to_iris()
             if ts_type == 'hourly':
-                cube.coords('time').convert_units('hours since 1900-01-01')
+                cube.coord('time').convert_units('hours since 1900-01-01')
             gridded = GriddedData(cube, var_name=var_name, ts_type=ts_type, convert_unit_on_init=False)
 
             if EMEP_prefix in ['WDEP', 'DDEP']:
@@ -333,7 +333,7 @@ class ReadMscwCtm(object):
         if units == '' and prefix == 'AOD': #
             new_unit = '1'
         elif units == '' and prefix == 'AbsCoef':
-            new_unit = '1/m'
+            new_unit = 'm-1'
             new_unit = units
         elif units == 'mgS/m2' or units == 'mgN/m2':
             raise NotImplementedError('Species specific units are not implemented.')
--- a/pyaerocom/io/read_mscw_ctm.py
+++ b/pyaerocom/io/read_mscw_ctm.py
@@ -12,7 +12,6 @@ import sys
 import os
 import glob
 
-from xarray.coding.times import infer_calendar_name
 import pyaerocom as pya
 from pyaerocom import const, print_log, logger
 from pyaerocom.exceptions import VarNotAvailableError, VariableDefinitionError
--- a/pyaerocom/io/read_mscw_ctm.py
+++ b/pyaerocom/io/read_mscw_ctm.py
@@ -344,8 +344,6 @@ class ReadMscwCtm(object):
             new_unit = '1'
         elif units == '' and prefix == 'AbsCoef':
             new_unit = '1/m'
-        elif units == 'mgS/m2' or units == 'mgN/m2':
-            raise NotImplementedError('Species specific units are not implemented.')
             new_unit = units
         return new_unit
 
--- a/pyaerocom/io/read_mscw_ctm.py
+++ b/pyaerocom/io/read_mscw_ctm.py
@@ -11,6 +11,8 @@ import numpy as np
 import sys
 import os
 import glob
+
+from xarray.coding.times import infer_calendar_name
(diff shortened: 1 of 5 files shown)


pyaerocom/io/read_mscw_ctm.py
```diff
@@ -11,6 +11,8 @@
 import sys
 import os
 import glob
+
+from xarray.coding.times import infer_calendar_name
```
ejgal, line 15:
```suggestion
```

ejgal:
@jgliss, can you have a look at this please?
This was needed to get the rerun of EMEP report working.

pyaerocom/io/read_mscw_ctm.py
```diff
@@ -255,6 +259,16 @@ def read_var(self, var_name, ts_type=None, **kwargs):
         GriddedData
         """
 
+        if isinstance(ts_type, dict):
```
jgliss, line 262: In which case is ts_type a dictionary here? Can this happen in the higlevel tools? I.e. in `Colocator`?

pyaerocom/io/read_mscw_ctm.py
```diff
@@ -302,6 +318,8 @@ def read_var(self, var_name, ts_type=None, **kwargs):
             data.time.attrs['standard_name'] = 'time'
             data.attrs['units'] = self.preprocess_units(data.units, EMEP_prefix)
             cube = data.to_iris()
+            if ts_type == 'hourly':
+                cube.coords()[0].convert_units('hours since 1900-01-01')
```
jgliss, line 322: Is it certain that first index is **always** the time dimension?

jgliss: Maybe it is better to try `cube.coord('time')` to make sure you are acting on the time dimension.

pyaerocom/io/read_mscw_ctm.py
```diff
@@ -324,6 +342,8 @@ def preprocess_units(units, prefix=None):
         new_unit = units
         if units == '' and prefix == 'AOD': #
             new_unit = '1'
+        elif units == '' and prefix == 'AbsCoef':
+            new_unit = '1/m'
         elif units == 'mgS/m2' or units == 'mgN/m2':
             raise NotImplementedError('Species specific units are not implemented.')
```
jgliss, line 348: I think this is "sort of" implemented now so it may be okay to remove this "elif" for N and S units.

pyaerocom/io/test/test_read_mscw_ctm.py
```diff
@@ -37,6 +37,13 @@ def test_read_emep_read_var(path_emep):
         r = ReadMscwCtm(data_dir=data_dir)
         r.read_var('vmro3', 'daily')
 
+@testdata_unavail
+def test_ts_type_dict(path_emep):
+    data_dir = path_emep['data_dir']
+    r = ReadMscwCtm(data_dir=data_dir)
+    data = r.read_var('vmro3', ts_type={'vmro3':'daily'})
+    assert data.ts_type == 'daily'
```
jgliss, line 45: I still do not understand why one would pass ts_type as dict in `read_var` and if this happens somewhere in the higher level code, it should probably be fixed there (if possible). @ejgal can you check that?

jgliss (commented):
All good, but I have some minor comments, maybe you could check them first, thanks!

pyaerocom/io/test/test_read_mscw_ctm.py
ejgal: It's from Colocator:
```
model_ts_type_read : :obj:`str` or :obj:`dict`, optional
        may be specified to explicitly define the reading frequency of the
        model data. Not to be confused with :attr:`ts_type`, which specifies
        the frequency used for colocation. Can be specified variable specific
        by providing a dictionary.
```
This can be passed all the way to the reader in some cases.
In ReadGridded._eval_vert_which there is also a check for ts_type being a dict, so there might be a usecase for this.

We could make sure to extract the ts_type for the variable with:
```
if isinstance(kwargs['ts_type'], dict):
    kwargs['ts_type'] = kwargs['ts_type'][var_name]
```
right before calling read_var.

Doing this does not break any of the tests at least.

pyaerocom/io/read_mscw_ctm.py
ejgal:
```suggestion
                cube.coords('time').convert_units('hours since 1900-01-01')
```

ejgal: Good point. Not sure how much the EMEP files can vary, this seems like a better solution.

pyaerocom/io/read_mscw_ctm.py
ejgal, line 348:
```suggestion
```

pyaerocom/io/test/test_read_mscw_ctm.py
ejgal: Should I try to fix this as proposed, or do we leave it for now @jgliss?

pyaerocom/io/read_mscw_ctm.py
ejgal: I tested it now, and I think the "sort of" implementation we did is not included in the release_0_10_0 branch.

pyaerocom/io/test/test_read_mscw_ctm.py
jgliss: Feel free to update as proposed, but please make an issue for it, so I can check at some point later what was the idea behind it :) ? It looks like this is for some edge cases and I am not sure anymore. But it seems it is also placed wrongly in if it is evaluated in `ReadGridded._eval_vert_which` ...

pyaerocom/io/read_mscw_ctm.py
ejgal: I added this back in for now.

ejgal:
This is updated now, can you have a brief look @jgliss?

pyaerocom/io/test/test_read_mscw_ctm.py
ejgal: Done: #276

pyaerocom/io/read_mscw_ctm.py
```diff
@@ -302,6 +308,8 @@ def read_var(self, var_name, ts_type=None, **kwargs):
             data.time.attrs['standard_name'] = 'time'
             data.attrs['units'] = self.preprocess_units(data.units, EMEP_prefix)
             cube = data.to_iris()
+            if ts_type == 'hourly':
+                cube.coords('time').convert_units('hours since 1900-01-01')
```
jgliss, line 312: I don't think `cube.coords('time')` works (cube.coords method returns a list I think), I think the syntax is `cube.coord('time')`. Please check that. Thanks.
```suggestion
                cube.coord('time').convert_units('hours since 1900-01-01')
```

pyaerocom/io/read_mscw_ctm.py
```diff
@@ -324,9 +332,11 @@ def preprocess_units(units, prefix=None):
         new_unit = units
         if units == '' and prefix == 'AOD': #
             new_unit = '1'
+        elif units == '' and prefix == 'AbsCoef':
+            new_unit = '1/m'
```
jgliss, line 336:
```suggestion
            new_unit = 'm-1'
```
just for consistency, I think we should avoid /unit and rather use unit-1 nomenclature.

jgliss (changes requested):
Can you please check the minor issues? Cheers!

jgliss (approved):
Looks good to me =) Takk!

ejgal:
Thanks! I tested .coord interactively, but of course ended up writing coords ;)

jgliss:
> Thanks! I tested .coord interactively, but of course ended up writing coords ;)

That's why we should always review each other's changes =)

PR #274, opened 2020-12-16
Fix issue 273

jgliss:


jgliss:
@hansbrenna waiting for the tests to complete but this may work now. Can you review?

hansbrenna:
I think this looks fine. Will it print a lot of warnings if a lot of stations are excluded? I guess that's fine, you get a feeling for how much data you're throwing out.

hansbrenna (approved):

jgliss:
Thanks @hansbrenna

PR #280, opened 2020-12-17
Enable computation of wet deposition from wconc and pr in ReadEbas

jgliss:


jgliss:
@hansbrenna can you please review this asap? This is needed for the evaluation of wet deposition, it provides code that enables the EBAS reading of:

- wetso4 (from concprcpso4 and pr)
- wetno3 (from concprcpno3 and pr)
- wetnh4 (from concprcpnh4 and pr)

I need these changes in branch `dev-wetdep-eval` to continue with implementation of the high level evaluation of these variables (I suspect it will not work out of the box, at least not for all models, due to the units).

I also implemented a new feature that makes sure that measurements are invalidated that are not in the temporal resolution provided in the EBAS NASA Ames files, using start and stop timestamps of each measurement and ensuring that the associated measurement time corresponds to the file frequency (with +/- 5% tolerance level). 

For instance, if the NASA Ames file says it is daily, then all measurements in that file are flagged invalid that are not within a measurement period of (86400 +/- 4320) s, as inferred from the provide start / stop meas info.

hansbrenna (approved):
I believe it's fine :)

Issue #133, opened 2020-06-05
We need tests for modules under pyaerocom/web


jgliss:
Tests for pyaerocom/web modules are missing and urgently needed as there is several parallel developments going on at the moment. @hansbrenna and @ejgal, please keep that in mind. 

Tests should, if possible, rely on the content of testdata-minimal, so that they are executed in the automatic CI. If needed testdata-minimal may be extended by small, representative, subsets of your data.

The setup of the test environment is defined in conftest.py

Creating dummy data (cf. #131) is also an option and may be possible to use for testing certain things.

Let me know if you have any questions or suggestions!

jgliss:
I started with this and implemented some tests for `ObsConfigEval` in #136.

ejgal:
Here is the current test coverage of the web module.
Seems next step on this issue should be writing some tests for trends_evaluation.

```
Name                                                                   Stmts   Miss  Cover
------------------------------------------------------------------------------------------
pyaerocom/web/aerocom_evaluation.py                  658    207    69%
pyaerocom/web/helpers.py                              56     11    80%
pyaerocom/web/helpers_evaluation_iface.py            464    211    55%
pyaerocom/web/helpers_trends_iface.py                 67     59    12%
pyaerocom/web/trends_evaluation.py                   832    745    10%
pyaerocom/web/utils.py                               110    101     8%
```

jgriesfeller:
aerocom evaluation is not used for new stuff anymore. Closing this issue.

PR #281, opened 2020-12-18
Test for aerocom evaluation

ejgal:
Check that the expected folders are created during `run_evaluation` and that
the timeseries for one station has correct values.

#133


jgliss (approved):
Awesome! Tusen takk =)

Issue #296, opened 2021-01-12
Remove outlier removal in low-level colocation routines


jgliss:
Currently, the low-level colocation routines (in *colocation.py*) are overloaded with functionality that does not belong there. Most prominent is the removal of outliers and whether or not that should be done for obs, model, or both, while accounting for checks of variable aliases, etc., see also #50. 

Thus, outlier removal in model and obsdata should be done ***before*** calling one of the low-level colocation routines (which should only do colocation, indeed, and not data preparation). Hence, the logic of outlier removal belongs in the *Colocator* class which provides a high-level interface for colocation and is also used in the web-tools.

jgliss:
This was done.

Issue #50, opened 2019-10-10
Review parameter naming and clarify in colocation routines


jgliss:
E.g.:

- In low-level colocation routines: get rid of `var_ref_keep_outliers` and `var_keep_outliers` and handle directly through args `var_outlier_ranges` and `var_ref_outlier_ranges`.
- Change input type of the latter (currently dict, passed through, should be list and handled in `Colocator` class just before calling the respective method.
- Harmonise naming of attributes `ColocationSetup` and input args for low-level colocation routines (e.g. `model_keep_outliers` vs. `var_keep_outliers`). Check all other args.

jgliss:
@AugustinMortier It's on the todo list, thanks for your helpr ;)

jgliss:
All relevant outlier removal settings should also be added to metadata dict of `ColocatedData` object that is returned by the colocation methods.

PR #297, opened 2021-01-12
Revise logic related to outlier removal in low and high level colocation routines

jgliss:
Related to #50 and #296.


jgliss:
@hansbrenna I reorganised a little the handling of outlier removals in colocation routines. This comes with minor changes in the API, that is, if you want to set outlier ranges for obs explicitly in a config file, you should use `obs_outlier_ranges` instead of `var_ref_outlier_ranges` (even though the old way should still work as well). And if you happen to want to remove outliers from models (which I never do), you can specify that via `model_remove_outliers` (as `remove_outliers` will only be applied to obs, as before). 

Most importantly, outlier removal is now done in Colocator and the lowlevel colocation functions do not support outlier removal anymore.

Can you briefly review please? 

Cheers, 
Jonas

jgliss:
@hansbrenna next one ready for review :) would you be able to check it out? Many thanks in advance!

hansbrenna (approved):
Good to clean up the outlier handling. It was redundant before :)

jgliss:
> Good to clean up the outlier handling. It was redundant before :)

Thanks!

PR #310, opened 2021-02-02
New obsdata reading routines for CAMS84 project

jgliss:
Ungridded reading routines for the following data-sets are to be implemented:

- [x] North American AirNow dataset 
- [x] EEA data
- [x] Chinese MarcoPolo dataset
- [x] Processing of combined obs datasets
- [x] Using HTAP regions for regional statistics

Focus (for now) should be on concentrations / mixing ratios of PM10, PM2.5, O3 and NO2. 

**Note:** before this merged into `main-dev` the new routines should have docstrings and (at least essential) tests.


jgliss:
@hansbrenna last but not least (on this review marathon) ;)

hansbrenna (approved):

PR #312, opened 2021-02-03
Updates in new EEA NRT reading routine

jgliss:


jgliss:
@jgriesfeller here are some minor updates I did so far, most importantly

- I renamed `vmro3` to `conco3` 
- Made sure, `ts_type` is only assigned once to `StationData` (it uses now the info in the data files rather than `hourly`)
- Made sure variable units are assigned based on unit reported in first line of the file

jgliss:
@jgriesfeller I tested the current version locally and now it runs through the web processing (PM10, PM2.5). 

For implementation of vmro3 (and later other VMR variables), the most consistent way of doing this is via `AUX_REQUIRES` and `AUX_FUNS` and base class method `compute_additional_vars`, to be consistent with the other readers. As conversion function, `mathutils.concx_to_vmrx` can be used. I can implement this if you like, but it will not be before tomorrow until I get to this.

jgriesfeller:
I hope to be able to look at the calculation of vmro3 tomorrow morning. Since I added 4 more similar variables to the reading in another branch, it's probably the best if I implement that. So let's finish this PR for concpm10 and concpm25?

jgliss:
Yes, would be great if you could implement that, so I leave the state as it is here. Could you review please so I can merge this into dev-cams84?

pyaerocom/io/read_eea_aqerep.py
```diff
@@ -235,7 +254,8 @@ def read_file(self, filename, vars_to_retrieve=None,
                     if idx in time_indexes:
                         # make the time string ISO compliant so that numpy can directly read it
                         # this is not very time string forgiving but fast
-                        data_dict[header[idx]][lineidx] = np.datetime64(rows[idx][0:10]+'T'+rows[idx][11:19]+rows[idx][20:])
+                        dtstr = rows[idx][0:10]+'T'+rows[idx][11:19]+rows[idx][20:]
```
jgriesfeller, line 257: I did this in one line on purpose to save some time. This is run millions of times...

pyaerocom/io/read_eea_aqerep.py
```diff
@@ -389,14 +426,13 @@ def get_station_coords(self, meta_key):
...
-        meta_key : `str`
+        meta_key : str
             string with the internal station key
         """
-
         ret_data = {}
-        ret_data['latitude'] = self._metadata[meta_key][self.LATITUDENAME]
-        ret_data['longitude'] = self._metadata[meta_key][self.LONGITUDENAME]
-        ret_data['altitude'] = self._metadata[meta_key][self.ALTITUDENAME]
+        ret_data['latitude'] = float(self._metadata[meta_key][self.LATITUDENAME])
```
jgriesfeller, line 433: this float is a Python float that needs to be converted to numpy if used. Use np.float or np.float_ instead,

pyaerocom/io/read_eea_aqerep.py
```diff
@@ -555,14 +605,8 @@ def read(self, vars_to_retrieve=None,
...
-    r = ReadEEAAQEREP()
-    r.logger.setLevel(logging.INFO)
-    var_names_to_test = ['concpm25', 'vmro3','concpm10', ]
-    station_id = 'AT2KA71'
-    for var_name in var_names_to_test:
-        data = r.read(vars_to_retrieve = [var_name])
-        print('data read')
-        print(data[station_id])
-        print(data[station_id][var_name])
+    ddir = '/home/jonasg/MyPyaerocom/data/obsdata/EEA_AQeRep.NRT/download'
```
jgriesfeller, line 608: we should find a way for testing that does not replace each others testing routines...
This is not ideal since I will add mine with the newest commit again...

pyaerocom/io/read_eea_aqerep.py
```diff
@@ -389,14 +426,13 @@ def get_station_coords(self, meta_key):
...
             string with the internal station key
         """
-
         ret_data = {}
-        ret_data['latitude'] = self._metadata[meta_key][self.LATITUDENAME]
-        ret_data['longitude'] = self._metadata[meta_key][self.LONGITUDENAME]
-        ret_data['altitude'] = self._metadata[meta_key][self.ALTITUDENAME]
+        ret_data['latitude'] = float(self._metadata[meta_key][self.LATITUDENAME])
+        ret_data['longitude'] = float(self._metadata[meta_key][self.LONGITUDENAME])
+        ret_data['altitude'] = float(self._metadata[meta_key][self.ALTITUDENAME])
```
jgriesfeller, line 435: np.float again

pyaerocom/io/read_eea_aqerep.py
```diff
@@ -435,9 +471,15 @@ def read(self, vars_to_retrieve=None,
         elif isinstance(vars_to_retrieve, str):
             vars_to_retrieve = [vars_to_retrieve]
 
+        if len(vars_to_retrieve) > 1:
+            raise NotImplementedError('So far, only one variable can be read '
+                                      'at a time...')
+        var_name = vars_to_retrieve[0]
+        const.print_log.info('Reading EEA data')
```
jgriesfeller, line 478: what's the reason not to use logger directly? (just curious)

jgriesfeller (commented):
Please have a look at the comments.

pyaerocom/io/read_eea_aqerep.py
jgliss: I want it to crash for now if more variables are input. This is just because of time pressure, it should also be able to read multiple variables in one go.

pyaerocom/io/read_eea_aqerep.py
jgliss: float also works since it is not written into the numpy _data array at the moment, but is only in the metadata. but I can change if you want.

pyaerocom/io/read_eea_aqerep.py
jgliss: Sorry that I removed your code. I think testing should not happen under `__main__`, which is for playing around and changes all the time also in all other pyaerocom modules. If we want to test some code, we should do this in a separate python file. Best would be to create a "representative" small subset (best <1MB) of the data and add it to testdata-minimal and add a test module `pyaerocom/io/test/test_read_eea_aqerep.py` that runs tests based on that subset.

jgliss:
@jgriesfeller I think I addressed all points in the latest push. Can you check and confirm the merge? For the multivar reading: the easiest would be probably to rename `read` to `read_var` and implement read as:

```python
def read(self, vars_to_retrieve, **kwargs):
   # code that checks that vars_to_retrieve is a list
   
   # empty data object
   data_out = UngriddedData()

   for var in vars_to_retrieve:
      data_out.append(self.read_var(var, **kwargs))
   return data_out

PR #313, opened 2021-02-04
added stuff for EEA AQeRep.v2 reading

jgriesfeller:
Hi Jonas, 

this is the EEA reading for the validated data in addition. I just put the NRT stuff into a base class and inherited that with the NRT and validated reading just changing the path and the product code

Regarding __main__ I like the fact that there's an example in there that works. I just put a user name dependency in there so that is does not disturb other developers

BTW, should we maybe rename ReadEEAAQEREP to ReadEEAAQEREPNRT? Because that is essentially what it's doing.


jgriesfeller:
the PR is ready for review now

pyaerocom/io/helpers_units.py
```diff
@@ -202,3 +203,29 @@ def unitconv_wet_depo_from_emep(data, time, ts_type = "monthly"):
...
+        input data in output unit
+    """
+
+    conv_factors = {}
+    conv_factors['vmro3'] = np.float_(0.5)
+
+    try:
+        output_data = input_data * conv_factors[var_name]
+    except KeyError:
+        output_data = input_data
```
jgliss, line 230: This is too dangerous, if we do it this way for now, at least raise an Exception if if is not done for O3.

```suggestion
    if not var_name in conv_factors:
        raise ValueError(f'Cannot convert {var_name}, missing conversion factor...')
    output_data = input_data * conv_factors[var_name]
```

pyaerocom/io/read_eea_aqerep_base.py
```diff
@@ -0,0 +1,643 @@
...
+    FILE_MASKS['conco3'] = '**/*_7_*_timeseries.csv'
+    FILE_MASKS['vmro3'] = '**/*_7_*_timeseries.csv'
+    FILE_MASKS['concno2'] = '**/*_8_*_timeseries.csv'
+    FILE_MASKS['concco'] = '**/*_10_*_timeseries.csv'
+    FILE_MASKS['concno'] = '**/*_38_*_timeseries.csv'
+    FILE_MASKS['concpm25'] = '**/*_6001_*_timeseries.csv'
+
+    # conversion factor between concX and vmrX
+    CONV_FACTOR = {}
+    CONV_FACTOR['vmro3'] = np.float_(0.5)
```
jgliss, line 125: You have conv factor here and in the function in helpers_units. Can we delete the function in helpers_units then?

jgliss (changes requested):
I went over it quickly but it is a little hard to see what really changed since the essential code was moved into a new module, so the diff is not really telling. Seems okay from a quick glance, apart from the new function in `io/helpers_units.py`, please address my comments regarding this.

pyaerocom/io/read_eea_aqerep_base.py
jgriesfeller: I did not want this in the helper_units.py, but when I deleted that, I could not use AUX_REQUIRES anymore since one can only live with the other it seems. But I thought it would be good to follow the practice to use AUX_REQUIRES.
I was basically waiting on your comment on this :-)

jgliss: Okay, but if we need that function, then it should at least raise an Exception. I made a code suggestion, which you could just commit directly.

jgriesfeller: agreed

jgriesfeller: I have committed your suggestion

jgliss (approved):
Looks good :)

PR #315, opened 2021-02-05
Updates needed to combine several observations into one superobservation

jgliss:
Needed in CAMS84 to combine EEA, AirNow and chinese (MarcoPolo) data into a single super-obs network. 

**NOTE**: this PR will also includes some updates related to menu creation of the web tools, for instance, to have more control on the order of observations and models in menu and heatmap display.

PR #316, opened 2021-02-08
Further updates related to processing of regions

jgliss:
NOTE: this includes API changes related to handling of regions, e.g. data/regions.ini is deprecated and all the related logic is now in `pyaerocom/region_defs.py`.

Also, processing of HTAP regions for the web-tools is implemented here.


jgliss:
@hansbrenna this one is based on top of #315 and enables using binary masks for region processing. It includes quite some API updates related to the Region class, etc., e.g. there is no regions.ini file anymore. 

It also cleans the API of `ColocatedData` a little. Would be great if you could briefly check that as well, after #315. I used this to run this

https://aerocom-evaluation.met.no/overall.php?project=cams84&exp=eval-2020

so basic functionality should be given (also the tests pass, which is promising).

hansbrenna:
This REALLY is a lot, but I think you can merge it. I will have to look more carefully at the new functionality in these PRs.

hansbrenna (approved):

jgliss:
> This REALLY is a lot, but I think you can merge it. I will have to look more carefully at the new functionality in these PRs.

Thanks a lot for checking on such short notice @hansbrenna. We will review all this properly when merging `dev-cams84` into `main-dev`, for that, I will also add a lot more tests.

PR #318, opened 2021-02-16
Minor fixes related to trends processing

jgliss:
E.g., new EBAS dataset contains also 20daily SO2 data which cannot be handled by pyaerocom, and will lead to a crash in `UngriddedData.to_station_data_all`, rather than ignoring the corresponding sites. This should be fixed.


pyaerocom/stationdata.py
```diff
@@ -755,10 +755,6 @@ def _ensure_same_var_ts_type_other(self, other, var_name):
 
             from pyaerocom.helpers import get_lowest_resolution
             ts_type = get_lowest_resolution(ts_type, ts_type1)
-        from pyaerocom.tstype import TsType
-        _tt = TsType(ts_type)
-        if _tt.mulfac != 1:
-            ts_type = _tt.next_lower.val
```
jgliss, line 761: This is irrelevant now, after recent updates in TsType.

hansbrenna (approved):

PR #326, opened 2021-02-25
Add code to compute "representative" testdata for the test suite

jgliss:
Reason: it is quite cumbersome to add "real" data into the minimal testdataset used by the CI engine and we should aim for keeping the minimal testdataset as minimal as possible (it already has 127MB of data in it). 
Instead, in addition to the real data used for tesing, it would be useful to test some core functionality against fakedata that is created on the run when the tests are being executed.


jgliss:
Site note. This should involve a revision of file `pyaerocom/test/synthetic_data.py` which will probably be obsolete after the more general set of testdata is available

PR #332, opened 2021-03-01
clone and add pyaerocom-tutorials when building docs

jgliss:


jgliss:
@adonath since we talked about it: I found a way to clone another repo needed to build the docs, but not via RTD configuration file but using python build in subprocess module and calling git clone directly in conf.py.

jgliss:
@hansbrenna here is the PR that will include the pyaerocom-tutorials repo when building the docs on readthedocs. Note that this goes directly into master. Can you review?

PR #335, opened 2021-03-03
Some bug fixes in webtools

jgliss:
2 bugs are fixed that should only affect json file calculation of gridded / gridded analyses in web-tools.


hansbrenna (approved):
Looks fine.

Issue #324, opened 2021-02-25
New AeroVal projects structure


AugustinMortier:
@jonasg In order to follow the new projects structure considered, the following needs to be done:

- [ ] experiments.json file in project repository, with information on the regional scale ("World", "Europe", ...), available statistics ("monthly", "daily", "hourly"), years (2020, 2019, 2018, ...), ...
```
{
    "Gliss et al 2020": {
        "years": [2010]
        "stats": ["monthly"],
        "region": "World"
    },
    "Optics HighRes": {
        "years": [2010]
        "stats": ["monthly", "daily", "hourly"],
        "region": "Europe"
    }
}
```
- [ ] menu.json in each experiment. Removed the experiment level.
- [ ] ranges.json in each experiment
- [ ] regions.json in each experiment
- [ ] zones.json in each experiment
- [ ] statistics.json in each experiment
- [ ] `hm/glob_stats.json` monthly file renamed to `hm/glob_stats_monthly.json` so we could support hourly statistics

AugustinMortier:
Maybe the years could actually be strings too, so in the case of trends, we could have periods instead of single years:
```
"years": ["2000-2020"]
```

jgliss:
@AugustinMortier thanks for the writing it up, sounds all very doable. As discussed, I will start looking into this after the open PR's for version 0.11.0 release are incorporated (see #325) and v0.11.0 is on the way. Since the required updates for the new AeroVal structure and features require major changes in the API of the web-tools, I will assign this issue to version 0.12.0.

jgliss:
@hansbrenna this will also be relevant for you, but will come after 0.11.0 is released.

AugustinMortier:
regions should have a unique id, so they could have the same name (e.g: EUROPE (rectangle) / Europe (HTAP))
    "WORLD": {
        "name": "World"
        "minLat": -90,
        "maxLat": 90,
        "minLon": -180,
        "maxLon": 180
    },

jgriesfeller:
Just a comment about names:
I at least will use the rectangular and the HTAP regions in parallel, so the names should not be identical. EUROPE and Europe might confuse users, but I don't have a good proposal to solve that right now. I guess this was the reason we kept the original HTAP names in the old aerocom-tools

jgliss:
Guess this can be closed now :)

PR #336, opened 2021-03-04
Updates for AeroVal 

jgliss:
See #324


jgliss:
@avaldebe this is sort of a monster, I would suggest to move this directly into main-dev and develop missing tests from there, what do you think?

avaldebe:
> @avaldebe this is sort of a monster, I would suggest to move this directly into main-dev and develop missing tests from there, what do you think?

sure, just met me know what modules and methods to prioritate

PR #342, opened 2021-03-10
add first version of .coveragerc file

jgliss:


jgliss:
@hansbrenna @MichaelSchulzMETNO @jgriesfeller I set up codecov.io which will now always be run when our [CI.yml](https://github.com/metno/pyaerocom/blob/ci-update/.github/workflows/CI.yml) github (which runs the test-suite) action is triggered (on pushes and PRs). You can see above that for this PR codecov informed that if we merge this PR into master, coverage will decrease by ca 4%. In this case this is fine as I flagged some files that should not be included in the coverage check (e.g. files containing tests) by adding a file called [.coveragerc](https://github.com/metno/pyaerocom/blob/ci-update/.coveragerc). to establish an initial coverage state in master, that will form the base used by codecov to evaluate whether a future commit will result in an increase or decrease of test coverage (which is why this PR goes into master branch). Detailed information about pyaerocom's coverage status can be found here:

https://app.codecov.io/gh/metno/pyaerocom

jgliss:
@jgriesfeller could you briefly check / confirm this one? Thanks!

jgriesfeller (approved):
Looks good to me

PR #343, opened 2021-03-10
Emep updates

jgliss:
Revision of EMEP gridded reader.


jgliss:
@hansbrenna I think this is ready for review too. It is sort of relevant for you, since it relates to direct reading of EMEP output. I did a refurbishing of the EMEP reader and tried to increase associated test coverage. Could you take a look at it please?

jgliss:
@hansbrenna Tested it on lustre and running into `MemoryError` when `ReadMscwCtm.compute_var('concno3', 'daily')` is run based on Base_day.nc file. I think the problem is the `to_iris` conversion of the xarray DataArrays. Funny is, it works on my local machine...

jgliss:
I updated the EMEP reader now, so that it computes auxiliary variables based on loaded DataArray instances, rather than converting the individual aux variable fields into iris Cubes and then applying aux func. This way it worked to evaluate concno3 on lustre, without a MemoryError. However, we need to test this a little more, especially also for hourly data.

jgliss:
Results from Svetlana's test experiment are online here:

https://aerocom-evaluation.met.no/main.php?project=playground&exp=eqsam

jgliss:
@hansbrenna I think this is ready for review now. Could you check it out?

pyaerocom/io/read_mscw_ctm.py
```diff
@@ -252,82 +393,68 @@ def read_var(self, var_name, ts_type=None, **kwargs):
...
-        gridded.metadata['from_files'] = filepath # ReadGridded cannot concatenate several years of data if this is missing
+        gridded.metadata['from_files'] = self.filepath
 
         # Remove unneccessary metadata. Better way to do this?
         for metadata in ['current_date_first', 'current_date_last']:
             if metadata in gridded.metadata.keys():
                 del(gridded.metadata[metadata])
         return gridded
 
+    def _read_var_from_file(self, var_name_aerocom, ts_type):
```
hansbrenna, line 445: Should we have docstrings for private methods as well? It can be difficult to understand the difference between all the methods called something with read or load.

hansbrenna (approved):

pyaerocom/io/read_mscw_ctm.py
jgliss: Absolutely! Thanks for pointing this out :)

PR #347, opened 2021-03-11
draft of applying mon->yr constraints when calculating overall statistics

hansbrenna:
This might be a neat solution. Use the non-NaN stations in the `yearly` colcoacated data object to select the stations used when calculating the heatmap statistics.


hansbrenna:
There are still some issues I need to work out. Right now it only works for 'WORLD' for some reason

jgliss:
Great that you work on this. However, this needs to be an option that can be specified in the config file, and by default, it should do it without applying this yearly constraint (you can then activate the option for cams61)!

jgliss (changes requested):
Indeed, this seems to be a neat way of doing this, but make sure it is optional in the config file, and default processing should do it the way it used to do it. Perhaps add a boolean attr `annual_stats_constrained` (or similar) in the `AerocomEvaluation` class, defaulting to `False`.

hansbrenna:
I agree

hansbrenna:
The bug that makes it work only for WORLD seems to be in my main branch.

jgliss:
@hansbrenna consider making this a draft PR until the feature is ready.

jgliss:
@hansbrenna I think I have something which I would call BETA finished ;) I guess, you should probably revise what I did here. It seems to work for the colocated data file you sent me and CI also passes (even though this does not yet include tests for the new option).

jgliss:
@hansbrenna I think the failing CI here is for the PR is due to failing tests in `Hans_main_dev`, for `dev-annual-limit` all tests seem to pass now.

jgliss (approved):
@hansbrenna I had a quick look over what we did here and it seems okay. Please review as well, since I also modified quite a bit here. Cheers!

hansbrenna (commented):
I think this looks fine. I will merge it now.

Issue #348, opened 2021-03-12
Regions for web processing no longer working properly


hansbrenna:
I have tested this with the current state of main-dev and Hans_main_dev both before and after merging main-dev. After merging main-dev a few days ago, the only region which is processed in the web interface is WORLD. Has there been any API changes that requires me to change my CFG file or is this a bug?

I debugged into web.helpers_evaluation_iface._process_heatmap_data() and the regions_id dictionary only contains the WORLD region.

The config file is [here](https://gitlab.met.no/aerocom-evaluation/data/-/blob/master/config_files/cfg_cams61_p3_BASE.py)

![Screenshot from 2021-03-12 09-57-33](https://user-images.githubusercontent.com/12746882/110916724-64977280-8319-11eb-9da4-826d701dd439.png)

hansbrenna:
Seems to only not work for country regions. I tried in main-dev with default regions and then it works.

jgliss:
Need to check that, there has been substantial changes related to the regions processing (you should know since you reviewed the PR ;))

jgliss:
Boils down to test coverage, as I try to make clear since weeks... and which is why I am saying that only official releases should be used for production, at least until we have better test coverage, see here:

https://app.codecov.io/gh/metno/pyaerocom

But writing tests takes time, which needs to be considered and acknowledged in such a big project!

hansbrenna:
I'm struggling to get this to work. I'm not sure of the intended uses of regs, regborders and reg_names in init_web_regions(), and also how the contents of them are supposed to be.

In the for loop  below, the contents of regs has to be og type pyaerocom.region.Region since it needs the attribute region_id. How do i initialize the country regions as type pyaerocom.region.Reion?

```Python
    for regname, reg in regs.items():
        regnames[reg.region_id] = regname
```

hansbrenna:
I think I'm on it now

jgliss:
If you fix it, please make PR into main-dev for the fix

jgliss:
So good if you start off with main-dev ;)

hansbrenna:
i have done that

hansbrenna:
It works now. Going to make a PR.

hansbrenna:
I think this can be closed

PR #349, opened 2021-03-12
fix for #348

hansbrenna:


jgliss (approved):
Looks good, even though it would be good to add a test for the new function. But I am currently working on tests for helpers_evaluation_interface.py anyways, so I will take that up there! But for the future: if you add / modify code, make sure it is covered by tests, codecov makes it easy to spot the affected lines of code!

Issue #341, opened 2021-03-09
Check processing of model climatology fields (yr 9999 in filename)


jgliss:

jgliss:
This has not been done for v0.11.0 as originally indended, but has been fixed in #354 which will go into v0.12.0.

PR #354, opened 2021-03-16
Implement support for evaluation of model climatology files #341

jgliss:
AeroCom model climatology files are usually stored with year 9999 in the filename. Since pyaerocom checks whether the time inferred from the filename corresponds to what is provided in the time dimension of the NetCDF file, there will need to be some exceptions in case the filename uses 9999.

In addition, pandas does not support timestamps for year 9999. 

Tasks:

- [x] ensure AeroCom files with year 9999 can be read in (and leave time dimension therein unchanged).
- [x] ensure that such data can successfully be co-located without getting confused with regard to the time dimension.
- [x] add tests that ensure these things are working as expected.


jgliss:
I think the feature `model_use_climatology` in `Colocator` will not be used very frequently, therefore I added some restrictions that make sure, that make sure it only works for single year setups (since there is only on 9999 file representing a climatological field).

It remains unclear, and will not be solved in this PR, how to deal with model (and obs) climatologies in a more general case.

pyaerocom/helpers.py
```diff
@@ -1319,14 +1339,13 @@ def start_stop_from_year(year):
...
-        - :obj:`pandas.Timestamp`: end timestamp
+    numpy.datetime64
+        start datetime
+    numpy.datetime64
+        stop datetime
     """
-    start = to_pandas_timestamp(year)
-    stop = to_pandas_timestamp('{}-12-31 23:59:59'.format(year))
+    start = np.datetime64(f'{year}-01-01T00:00:00')
+    stop = np.datetime64(f'{year}-12-31T23:59:59')
```
jgriesfeller, line 1348: This in principle leaves out the last second of the year. Usually it's better to use `np.datetime64(f'{year+1}-01-01T00:00:00')` and check for < 
Likely not important at this point, but somewhat wrong anyhow

jgliss: Hmm, I am sure there is a reason for that, as this was implemented deliberately at some point, not sure anymore why, I think it is related to data slicing, but I am rather sure, that it will change behaviour if we change that now, e.g. probably end up including the Jan of the following year in case of monthly data, etc.

jgriesfeller (approved):
The changes in pyaerocom/io/iris_io.py and pyaerocom/colocation_auto.py where so huge that itt's hard to tell for me if all of that makes sense. So I just hope that you ran enough tests.

pyaerocom/helpers.py
jgriesfeller: this is not an important comment and should not prevent you from merging. That's why I approved the PR.

jgliss:
> The changes in pyaerocom/io/iris_io.py and pyaerocom/colocation_auto.py where so huge that it's hard to tell for me if all of that makes sense. So I just hope that you ran enough tests.

I think I did add tests here, and colocation_auto.py is one of the modules with a rather good test coverage. But we will add more tests in main-dev before the next release.

PR #356, opened 2021-03-17
Adding glob_stats_yearly.json

hansbrenna:
This enables calculation of glob_stats_yearly.json which includes the spatial correlations as R/R_spearman


jgliss:
@hansbrenna is this a draft PR or should I review this maybe?

jgliss (changes requested):
Looks good, however, you should probably register the new file add it to method `AerocomEvaluation._heatmap_files`.

hansbrenna:
Thanks. I will do that.

jgliss (approved):
Looks good!

Issue #352, opened 2021-03-16
EBAS reading: infer ts_type also for intermediate frequencies 


jgliss:
Currently, ReadEbas checks if res. code corresponds to the most common meas. period MCP in the file (e.g. 86400s for daily), and if not, it checks if MCP corresponds to any of the base frequencies (i.e. hourly, daily,...) but not any permutation of it, like 2daily, 3daily.

This leads to some data not being imported, which needs to be fixed.

PR #357, opened 2021-03-17
Implement feature 352

jgliss:
These changes should enable to infer ts_type from total seconds also for 2daily, 3hourly, 20daily, etc. Adding this to the other EBAS dev branch, for simplicity. This refers to #352.


hansbrenna (approved):
I will merge and test it.

PR #368, opened 2021-04-07
Trends updates

jgliss:
This will mostly comprise a revision of how we compute trends with the goal to improve the flexibility of relevant input parameters for the trends computation, for instance, slope confidence, resampling constraints, frequency used to compute the trend (currently yearly), minimum number of yearly averages, etc.

I propose that all relevant settings should go into a new dict-like container class `TrendsSettings` that is used in `TrendsEngine`.

This PR should also include some efforts to improve the documentation of the trends computation and options the user has.


jgliss:
Changed base of this as parts of it are already merged into aeroval-updates, so it should go via that branch.

PR #373, opened 2021-04-09
Implement additional statistics for spatial and temporal correlation in web-tools

jgliss:


jgliss:
@hansbrenna I think this is ready for review.

hansbrenna (approved):

jgliss:
@hansbrenna I updated the code now after talking with @hildefagerli yesterday. Temporal R is now computed like **Option 1** in [this sketch](https://docs.google.com/presentation/d/1D8c_u6tO82p3hPuKP_Dz8ZS4NtAKIJoU2iC_EolXkzw/edit?usp=sharing) while for spatial correlation the equivalent of **Option 2** is used. For both, there is median and mean correlation being added to the heatmap json files. 

I also updated the code for the new annual constraint option, so that it works also for multi-year datasets (in which case invalid sites are identified on a year-to-year basis) and that it also works for 4D colocated data with lat / lon dimensions rather than station_name dimension (for satellite or model / model eval).

@hansbrenna can you review please?

Issue #375, opened 2021-04-09
Drop conversion of variable names to all lowercase


jgliss:
Currently, `Variable` class converts input variable names to all lowercase. This was originally done to support old variable names, such as ABS550AER. However, this is prone to and can create silent errors and is an issue with new conventions used in `ReadEbas` to support variable units such as `ug N m-3` as well as `ug m-3` for concentrations and differentiate them via the variable name, that is, e..g. `concNno3` vs `concno3`. 

Thus, this behavior should be removed on that level and may need to be handled for exception cases in associated reader classes (e.g. for evaluation of old data). 

Thus, this may break some functionality in case modeldata uses old variable conventions.

PR #376, opened 2021-04-09
Drop var name lowercase

jgliss:
See #375


jgliss:
@hansbrenna I think this could fix your problem with the new variable names, see also #375. Can you review please?

hansbrenna (approved):

Issue #378, opened 2021-04-13
ReadTheDocs build fails to import pyaerocom


jgliss:
The current [.readthedocs.yml](https://github.com/metno/pyaerocom/blob/master/.readthedocs.yml) uses conda to install all dependencies, which do not include pyaerocom itself. Thus, everything under [API](https://pyaerocom.readthedocs.io/en/latest/api.html) section in docs is empty. Need to find out how to specify in config file to run `python setup.py install` in addition to the installation of the requirements environment file.

PR #379, opened 2021-04-13
Fix broken RTD build

jgliss:
See #378


jgliss:
@jgriesfeller our docs were broken on ReadTheDocs and I think this should fix it. Could you briefly review please? Thanks!

jgriesfeller (approved):
looks fine

PR #372, opened 2021-04-08
EBAS variable updates (ebas_config.ini)

jgliss:
Not sure if this is really needed but I used it for the little test I did  here:

https://github.com/jgliss/my_scripts/blob/master/2021/EBASMC_fmfso4_sites.csv

(for details see issue #359)

PR #381, opened 2021-04-14
EBAS variable updates

jgliss:
Updates related to EBAS variable definitions (replaces former PR #372).


jgriesfeller:
Looks fine

Issue #384, opened 2021-04-15
AerocomEvaluation.update_interface() is not working properly anymore in main-dev


hansbrenna:
Running AerocomEvaluation.update_interface() by itself (not the call to update_interface in run_evaluation()) fails with the following error:

  File "/home/hansb/lustre/storeA/project/fou/kl/CAMS61/postprocessed_data/repositories/aerocom_evaluation/data_subbed/config_files/cfg_cams61_p3_BASE.py", line 708, in <module>
    stp.update_interface()

  File "/home/hansb/pyaerocom/pyaerocom/web/aerocom_evaluation.py", line 1728, in update_interface
    self.update_menu(**opts)

  File "/home/hansb/pyaerocom/pyaerocom/web/aerocom_evaluation.py", line 1738, in update_menu
    update_menu_evaluation_iface(self, **opts)

  File "/home/hansb/pyaerocom/pyaerocom/web/helpers_evaluation_iface.py", line 365, in update_menu_evaluation_iface
    avail = _get_available_results_dict(config)

  File "/home/hansb/pyaerocom/pyaerocom/web/helpers_evaluation_iface.py", line 225, in _get_available_results_dict
    tab = config.get_web_overview_table()

  File "/home/hansb/pyaerocom/pyaerocom/web/aerocom_evaluation.py", line 1668, in get_web_overview_table
    for f in self.all_map_files:

  File "/home/hansb/pyaerocom/pyaerocom/web/aerocom_evaluation.py", line 379, in all_map_files
    if not os.path.exists(self.out_dirs['map']):

KeyError: 'map'

hansbrenna:
i.e: stp.update_interface()

jgliss:
@hansbrenna I removed the code that creates all directories for an experiment into the processing methods of `AerocomEvaluation`. I did this, since I wanted to avoid folders being created in the wrong place if someone just instantiates the class with minimum required input (I often had experiment folders in random places because of that). 

But I see your point and it is indeed nice to have the folders around (and `AerocomEvaluation` knowing about them) also in other methods. I will check how this can be solved. For now, try calling:

```python
stp = pya.web.AerocomEvaluation(proj_id='bla', exp_id='blub')
stp.init_dirs()
stp.update_interface()
```

hansbrenna:
Thanks. I will do that.

PR #385, opened 2021-04-16
Further revise handling of output directories in AerocomEvaluation

jgliss:
Addresses #384


jgliss:
Merged this locally into updates-web, where I already had started related developments.

Issue #386, opened 2021-04-16
NO2 from EBAS in vmr


jgliss:
Most of the EBAS `vmrno2` reading fails at the moment as most data seems to be stored in units of ug N/m3. These should be converted assuming conversion factor.

This issues is closely linked with #369

Issue #369, opened 2021-04-08
Ozone from EBAS in vmr


hildefagerli:
This is the old issue that EBAS has the double amount of ozone data in ug/m3 than ppb (although obs are done in ppb, they are reported in ug/m3 and many sites do not report the conversion factor). Should have a default conversion factor (2) for ug/m3-ppb for sites that do not report conversion factor (except for high mountain sites probably)

jgliss:
This is a good point, I forgot about that, some pre-thoughts on that:

1. For individual NASA Ames files: need to check whether it can happen that one file contains both mass conc (unit ug m-3) and vmr (nmole mole-1), in different columns?
2. If 1 is True, then in such cases `ReadEbas` needs to pick the right column dependent on what AeroCom variable name the user specified (i.e., `conco3` vs `vmro3`), based on the corresponding units in column matches.
3. In cases where a NASA Ames file contains only one column match, e.g. `conco3` (ug m-3), but the user wants `vmro3` (nmole mole-1), automatic conversion should be attempted, in which case the following should be checked:  
   a. If the column meta data in the NASA Ames file specifies a conversion factor, use that.
   b. If not, check if temperature and pressure is accessible in metadata and compute conversion factor from that (using [concx_to_vmrx](https://github.com/metno/pyaerocom/blob/4945b1aebf51f637f824149744329a33b4ff9968/pyaerocom/mathutils.py#L767).
   c. If temp and / or pressure are not available, use station altitude and standard atmosphere to compute conversion factor.
   d. If a, b and c fail, we could use a constant conversion factor of 0.5.

If any of 3a - 3d is done, information should be added in the associated metadata block of the `UngriddedData`, about how the conversion was done and which assumptions were made.

PR #387, opened 2021-04-16
Convert EBAS gas data of NO2 and O3 reported as mass conc into vmr and vice versa

jgliss:
See #369 and #386 for details.


jgliss:
Strange, I just pushed to this branch (`ebas-gases-vmr`), but this is not registered / updated in this PR... @jgriesfeller or @hansbrenna any ideas?

jgliss:
Trying to close this PR and reopen.

jgliss:
@hansbrenna I think this is now also ready, with this update you will probably get more sites both for O3 and NO2, so probably good if you use this in the CAMS61 phase 3 experiments. Could you have a look please? Cheers

hansbrenna (approved):

PR #389, opened 2021-04-22
Harmonisation of ungridded reading API

jgliss:
This turned out to be a larger PR than expected, but here it is, harmonised largely the API of ungridded reading with the gridded reading API.

**NOTE**: this will likely break old scripts and has some major implications on the ungridded API.

Major renaming includes:

In lowlevel ungridded readers: 

- `DATASET_PATH` deprecated, use `data_dir`
- `dataset_path` renamed to `data_dir`
- `dataset_to_read` deprecated, use `data_id`

In `ReadUngridded`

- `data_dir` renamed to `data_dirs`
- `datasets_to_read` renamed to `data_ids`

### DIFF
--- a/pyaerocom/colocation_auto.py
+++ b/pyaerocom/colocation_auto.py
@@ -907,7 +907,7 @@ class Colocator(ColocationSetup):
         obs_reader = ReadUngridded(self.obs_id, data_dirs=self.obs_data_dir)
         try:
             obs_vars = obs_reader.get_vars_supported(self.obs_id,
-                                                 vars_to_read)
+                                                     vars_to_read)
         except ValueError:
             raise DataCoverageError('No observation variable matches found for '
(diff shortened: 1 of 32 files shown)


jgliss:
@jgriesfeller and @hansbrenna I already want to make you aware of these upcoming changes related to the ungridded API (even though it is not fully finalised). This will improve user friendliness, by using similar variable naming conventions in the gridded and ungridded API (e.g. `data_id` and `data_dir` instead of `dataset_to_read` and `DATASET_PATH`). However, this will not be implemented in the upcoming version 0.11.0 but only in 0.12.0, so there is sufficient time to adapt existing code.

jgliss:
@hansbrenna I think this is also ready for review, if you have time. Cheers

hansbrenna (approved):
I think this should be fine as well.

PR #393, opened 2021-05-20
added some hydrocarbon variables

hansbrenna:


jgliss:
@hansbrenna is this ready for review? I would be interested to test these new variables for trends processing.

pyaerocom/data/variables.ini
```diff
@@ -1173,7 +1173,7 @@ var_name = emibvoc
 description = Biogenic NMVOC emissions
 standard_name = tendency_of_atmosphere_mass_content_of_biogenic_nmvoc_expressed_as_carbon_due_to_emission
 var_type = Emission fluxes
-unit = kg m-2 s-1
+unit = mg C m-2 h-1
```
jgliss, line 1176: Should we apply the same rule here with var_name? i.e. name it emiCbvoc?

pyaerocom/molmasses.py
```diff
@@ -12,7 +12,11 @@
              'no2'      : 46.0055,
              'hno3'     : 63.01,
              'nh3'      : 17.031,
-             'co'       : 28.010}
+             'co'       : 28.010,
+             'isop'     : 68.12,
+             'glyoxal'  : 58.036,
+             'glyox'    : 58.036,
+             'hcho'     : 30.026,}
```
jgliss, line 19: I trust these are correct ;)

pyaerocom/web/helpers_evaluation_iface.py
```diff
@@ -1417,6 +1421,13 @@ def compute_json_files_from_colocateddata(coldata, obs_name,
             _write_stationdata_json(ts_data, out_dirs['ts'])
 
     if coldata.ts_type == 'hourly':
+        #writing hourly heatmap
+        hm_fname_hourly = HEATMAP_FILENAME_EVAL_IFACE_HOURLY
+        hm_file_hourly = os.path.join(out_dirs['hm'], hm_fname_hourly)
+        hm_data = hm_all['hourly']
+        _add_entry_heatmap_json(hm_file_hourly, hm_data, web_iface_name, obs_var,
```
jgliss, line 1428: would be great if you could add some tests for the diurnal processing!

jgliss (commented):
Looks good, just have some minor comments. Please add a test that runs the diurnal processing code that you implemented in `compute_json_files_from_colocateddata`. There is some functions in `import pyaerocom._conftest_helpers` that create dummy colocated data, also hourly, that you could use perhaps.

pyaerocom/molmasses.py
hansbrenna: They are

pyaerocom/web/helpers_evaluation_iface.py
hansbrenna: I will try. good opportunity to start writing tests

pyaerocom/data/variables.ini
hansbrenna: I set the unit to be the same as in the netcdf files to avoid unit conversion. This is not ideal, but I want to leave it like this. I'll think about the naming convention

jgliss (approved):
maybe we can skip tests for the diurnal processing as this would need to be rewritten for the new aeroval anyways. good for me if you merge this.

Issue #395, opened 2021-05-31
requirements on setup.py


avaldebe:
plan to submit a PR later this week with addressing the following

- [x] update classifiers
- [x] add minimum python version 
- [x] add dependencies
- [x] use the correct python version on CI

jgliss:
@avaldebe not sure what is the best way to handle dependencies but I would prefer to have a single location where we have all requirements. Currently, this is the [pyaerocom_env.yml](https://github.com/metno/pyaerocom/blob/master/pyaerocom_env.yml) which is also used in the [CI.yml](https://github.com/metno/pyaerocom/blob/master/.github/workflows/CI.yml) workflow file. Perhaps, *setup.py* could just import requirements from *pyaerocom_env.yml*?

avaldebe:
@jgliss 

It is not straight forward as the `iris` conda package correspond to the `scitools-iris` [python package](https://pypi.org/project/scitools-iris/) , and I'm not sure if the python package version matches the conda package version.

In any case, installing the requirements on a Python 3.6 virtual environment has been challenging.
`scitools-iris` requires `pyke`, which needs to be installed as `scitools-pyke`...

Also, `scitools-iris` recently dumped support for Python 3.6
Maybe the minimum versions of `scitools-iris` and Python should be bumped.

How about using a build tool that uses `pyproject.toml`?
I quite like poetry, because of its dependency resolver.

jgliss:
@avaldebe I understand. I would quite like to keep conda as the main way of installing pyaerocom (preferably, users should use the latest version via *conda-forge*). That way, dependencies are also resolved nicely. I am not familiar with these other build tools, like `poetry`, but if you have a nice alternative solution to the conda option, of course we can add that! And I would say we drop support for P3.6 in any case. What do you think?

avaldebe:
> @avaldebe I understand. I would quite like to keep conda as the main way of installing pyaerocom (preferably, users should use the latest version via _conda-forge_). That way, dependencies are also resolved nicely. 

OK, but this would mean duplicating the dependencies. That way we can be explicit with the conda/python package names and versions.

> I am not familiar with these other build tools, like `poetry`, but if you have a nice alternative solution to the conda option, of course we can add that! And I would say we drop support for P3.6 in any case. What do you think?

[...]

PR #396, opened 2021-05-31
add python classifiers and dependencies

avaldebe:
fixes issue #395

- [x] update classifiers
- [x] add minimum python version
- [x] add dependencies
- [x] use the correct python version on CI
- [x] drop python 3.6 support

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -53,7 +53,26 @@ setup(
     ],
 
     python_requires = '>=3.6,<4',
-    install_requires    =   [],
+    install_requires = [
+        'scitools-iris>=2.1.0',
+        'scitools-pyke>=1.1.1',
+        'xarray>=0.16.0',
+        'cartopy>=0.16.0',
+        'matplotlib>=3.0.1',
+        'scipy>=1.1.0',
+        'pandas>=0.23.0',
+        'seaborn>=0.8.0',
+        'geonum',
+        'LatLon23', # required by geonum
+        'SRTM.py', # required by geonum
+        'numpy',
+        'simplejson',
+        'requests',
+        'reverse-geocode',
+        'tqdm',
+        'openpyxl',
+        'nbsphinx', # for docs via rtd configuration file v2
+    ],
     dependency_links    =   [],
     description = ('pyaerocom model evaluation software'),
     long_description = readme,
--- a/setup.py
+++ b/setup.py
@@ -38,12 +38,14 @@ setup(
         # Indicate who your project is intended for
         'Intended Audience :: Science/Research',
         'Intended Audience :: Education',
+        'Topic :: Scientific/Engineering :: Atmospheric Science',
 
         # Pick your license as you wish (should match 'license' above)
-        'License :: OSI Approved :: BSD License',
+        'License :: OSI Approved :: GNU Lesser General Public License v3 (LGPLv3)'
 
         # Specify the Python versions you support here. In particular, ensure
         # that you indicate whether you support Python 2, Python 3 or both.,
+        'Programming Language :: Python :: 3 :: Only',
         'Programming Language :: Python :: 3.6',
         'Programming Language :: Python :: 3.7',
         'Programming Language :: Python :: 3.8',
--- a/setup.py
+++ b/setup.py
@@ -71,8 +71,8 @@ setup(
         'reverse-geocode',
         'tqdm',
         'openpyxl',
-        'nbsphinx', # for docs via rtd configuration file v2
     ],
+    extras_require={'docs':['nbsphinx']},
     dependency_links    =   [],
     description = ('pyaerocom model evaluation software'),
     long_description = readme,
--- a/setup.py
+++ b/setup.py
@@ -46,13 +46,12 @@ setup(
         # Specify the Python versions you support here. In particular, ensure
         # that you indicate whether you support Python 2, Python 3 or both.,
         'Programming Language :: Python :: 3 :: Only',
-        'Programming Language :: Python :: 3.6',
         'Programming Language :: Python :: 3.7',
         'Programming Language :: Python :: 3.8',
         'Programming Language :: Python :: 3.9',
     ],
 
-    python_requires = '>=3.6,<4',
+    python_requires = '>=3.7,<4',
     install_requires = [
         'scitools-iris>=2.1.0',
         'scitools-pyke>=1.1.1',
--- a/setup.py
+++ b/setup.py
@@ -72,7 +72,10 @@ setup(
         'tqdm',
         'openpyxl',
     ],
-    extras_require={'docs':['nbsphinx']},
+    extras_require = {
+        'docs':['nbsphinx'],
+        'test':['pytest>=3','pytest-dependency','pytest-cov'],
+    },
     dependency_links    =   [],
     description = ('pyaerocom model evaluation software'),
     long_description = readme,
--- a/setup.py
+++ b/setup.py
@@ -44,7 +44,10 @@ setup(
 
         # Specify the Python versions you support here. In particular, ensure
         # that you indicate whether you support Python 2, Python 3 or both.,
-        'Programming Language :: Python :: 3.6'
+        'Programming Language :: Python :: 3.6',
+        'Programming Language :: Python :: 3.7',
+        'Programming Language :: Python :: 3.8',
+        'Programming Language :: Python :: 3.9',
     ],
 
     install_requires    =   [],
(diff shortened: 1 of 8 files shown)


avaldebe:
@jgliss 
[CI / build (3.6)](https://github.com/metno/pyaerocom/pull/396/checks?check_run_id=2737929153) was cancelled after 6h trying to update conda for Python 3.6. Maybe creating a new environment would have been faster, as it does not need to update every package on the base env.

setup.py
```diff
@@ -38,16 +38,41 @@
...
 
         # Pick your license as you wish (should match 'license' above)
-        'License :: OSI Approved :: BSD License',
+        'License :: OSI Approved :: GNU Lesser General Public License v3 (LGPLv3)'
 
         # Specify the Python versions you support here. In particular, ensure
         # that you indicate whether you support Python 2, Python 3 or both.,
-        'Programming Language :: Python :: 3.6'
+        'Programming Language :: Python :: 3 :: Only',
+        'Programming Language :: Python :: 3.6',
```
jgliss, line 49: maybe we should drop support for 3.6?

setup.py
```diff
@@ -38,16 +38,41 @@
...
+        'geonum',
+        'LatLon23', # required by geonum
+        'SRTM.py', # required by geonum
+        'numpy',
+        'simplejson',
+        'requests',
+        'reverse-geocode',
+        'tqdm',
+        'openpyxl',
+        'nbsphinx', # for docs via rtd configuration file v2
```
jgliss, line 74: Not sure, but does this have the potential to mess with conda env, if the user uses conda and does (from local clone):

```bash
conda activate pyaenv # env with all deps installed
python setup.py install # setuptools / pip could get confused with existing deps in pyaenv?
```

jgliss (commented):
Looks good, I am only concerned that setuptools could mess with conda envs when running `python setup.py install`.

setup.py
avaldebe: OK, but not on this PR.
I can write a new one once is PR is merged.

avaldebe: Unless we need to drop 3.6 to pass CI

setup.py
avaldebe: It should not, as long as iris is consistent with pip/conda versions.
It the case `python setup.py install` on CI would mess the environment, but all test passed so I would say it works.

jgliss (approved):

setup.py
jgliss: Let's try it out, nice if RTD would not need conda anymore.

setup.py
avaldebe: it seems dumping python 3.6 wont be nessesary

jgliss (approved):
looks good, I just hope that the pipy setup for the dependencies remains as stable as the conda resolver in the future, especially wrt iris updates. But I guess we will find out along the way. Probably, the installation instructions in the docs need some updates when we release these updates. One more question @avaldebe does the caching of the env detect changes in the deps and set it up from scratch if so?

avaldebe:
> ... does the caching of the env detect changes in the deps and set it up from scratch if so?

```YAML
    - name: Cache pip
      uses: actions/cache@v2
      with:        path: ~/.cache/pip
          key: ${{ runner.os }}-pip-${{ hashFiles('setup.py') }}
          restore-keys: |
               ${{ runner.os }}-pip-
```

`hashFiles('setup.py')` is the part of the cache key, so any changes to `setup.py'` will invalidate the cache

jgliss:
@avaldebe we discussed a little this morning in the Aerocom meeting about pip vs conda and we concluded the following:

- drop support for python 3.6
- stick with conda as package manager and not mix it with pypi and a too smart setup.py (especially also because of potential conflicts with iris and installation of pyaerocom into existing envs)

Thus, could you update this PR such that we are having requirements only in the conda environment file and stick with a simple setup.py that does not install any dependencies?

Not sure how this will affect the changes in the CI, but perhaps installing the conda envs should be sufficiently fast if we drop P3.6?

avaldebe:
> @avaldebe we discussed a little this morning in the Aerocom meeting about pip vs conda and we concluded the following:
> 
> * drop support for python 3.6

easily done

> * stick with conda as package manager and not mix it with pypi and a too smart setup.py (especially also because of potential conflicts with iris and installation of pyaerocom into existing envs)

I do not agree with this. IMO, packages on `pypi` should be usable from a pip install.

The [pypi description](https://.org/project/pyaerocom/) does not mention a further installation step.
The [pip install docs](https://pyaerocom.readthedocs.io/en/latest/install.html#via-pypi) do not mention further installation steps.
The current `setup.py` is incomplete, so that pyaerocom is only usable as a conda package.
Without addressing this issues, this package should be [yanked from `pypi`](https://pypi.org/help/#yanked)
or deleted altogether.

> > Thus, could you update this PR such that we are having requirements only in the conda environment file and stick with a simple setup.py that does not install any dependencies?

The current conda setup does not install any dependencies on CI.

```
python -m pip install . --no-deps
```

> Not sure how this will affect the changes in the CI, but perhaps installing the conda envs should be sufficiently fast if we drop P3.6?

It will be fast enough for avoiding a time out.
If I remember correctly, the Python 3.7 environment setup took a ~90 min, Python 3.8 ~30 min and Python 3.9 ~4 min.
Compared with the 3 min setup via pip, testing the python matrix on a conda environment is a literal waste of time.
Even the RTD build time benefits from a functional pip setup.

I think that there are at least two ways to keep the speed of installation and testing on CI.

IMO, the simpler and cleaner way would be to update the documentation.
Replacing `python setup.py install` by  `pip install --no-deps .`,
and `python setup.py develop` by `pip install --no-deps --editable .`
on [install from source docs](https://pyaerocom.readthedocs.io/en/latest/install.html#installing-from-source).

Other option would be to declare all requirements as extra/optional so in order to usable pip install one would have to
invoke `pip install pyearocom[deps]` (or `pip install -e .[deps]` for a development install from source). This would be an anti-pattern, as `pip install` has the `--no-deps` option for uses cases like the one on my previous solution.

Please consider this options, or invite me to the next Aerocom meeting so I can explain the benefits of having an usable pip installable benefits.

In the meantime, I'll drop Python 3.6 support and set the conda CI job as a matrix, so the time differences become evident.

jgliss:
@avaldebe, okay, you convinced me, that it should also work via pip :) 
We will make sure then that conda users that install from source into an existing env will use the `pip install --no-deps` way. My hope is anyways, that we will manage to release more frequently, so users can just install the release via conda-forge or via pip. Could you please update the installation instructions in the docs accordingly?

avaldebe:
> We will make sure then that conda users that install from source into an existing env will use the `pip install --no-deps` way. My hope is anyways, that we will manage to release more frequently, so users can just install the release via conda-forge or via pip. Could you please update the installation instructions in the docs accordingly?

I updated the installation instructions, but before pushing I wonder if I should "trim down" the conda matrix.
The conda job took 72 min to run with Python 3.7, 10 min Python 3.8 and 5 min with Python 3.9.
So maybe we could drop the Python 3.7 from the conda matrix, and rely on the pip tests for 3.7.

What do you think?

jgliss:
> > We will make sure then that conda users that install from source into an existing env will use the `pip install --no-deps` way. My hope is anyways, that we will manage to release more frequently, so users can just install the release via conda-forge or via pip. Could you please update the installation instructions in the docs accordingly?
> 
> I updated the installation instructions, but before pushing I wonder if I should "trim down" the conda matrix.
> The conda job took 72 min to run with Python 3.7, 10 min Python 3.8 and 5 min with Python 3.9.
> So maybe we could drop the Python 3.7 from the conda matrix, and rely on the pip tests for 3.7.
> 
> What do you think?

Sounds good!

avaldebe:
@jgliss 

Here is the rendered version of the install docs
https://pyaerocom--396.org.readthedocs.build/en/396/install.html

jgliss:
> @jgliss
> 
> Here is the rendered version of the install docs
> https://pyaerocom--396.org.readthedocs.build/en/396/install.html

Great, thanks a lot!

jgliss:
@avaldebe for me it is fine if you merge this, or is there anything missing?

avaldebe:
> @avaldebe for me it is fine if you merge this, or is there anything missing?

do not have such privileges

jgliss:
> > @avaldebe for me it is fine if you merge this, or is there anything missing?
> 
> do not have such privileges

Ups, now you do :)

PR #397, opened 2021-06-08
Add lint job to CI

avaldebe:
a follow up to PR #396


avaldebe:
Two days ago (on Oct 31st) `netCDF4-1.5.8` [released wheels](https://pypi.org/project/netCDF4/1.5.8/#files) for Python 3.10,
so it makes seance to start testing against this version.

Alas `cf-units` and `pyproj` have not released wheels for 3.10, so the the test failed to install.
In the meantime, the Python 3.10 build CI job is allowed to fail the test step

https://github.com/metno/pyaerocom/blob/4a925ecec5a0a7ebd69b8ecf63770c3b4e55dd72/.github/workflows/CI.yml#L60-L62

PR #400, opened 2021-06-18
Av trend patch

jgliss:
merging #368 with #336 and adding / fixing some tests.


jgliss:
@hansbrenna I would appreciate a review of this one but since it is a lot, and since it goes into aeroval-updates, I think we can also skip the detailed review at this stage. Please let me know, if you manage to take a quick look at it. Thanks!

I think we can also accept the codecov failure here, I will write the missing tests in **aeroval-updates** after merging.

@jgriesfeller note that this will be merged soon, so then **aeroval-updates** will be the new branch to use (for the next days).

hansbrenna:
I can review it now

hansbrenna (approved):
This was a lot. I could see no obvious errors at this point.

PR #407, opened 2021-09-03
fix #406

### LINKED ISSUE #406: Bug in new AeroVal tools: default colocation always regrids modeldata to 5x5 degrees lat/lon
This is a result of attr `ColocationSetup.regrid_res_deg` being set to 5 as default in `aeroval-updates` (coming from `av-trend-patch`).

jgliss:


jgriesfeller:
ignoring the codecov complaints, this looks fine

jgliss:
> ignoring the codecov complaints, this looks fine

will merge then!

Issue #405, opened 2021-09-03
Reading of EEAAQeRep.v2 data crashes pyaerocom when a file is incomplete


jgriesfeller:
as found in issue #404  the reader crashes when a file stops in the middle of a line with data (coming from an incomplete download)
by adding another exception after https://github.com/metno/pyaerocom/blob/b9b8ec397cdf956eed995d6d361486992e54350b/pyaerocom/io/read_eea_aqerep_base.py#L303-L304 this can be easily avoided

hansbrenna:
It crashed again, so there are more files that have the same problem.

jgriesfeller:
While testing I found another error:
this exception is wrong since it's allowed for countries not to provide data for every variable we might want to read:
https://github.com/metno/pyaerocom/blob/6612ec6ced010eeba63e96136796c262b397d299/pyaerocom/io/read_eea_aqerep_base.py#L456-L461

It's also annoying when testing locally since the EEA dataset is quite big

jgriesfeller:
should be fixed in #408

jgriesfeller:
closing issue due to merged PR

hansbrenna:
@jgriesfeller this still does not seem to work. When using main-dev I stille get the same problem

jgriesfeller:
It seems that there are files that don't even have a correct header (meaning that even the 1st line is broken). Preparing a patch and do proper testing this time...

jgriesfeller:
@hansbrenna I have merged some more robust reading code to main-dev. Please check and report errors.
Please close the issue in case the reading works for you now.

jgliss:
@jgriesfeller, seems like there are still some issues with the EEA reading, @dulte is looking into it to try and fix and will create a small PR for it.

jgriesfeller:
Very weird since the file found looks fine and I thought I went through the whole data set while testing. Investigating...

jgriesfeller:
it turnerd out that the file is actually UTF-16 encoded:
```
/lustre/storeA/project/aerocom/aerocom1/AEROCOM_OBSDATA/EEA_AQeRep.v2/renamed/SE/SE_6001_31058_2018_timeseries.csv: Little-endian UTF-16 Unicode text, with CRLF line terminators```

Making sure that these files are skipped for now

jgriesfeller:
closing this since it seems to work now.

Issue #404, opened 2021-09-02
Reading EEAAQeRep.v2 does not work


hansbrenna:
I'm trying to read EEAAQeRep.v2 and it does not work in either aerocom-updates or av-trend-patch branches. Is this a known problem @jgliss ?

```Python
import pyaerocom as pya
pya.io.ReadUngridded().read('EEAAQeRep.v2','concpm10')
```

 ```Python
92%|████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████▎                   | 12767/13952 [24:43<02:19,  8.50it/s]---------------------------------------------------------------------------
IndexError                                Traceback (most recent call last)
<ipython-input-3-5ba5e4aa2cd8> in <module>
----> 1 test=reader.read('concpm10')

~/pyaerocom/pyaerocom/io/read_eea_aqerep_base.py in read(self, vars_to_retrieve, files, first_file, last_file, metadatafile)
    554             _file = files[i]
    555             try:
--> 556                 station_data = self.read_file(_file, var_name=var_name)
    557             except TemporalResolutionError as e:
    558                 const.print_log.warning(f'{repr(e)}. Skipping file...')

~/pyaerocom/pyaerocom/io/read_eea_aqerep_base.py in read_file(self, filename, var_name, vars_as_series)
    300                         # sometimes there's no value in the file. Set that to nan
    301                         try:
--> 302                             data_dict[header[idx]][lineidx] = np.float_(rows[idx])
    303                         except ValueError:
    304                             data_dict[header[idx]][lineidx] = np.nan

IndexError: list index out of range

```

hansbrenna:
It fails also for concpm25 and conco3

jgriesfeller:
Since this is my code: That looks like a corrupted file. 
The filename would be interesting. You could just skip the line by extending the except statement after line 303.

But I can check myself as soon as I have finished some other work for Michael...

hansbrenna:
I think it's this file that causes trouble

/lustre/storeA/project/aerocom/aerocom1/AEROCOM_OBSDATA/EEA_AQeRep.v2/renamed/SE/SE_5_31069_2017_timeseries.csv'

hansbrenna:
@jgriesfeller the file has an incomplete line at the end:

[...]

PR #408, opened 2021-09-07
made EEA AQeRep.v2 reading more robust; added possibility to read gzipped files

jgriesfeller:
This pull request addresses the failed reading of incomplete files @hansbrenna discovered (issues #404 and #405) 
I also adds the capability to read compressed csv.gz files instead of just csv files which makes local development easier since the EEA data is quite huge.

To make use of that, one just has to gzip all the files in the data directories.

In case you wonder why I removed the DataSourceError exception: https://github.com/metno/pyaerocom/blob/6612ec6ced010eeba63e96136796c262b397d299/pyaerocom/io/read_eea_aqerep_base.py#L456-L461 It is possible for a country to not provide all variables we might want to read. So this exception is just wrong

### DIFF
--- a/pyaerocom/io/read_eea_aqerep_base.py
+++ b/pyaerocom/io/read_eea_aqerep_base.py
@@ -69,7 +69,7 @@ class ReadEEAAQEREPBase(ReadUngriddedBase):
     _FILEMASK = '*.csv'
 
     #: Version log of this class (for caching)
-    __version__ = '0.05'
+    __version__ = '0.06'
 
     #: Column delimiter
     FILE_COL_DELIM = ','
--- a/pyaerocom/io/read_eea_aqerep_base.py
+++ b/pyaerocom/io/read_eea_aqerep_base.py
@@ -32,6 +32,10 @@ Example
 -------
 look at the end of the file
 """
+import pathlib
+import shutil
+import tempfile
+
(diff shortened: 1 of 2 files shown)


pyaerocom/io/read_eea_aqerep_base.py
```diff
@@ -278,8 +294,10 @@ def read_file(self, filename, var_name,
             lineidx = 0
             for line in f:
                 rows = line.rstrip().split(file_delimiter)
+                # skip line if the # rows is not sufficient
+                if len(rows) < max_file_index_to_keep:
```
hansbrenna, line 298: I guess this is the fix for #404 / #405. Looks good.

hansbrenna (approved):
This looks fine to me

pyaerocom/io/read_eea_aqerep_base.py
jgriesfeller: I just found another error while testing with all zipped source files. updated PR coming soon...

jgriesfeller: forgot to increase version number of the reader

jgriesfeller: merging PR...

PR #410, opened 2021-09-07
add some AeroVal tests

avaldebe:
add tests to `pyaerocom.aeroval`

PR #414, opened 2021-09-07
Fix issue 413

jgliss:
Small PR that makes sure menu creation for AeroVal experiment does not crash anymore in case there are some outdated map json files lying around in the output folder. The latter can happen if some models or observations are removed in a future version of a config file and have not been cleared yet.


jgliss:
@avaldebe skipping the tests here as I believe this will be done in #410

jgliss:
@jgriesfeller  could you also have a quick glance at this one, so I can merge #414 (which already incorporates the changes here). Cheers!

jgriesfeller:
ready for merging

PR #422, opened 2021-09-10
Increase flexibility of ColocationSetup attributes 

jgliss:
Inheritance from ConstrainedContainer, in ColocationSetup disables updating attr `resample_how` from str type (default) to dict type (nested). Until a consistent API for resample_how and min_num_obs is defined, the strictness in ColocationSetup wrt to attr. types needs to be removed.


hansbrenna (approved):
looks good.

jgliss:
> looks good.

thanks, just adding some tests before merging

Issue #430, opened 2021-09-14
Dependencies in setup.py tend to mess with conda config


jgliss:
When installing pyaerocom in developing mode into an existing conda environment via 

```bash
pip install --no-deps -e .
``` 

the command line interface crashes, that is:

```bash
pya --ppiaccess
```

crashes as somewhere during the import of pyaerocom the "scitools-pyke>=1.1.1" is not given:

```bash
pkg_resources.DistributionNotFound: The 'scitools-pyke>=1.1.1' distribution was not found and is required by pyaerocom
```

I think this is due to the fact that "scitools-pyke>=1.1.1" is listed in setup.py as requirement @avaldebe, which is actually not a direct requirement for pyaerocom but I believe for iris (and pip should install that via iris). We could add this in the conda env but I resist a little and think we should re-discuss the recommended installation pathways (I, personally do not like the setup.py to be too smart, trying to solve the dependencies - via pip - as well, particularly because this has such high potential to mess with conda environments if not invoked in the correct way, with pip and --no-deps).

@jgriesfeller also had some issues in his pycharm configuration because of that, I think.

For now I will make a PR that removes pyke from setup.py so that the CLI works again.

We should discuss this on Monday I think.

jgliss:
But I guess removing pyke from setup.py will end up in crashing CI, @avaldebe, right?

jgriesfeller:
Regarding pycharm: During the analysis of the project it comes by the error and then it offers to install the dependency using conda (if one is using a conda env for the development).
Because the institute provides the production software for the meteorologists using conda now, we need to make sure that pyaerocom can be easily installed using that.
Personally I don't like pip and use it only if there is no other choice (e.g. missing packages in conda). I don't want that to be the default way of installing pyaerocom. Only the Python world creates pip packages, while conda is quite open these days (e.g. R, nco, cdo, etc).

To be discussed...

avaldebe:
> I think this is due to the fact that "scitools-pyke>=1.1.1" is listed in setup.py as requirement @avaldebe, which is actually not a direct requirement for pyaerocom but I believe for iris (and pip should install that via iris). We could add this in the conda env but I resist a little and think we should re-discuss the recommended installation pathways (I, personally do not like the setup.py to be too smart, trying to solve the dependencies - via pip - as well, particularly because this has such high potential to mess with conda environments if not invoked in the correct way, with pip and --no-deps).

[...]

PR #431, opened 2021-09-14
update dependencies in setup.py

jgliss:


jgliss:
As expected, CI is failing because of missing pyke dependency (see #430)

jgliss:
> As expected, pypi CI is failing because of missing pyke dependency (see #430)

PR #435, opened 2021-09-15
Minor fixes and add AeroVal in docs

jgliss:


jgliss:
@dulte could you review this PR please? Cheers

jgliss:
note @dulte the slight decrease in coverage is okay here, as @avaldebe is adding tests for the AeroVal code in #410 and we don't want to mess with that here.

pyaerocom/aeroval/helpers.py
```diff
@@ -181,10 +184,30 @@ def reorder_experiments_menu_evaluation_iface(menu_file, exp_order=None):
     new = sort_dict_by_name(current, pref_list=exp_order)
     write_json(new, menu_file, indent=4)
 
-def _check_statistics_periods(periods):
+def _check_statistics_periods(periods: list) -> list:
```
dulte, line 187: Shouldn't list annotations have type of list elements annotated? So list[str]?

jgliss: Good point, sounds reasonable (I am also newish to the world of type annotations).

dulte: Same, only knew about it because someone commented on one of my PRs a couple of weeks ago :P Other than that, the PR looks good

dulte:
I think this is ready to be merged

PR #439, opened 2021-09-16
Fix 438 (among other small bugs related to AeroVal tools)

jgliss:
This PR incorporates the changes made in #435 so it should be merged after that one.


dulte (approved):
Looks good :)

Issue #423, opened 2021-09-10
web_interface_name functionality for aeroval broken in main-dev


hansbrenna:
using a different web interface name than the obs_id for the obs in aeroval is broken in main-dev. The problem is in the method

ExperimentOutput._is_part_of_experiment()

jgliss:
The purpose of `ExperimentOutput._is_part_of_experiment()` is to make sure outdated json files are not added to the menu, when it is created. I see that this will also ignore map json files that make use of "web_interface_name". Should be easy to fix by testing whether the name is either key of the `ObsCollection` in `EvalSetup` ***or*** a valid web_interface_name.

PR #443, opened 2021-09-21
Attempt fix of #423

jgliss:


dulte (commented):
Looks good. Trying to run it now and see if it works. Will approve as soon as it is done

dulte (approved):
Looks good, and seems to work as intended

PR #445, opened 2021-09-21
Adds only trends statitics if add_trends

dulte:
Menu options for trend statistics will now only be made if add_trends is true. If user sets add_trends to true, but data is to short to make trends, empty menu options will show. So further check will have to be made


jgliss (approved):
Looks good as far as I can see :)

PR #454, opened 2021-09-27
Adds options for statistics.json for extended statistics

dulte:
R-space and R-time are now added to statistics.json. They have the option overall_only


jgliss (approved):
Looks good, I tested the AeroVal interface for a test run and R-Space and R-Temporal show up there, however they also show up in the menu on the "Evaluation" tab, but I guess this needs to be checked by  @AugustinMortier since I saw that you added "overall_only": true. So from my side this can be merged. But please follow up with @AugustinMortier on the frontend part.

PR #457, opened 2021-09-28
Set StatisticsSetup.MIN_NUM to 1 (before it was 3)

jgliss:


jgliss:
@dulte, this should hopefully display the bias maps correctly for the yearly averages. Can you check?

dulte (approved):
Looks good

PR #458, opened 2021-09-29
Adds new, general exception

dulte:
Makes a possible fix for crashing when reading EEA data.

### DIFF
--- a/pyaerocom/io/read_eea_aqerep_base.py
+++ b/pyaerocom/io/read_eea_aqerep_base.py
@@ -282,8 +282,8 @@ class ReadEEAAQEREPBase(ReadUngriddedBase):
                 # UTF-8 decoding error
                 if suffix == '.gz':
                     os.remove(f_out.name)
-                raise EEAv2FileError(
-                    f'Found corrupt file {filename}. consider deleteing it')
+                    raise EEAv2FileError(
+                        f'Found corrupt file {filename}. consider deleteing it')
 
             # create output dict
             if len(header) < max_file_index_to_keep:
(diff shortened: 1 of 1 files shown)


dulte:
I might have to add some test for this feature before merging

pyaerocom/io/read_eea_aqerep_base.py
```diff
@@ -282,8 +282,8 @@ def read_file(self, filename, var_name,
                 # UTF-8 decoding error
                 if suffix == '.gz':
                     os.remove(f_out.name)
-                    raise EEAv2FileError(
```
jgriesfeller, line 285: This is wrong
EEAv2FileError is independent from the .gz suffix. pyaerocom should not delete files. Never
The os.remove just deletes a temp file in case a zipped file was read (which is not the case on PPI, but when I develop at home)

pyaerocom/io/read_eea_aqerep_base.py
```diff
@@ -627,6 +627,11 @@ def read(self, vars_to_retrieve=None,
                 self.logger.warning('{} has TemporalResolutionError'.format(_file))
                 const.print_log.warning(f'{repr(e)}. Skipping file...')
                 continue
+            except Exception as e:
```
jgriesfeller, line 630: It might be better to just catch the UnicodeDecodeError error, but in principle I am fine with this change

jgriesfeller (changes requested):
just have a look at the comments. The first one needs to be addressed

pyaerocom/io/read_eea_aqerep_base.py
dulte: Yes, I agree, it is safer. I will change that and push a new version

jgriesfeller (approved):
approved

pyaerocom/io/read_eea_aqerep_base.py
dulte: Removed this indentation

jgliss:
Feel free to merge (I guess @jgriesfeller knows better if that is okay) :)

PR #460, opened 2021-09-30
Seasalt

dulte:
Sea salt pm 25, and more


jgliss:
@dulte can you check the failing tests first and make sure CI passes?

dulte:
It was the test function that wasn't updated with the new definition of sea salt. This is fixed now

jgliss:
@dulte we just merged  #410 into main-dev, could you add the updates to this branch before we merge?

dulte:
Pulled the newest updates from main-dev and pushed it to this branch

pyaerocom/data/emep_variables.ini
```diff
@@ -68,7 +68,8 @@ concrdn=SURF_ugN_RDN
 concso2=SURF_ug_SO2
 concso4=SURF_ug_SO4
 concss=SURF_ug_SS
-concsspm25=SURF_ug_SEASALT_F
+#concsspm25=SURF_ug_SEASALT_F
+concssf=SURF_ug_SEASALT_F
```
jgliss, line 72: I guess this is still concsspm25 in ebas_config.ini. Also you probably need to add the new variable concssf in variables.ini.

pyaerocom/io/read_mscw_ctm.py
```diff
@@ -140,6 +140,12 @@ def calc_concNtnh(concnh3,concnh4):
     concNtnh.attrs['units'] = 'ug N m-3'
     return concNtnh
 
+def calc_concsspm25(concssf, concssc, coarse_fraction=0.13):
+    concsspm25 = concssf + coarse_fraction*concssc
```
jgliss, line 144: Ah, then co-location works if you use concsspm25!

jgliss (approved):
Looks good (but you might need to add concssf in variables.ini, not sure if it exists)

jgliss (changes requested):
Actually, cleanup of glob_defaults needed before approval, sorry for the confusion.

dulte:
I think most of the duplicates from glob default should be gone now

jgliss (commented):
Perhaps you could add a test module for glob_defaults in tests/aeroval and add a test to make sure there no duplicates. Not important to do that now, but I think this should be tested before we merge.

dulte:
Dicts made with {} will only keep the last duplicate entry, so that as soon as var_web_info actually becomes a dict in python, there will be no duplicates, meaning that we can test for it. If we change {} to dict(), it will crash instead, which might be better

jgliss:
> Dicts made with {} will only keep the last duplicate entry, so that as soon as var_web_info actually becomes a dict in python, there will be no duplicates, meaning that we can test for it. If we change {} to dict(), it will crash instead, which might be better

Sounds good!

jgliss (approved):
Looks good

PR #461, opened 2021-09-30
rewrote the file reading so that the entire file is read into memory …

jgriesfeller:
so that there's one place to determine if the file is UTF-8 or UTF-16 encoded or a complete failure. Saves a lot of if statements for the price of temporary RAM usage.

### DIFF
--- a/pyaerocom/io/read_eea_aqerep_base.py
+++ b/pyaerocom/io/read_eea_aqerep_base.py
@@ -274,6 +274,9 @@ class ReadEEAAQEREPBase(ReadUngriddedBase):
         else:
             read_filename = filename
 
+        # input files can be either UTF-8 or UTF-16 encoded
+        # try both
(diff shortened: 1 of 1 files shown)


jgliss:
Does this make #459 obsolete,  @jgriesfeller and @dulte?

dulte:
Yes, this makes the previous one obsolete. I can remove it

pyaerocom/io/read_eea_aqerep_base.py
```diff
@@ -455,9 +448,9 @@ def _read_metadata_file(self, filename=None):
             bad_line_arr = []
             for line in f:
                 rows = line.rstrip().split('\t')
-
-                if len(rows) < 24:
```
jgriesfeller, line 459: removed constant usage to make the code more roibust

dulte (approved):
Looks good

PR #462, opened 2021-09-30
Treat warnings as errors

avaldebe:
Deal with warnings raised while testing
- [x] ignore warnings raised when importing external modules
- [x] ignore warnings raised on external modules
- [x] address `DeprecationWarnings` raised on pyaercom modules
- [x] address warnings raised on internal modules (ignore when large refactor would be needed)
- [x] new warnings to raise an error (configure `filterwarnings`)

### DIFF
--- a/pyaerocom/griddeddata.py
+++ b/pyaerocom/griddeddata.py
@@ -15,7 +15,7 @@ from pathlib import Path
 import xarray as xr
 
 from pyaerocom import const, logger, print_log
-from pyaerocom._warnings_management import filter_warnings
+from pyaerocom._warnings_management import ignore_warnings
 from pyaerocom.helpers_landsea_masks import load_region_mask_iris
 from pyaerocom.tstype import TsType
 from pyaerocom.exceptions import (CoordinateError,
@@ -1759,8 +1759,8 @@ class GriddedData(object):
             data = self._resample_time_iris(to_ts_type)
         return data
 
-    @filter_warnings(const.FILTER_IRIS_WARNINGS,
-                     messages=["Using DEFAULT_SPHERICAL_EARTH_RADIUS."])
+    @ignore_warnings(const.FILTER_IRIS_WARNINGS,
+                     messages="Using DEFAULT_SPHERICAL_EARTH_RADIUS.")
     def calc_area_weights(self):
         """Calculate area weights for grid"""
         if not self.has_latlon_dims:
--- a/pyaerocom/griddeddata.py
+++ b/pyaerocom/griddeddata.py
@@ -678,7 +678,7 @@ class GriddedData(object):
 
 
     @ignore_warnings(
-        True, UserWarning, messages="Ignoring netCDF variable '.*' invalid units '.*'"
+        True, UserWarning, "Ignoring netCDF variable '.*' invalid units '.*'"
     )
     def load_input(self, input, var_name=None, perform_fmt_checks=None):
         """Import input as cube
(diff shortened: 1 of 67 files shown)


pyaerocom/griddeddata.py
```diff
@@ -676,6 +676,10 @@ def register_var_glob(self, delete_existing=True):
             f'to add this variable to pyaerocom/data/variables.ini.')
         return vardef
 
+
```
jgriesfeller, line 679: not sure if I want this warning to disappear. Might make error search (why data was not read more complicated)

pyaerocom/plot/mapping.py
```diff
@@ -556,7 +562,7 @@ def plot_nmb_map_colocateddata(coldata, in_percent=True, vmin=-100,
         valid = ~stacked.isnull()
         coords = stacked.latlon[valid].values
         lats, lons = list(zip(*list(coords)))
-        data = stacked.data[valid]
+        data = stacked.data[tuple(valid)]
```
jgriesfeller, line 565: Why `tuple(valid)`?

tests/aeroval/test_modelmaps_helpers.py
```diff
@@ -1,5 +1,3 @@
-import pytest
-from ..conftest import data_tm5
```
jgriesfeller, line 2: does this work if one is not importing `data_tm5`

tests/test_stationdata.py
```diff
@@ -1,15 +1,23 @@
+from contextlib import nullcontext as does_not_raise_exception
+
 import numpy as np
-import pytest
 import pandas as pd
-from xarray import DataArray
```
jgriesfeller, line 4: This looks wrong since it's used here:
https://github.com/metno/pyaerocom/blob/064d539f4d7dbc9ff390b89aeef1ad7b7031e6d7/tests/test_stationdata.py#L261

jgriesfeller (changes requested):
I think there are a couple of errors in there, and I had a question

tests/aeroval/test_modelmaps_helpers.py
avaldebe: `data_tm5` is a fixture, so it is not used here. `pytest` will use the fixture when running the test, but the test_module does not need to import it.

pyaerocom/plot/mapping.py
avaldebe: `valid` can not be a list, it needs to be a tuple in order to avoid a numpy FutureWarning

pyaerocom/griddeddata.py
avaldebe: Then we need to ignore it on the test, or in the pytest configuration

tests/test_stationdata.py
avaldebe: I `isort`ed the file. The import is still there, after `from matplotlib.axes import Axes`.

tests/aeroval/test_modelmaps_helpers.py
avaldebe: ... and pytest would fail at the test discovery stage if an import would be missing

pyaerocom/plot/mapping.py
avaldebe: this was the warning message, shown on CI

```
tests/plot/test_mapping.py::test_plot_nmb_map_colocateddata4D
  /home/runner/work/pyaerocom/pyaerocom/pyaerocom/plot/mapping.py:559: FutureWarning: Using a non-tuple sequence for multidimensional indexing is deprecated; use `arr[tuple(seq)]` instead of `arr[seq]`. In the future this will be interpreted as an array index, `arr[np.array(seq)]`, which will result either in an error or a different result.
```

avaldebe:
@jgriesfeller 

are you satisfied with the answers?
Is there is something specific I need to do in order to get this PR trough?

pyaerocom/plot/mapping.py
jgriesfeller: ok

tests/test_stationdata.py
jgriesfeller: Then I have misread the file, sorry for that.

jgriesfeller (approved):

jgriesfeller:
approved

PR #463, opened 2021-09-30
add ETEX1 observations

avaldebe:


magnusuMET:
New readers should preferably use pyaro-readers

Issue #464, opened 2021-10-01
Procseeing of new combined ungridded obs datasets does not work in new AeroVal tools


jgliss:
E.g. before, new observation datasets could be defined like this as obs entry in the config file:

```python
'EEA-ox-rural'       :   dict(obs_id='EEA-ox-rural',
                              obs_vars=['vmrox'],
                              obs_type='ungridded',
                              obs_vert_type='Surface',
                              web_interface_name='EEA-rural',
                              ts_type='daily',
                              obs_merge_how={
                                  'vmrox' : 'eval',},
                              obs_aux_requires = {
                                  'vmrox' : {
                                      'EEAAQeRep.v2' : ['vmro3','vmrno2'],
                                      }},
                              obs_aux_funs = {
                                  'vmrox' :
                                      # variables used in computation method need to be based on AeroCom
                                      # units, since the colocated StationData objects (from which the
                                      # new UngriddedData is computed, will perform AeroCom unit check
                                      # and conversion)
                                      '(EEAAQeRep.v2;vmro3+EEAAQeRep.v2;vmrno2)'
                                      },
                                  obs_aux_units = {
                                      'vmrox' : 'nmol mol-1'
                                      }
                              )
```

This does not work anymore and is likely due to the fact, that the corresponding logic to register new obs datasets in `ObsConfig` class was removed and not added anywhere else.

PR #465, opened 2021-10-01
Enable processing and dynamic registration of new ungridded datasets in AeroVal

jgliss:
See #464


pyaerocom/io/read_mscw_ctm.py
```diff
@@ -140,12 +141,34 @@ def calc_concNtnh(concnh3,concnh4):
...
+        1000*10,
+        288.15, #15 deg celcius
+        "ug m-3",
+        46.0055, # g/mol NO2
+        to_unit="nmol mol-1"
+    )
+
+    vmrox = vmrno2 + vmro3
+    vmrox.attrs["units"] = "nmol mol-1"
+    return vmrox
```
dulte, line 170: This is not tested by me yet, due to problems with the colocation (see message)

jgliss:
@dulte, even though your EEA evaluation looks still strange I think we should merge this and then further debug from main-dev. I added some tests based on AERONET for the merging (AODtot = AODf + AODc) and this seems to work fine. As discussed, perhaps there is something off with the EEA reader, which we should investigate. Perhaps you could make an issue that adresses the problem you encountered with EEA vmrox=vmrno2+vmro3.

Could you please review the recent changes I did here?

pyaerocom/griddeddata.py
```diff
@@ -2021,16 +2021,19 @@ def aerocom_savename(self, data_id=None, var_name=None,
...
+            if self.metadata['vert_code'] is not None:
+                vert_code=self.metadata['vert_code']
+            else:
+                try:
+                    from pyaerocom.io.fileconventions import FileConventionRead
+                    f = self.from_files[0]
+                    fconv = FileConventionRead().from_file(f)
+                    vert_code = fconv.get_info_from_file(f)['vert_code']
+                except Exception:
+                    pass
```
dulte, line 2033: Maybe write someting to logger before passing?

jgliss: Good point. I think this method tries to be too smart, I'll revise in #456

dulte (approved):
Looks good :)

Issue #444, opened 2021-09-21
aeroval: setting the min # of years for the trends makes the analysis crash


jgriesfeller:
I am trying to run the following aeroval 
[config file (metno internal!):](https://gitlab.met.no/aeroval/config/-/blob/griesie_dev/config_files/cfg_c3s_testing_cp.py)

it crashes with the following error message using main-dev @  commit c51049a6:
```  var = nanvar(a, axis=axis, dtype=dtype, out=out, ddof=ddof,
Traceback (most recent call last):
  File "/home/jang/data/aeroval-local-web/pyaerocom_config/config_files/cfg_c3s_testing_cp.py", line 272, in <module>
    res = ana.run()
  File "/home/jang/anaconda.2018/envs/pyaerocom-main-dev/lib/python3.9/site-packages/pyaerocom-0.12.0.dev2-py3.9.egg/pyaerocom/aeroval/experiment_processor.py", line 336, in run
    self._run_single_entry(model_name, obs_name, var_list)
  File "/home/jang/anaconda.2018/envs/pyaerocom-main-dev/lib/python3.9/site-packages/pyaerocom-0.12.0.dev2-py3.9.egg/pyaerocom/aeroval/experiment_processor.py", line 285, in _run_single_entry
    self.make_json_files(files_to_convert)
  File "/home/jang/anaconda.2018/envs/pyaerocom-main-dev/lib/python3.9/site-packages/pyaerocom-0.12.0.dev2-py3.9.egg/pyaerocom/aeroval/experiment_processor.py", line 105, in make_json_files
    self.coldata_to_json(file)
  File "/home/jang/anaconda.2018/envs/pyaerocom-main-dev/lib/python3.9/site-packages/pyaerocom-0.12.0.dev2-py3.9.egg/pyaerocom/aeroval/experiment_processor.py", line 37, in coldata_to_json
    return engine.run(coldata)
  File "/home/jang/anaconda.2018/envs/pyaerocom-main-dev/lib/python3.9/site-packages/pyaerocom-0.12.0.dev2-py3.9.egg/pyaerocom/aeroval/coldatatojson_engine.py", line 1413, in run
    hm_all = _process_heatmap_data(data, regnames, use_weights,
[...]

PR #470, opened 2021-10-07
small update to make the trends calculation work

jgriesfeller:
small PR to address #444. Untested with satellite data so far, but needed to make the trends work with model data.


jgriesfeller:
This PR does not cover the problem with satellite data as that crashes [here](https://github.com/metno/pyaerocom/blob/a45afeb4cc08da125c248a1a5f381a9f6671542e/pyaerocom/trends_helpers.py#L217)

jgliss:
Particularly, it should be investigated why this only leads to a crash in some cases, and in others it works fine.

PR #483, opened 2021-10-12
correct validity flag assignment

jgriesfeller:
correcting yesterdays PR


jgliss:
@jgriesfeller probably good to delete the branch after merging.

PR #493, opened 2021-10-14
some iris 3.1.0 adjustments

jgriesfeller:
1st try for an iris 3.1.0 patch


jgriesfeller:
I have removed the limitation of the iris version in setup.py. CI still works :-)

Issue #494, opened 2021-10-15
AERONET reading broken after bug fix in release v0.12.0


jgliss:
For AERONET version 3, latitude, longitude and altitude coordinates are provided for each measurement point. Thus, the `StationData` objects returned by the "read_file" methods of the AERONET readers contain lists of longitude, latitude and altitude arrays and not floats. 

In `ReadAeronetBase.read` the station metadata is then retrieved using `StationData.get_meta` with input args:

force_single_value=True
quality_check=True

The "quality check" is supposed to make sure that, if the coordinates are not floats but arrays or lists, that they match within a certain (hard coded) tolerance range. Due to a bug in former pyaerocom versions, this "quality check" was not done and AERONET sites with too variable coordinate arrays were imported regardless. 

Now, this "quality check" has been fixed in yesterdays release of v0.12.0 (now it is checked that the coordinates are within the specified tolerance ranges) and as a result, the reading of AERONET crashes due to the ValueError raised by  `StationData.get_meta` and the AERONET reader does not capture this Exception. This affects 4 out of the 1294 sites have an issue on that end.

PR #495, opened 2021-10-15
Add AeroVal config documentation example

jgliss:
This PR adds an AeroVal example configuration script in the ReadTheDocs page, that explains the main setup and options.

It also fixes a few issues that were detected after release of v0.12.1 when trying to run the example setup, e.g. #494.


jgliss:
@dulte and @jgriesfeller and @AugustinMortier please be aware of the changes here, I found some bugs and inconsistencies in the variable handling for the output files when creating the example config. The problem is that pyaerocom was using the AeroCom variable name for the output files and not the input variable name, which resulted in AeroVal not displaying "ang4487csaer" from NorESM2 in the model maps plots, because "ang4487csaer" is an alias for "ang4487aer".

I think this affects only cases where alias variables are used, and it is good that this was detected now, since this variable mappings is always a mess.

PR #511, opened 2021-10-28
Create CONTRIBUTING.md

hansbrenna:


avaldebe:
we should move Jonas and Hans to the end of the Team members section as previous members

Issue #515, opened 2021-11-04
Ungridded reading fails on moving stations; should just ignore them as Aeronet does


jgriesfeller:
While running latest cams84 analysis I discovered that the station locations seem to be not the same anymore:
```
 File "/home/jang/data/Python3/pyaerocom_new/pyaerocom/stationdata.py", line 582, in merge_meta_same_station
    raise CoordinateError(f'Station coordinates differ by '
pyaerocom.exceptions.CoordinateError: Station coordinates differ by more than 0.1 km.
```

https://github.com/metno/pyaerocom/blob/064d539f4d7dbc9ff390b89aeef1ad7b7031e6d7/pyaerocom/stationdata.py#L59

Allowing a distance of just 100m to consider a station the same is a bit too less and we allow much more at the Aeronet reading. 
I have not looked into the data itself really at this point, but for now I will try with 1km for now.

jgriesfeller:
It turned out that station moved `202.02240165266448km`

jgriesfeller:
In principle this kind of error should not stop the analysis but just skip the particular station. With Aeronet reading that is the case, but is seems to be handled differently

jgriesfeller:
with vmro3 one station moved `28.747274542372658km`

jgriesfeller:
with vmrno2 one station moved `202.02240165266448km`

jgriesfeller:
Solved the immediate problem by excluding stations where the reader failed:
```
AIRNOW_SITE_FILTER = dict(
    station_name=['IEPA TRAILER', 'Lebanon', 'Rahway PM', 'nan', 'COOK COUNTY TRAILER', 'Cedar Creek',
                  'HOLLAND', 'WATER PLANT'], negate='station_name')
```

jgriesfeller:
For Marco Polo ignoring these stations makes the reading work as well:
```
MARCO_POLO_SITE_FILTER = dict(station_name=['三水厂','人民公园','体育馆','党校','八中','农业园区','农科所','区政府',
                                            '园林处','奥体中心','审计局','州监测站','市中心','市二中','市四中','市图书馆',
                                            '市委','市委党校','市政府','市林业局','市气象局','市环保局','市环保站',
                                            '市环境监测站','市疾控中心','市监测站','市站','师范学院','开发区',
                                            '开发区管委会','技工学校','技术学院','来冶招待所','林科所','气象局',
[...]

PR #517, opened 2021-11-04
make aeroval work for the case when the model data variable has been aliased automatically

jgriesfeller:
Use case: Model uses old aerocom variable names (e.g. sconcpm10 instead of concpm10). Model data is read, but with the old variable naming. In this case everything worked except for the `menu.json` creation. Because I don't oversee all the options in the config file, and don't have a test case for all possibilities regarding the options of the model reading (like `model_rename_vars`, `model_add_vars`, or `model_use_vars`.

But I do think that the old variable names should be usable since there is a lot of model data in the aerocom database that use that naming scheme


pyaerocom/stationdata.py
```diff
@@ -56,7 +56,8 @@ class StationData(StationMetaData):
 
     #: maximum numerical distance between coordinates associated with this
     #: station
-    _COORD_MAX_VAR = 0.1 #km
+    # _COORD_MAX_VAR = 0.1 #km
+    _COORD_MAX_VAR = 1.0 #km
```
jgriesfeller, line 60: This is just a little relaxation that helped a little to read several years of Airnow and MarcoPolo data

pyaerocom/aeroval/experiment_output.py
```diff
@@ -674,6 +674,9 @@ def _check_ovar_mvar_entry(self, mcfg, mod_var, ocfg, obs_var):
             if obs_var in mrv and mrv[obs_var] == mod_var:
                 # obs variable is in model_rename_vars
                 return True
+            elif mod_var in get_aliases(obs_var):
```
jgriesfeller, line 677: not sure if I understand https://github.com/metno/pyaerocom/blob/ac9978576c35316ee4224164e7ae1508bed475e8/pyaerocom/aeroval/experiment_output.py#L632-L658 entirely. It might be that L632 should be 
```
mvar_aliases = get_aliases(obs_var)
```

jgriesfeller: because https://github.com/metno/pyaerocom/blob/ac9978576c35316ee4224164e7ae1508bed475e8/pyaerocom/data/aliases.ini
does not define aliases for the sconc* variables (but only sconc* being an alias for conc*) 
https://github.com/metno/pyaerocom/blob/ac9978576c35316ee4224164e7ae1508bed475e8/pyaerocom/aeroval/experiment_output.py#L632 does not make any sense at all. This really should be
```
mvar_aliases = get_aliases(obs_var)
```

pyaerocom/data/aliases.ini
```diff
@@ -36,7 +36,7 @@ emidust = emidu
 loaddust = loaddu
 wetdust = wetdu
 drydust = drydu
-sconcdust = sconcdu
```
jgriesfeller, line 39: sconcdust is an old aerocom phase 2 variable name and should not be used in pyaerocom. Use concdust instead.

pyaerocom/aeroval/experiment_output.py
jgriesfeller: due to this comment
https://github.com/metno/pyaerocom/blob/1538aa14d850c7862fe35c7ecbe4411a4f53218c/pyaerocom/aeroval/experiment_output.py#L666-L669

I will not change https://github.com/metno/pyaerocom/blob/1538aa14d850c7862fe35c7ecbe4411a4f53218c/pyaerocom/aeroval/experiment_output.py#L632 for now

The automatic use case and the CAMS61 setup works according to @hansbrenna . Converting the draft PR to a PR therefore

jgriesfeller:
I added with one of the changes another conc* variable. I will update the test

jgriesfeller:
slightly related to #515

dulte (approved):
Looks good :)

PR #526, opened 2021-11-22
Split tests with/without exceptions

avaldebe:
follow up to  #397

### DIFF
--- a/pyaerocom/_lowlevel_helpers.py
+++ b/pyaerocom/_lowlevel_helpers.py
@@ -839,9 +839,9 @@ def dict_to_str(dictionary, indent=0, ignore_null=False):
 
 def str_underline(title: str, indent: int = 0):
     """Create underlined string"""
-    lenght = indent + len(title)
+    length = indent + len(title)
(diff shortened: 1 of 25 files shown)


pyaerocom/_lowlevel_helpers.py
```diff
@@ -837,11 +837,11 @@ def dict_to_str(dictionary, indent=0, ignore_null=False):
     return s
 
 
-def str_underline(s, indent=0):
+def str_underline(title: str, indent: int = 0):
     """Create underlined string"""
-    s = indent * " " + f"{s}\n"
-    s += indent * " " + f"{len(s) * '-'}"
-    return s
+    lenght = indent + len(title)
```
jgriesfeller, line 842: lenght should probably called length, but the variable name in principle does not matter

avaldebe: done

jgriesfeller (approved):
I looked only at a few files thoroughly, but since the CI does not throw errors I trust you that everything is working

PR #527, opened 2021-11-24
[draft] make Aeronet reading more fault tolerable for encoding errors in the data fi…

jgriesfeller:
…les; make reading of .gz data files possible


pyaerocom/io/read_aeronet_sunv3.py
```diff
@@ -124,91 +129,118 @@ def read_file(self, filename, vars_to_retrieve=None,
...
+                shutil.copyfileobj(f_in, f_out)
+            filename = f_out.name
+            f_out.close()
+
+        try:
+            with open(filename, 'rt') as in_file:
+                lines = in_file.readlines()
+        except UnicodeDecodeError:
+            with open(filename, 'rt', encoding='ISO-8859-1') as in_file:
+                lines = in_file.readlines()
```
jgriesfeller, line 147: @avaldebe do you know a more elegant way of putting a context manager in a try block?
The goal would be to not read the entire file first. But the encoding error can occur anywhere, so one would need to change the encoding on the fly. 
Any suggestions (but don't think too much about it). 
But this is the 2nd obs network whose data files can have different encodings (the other is EEA)

avaldebe: google suggested me this solutions:

[Check for BOM markers](https://unicodebook.readthedocs.io/guess_encoding.html#check-for-bom-markers)

``` python
from codecs import BOM_UTF8, BOM_UTF16_BE, BOM_UTF16_LE, BOM_UTF32_BE, BOM_UTF32_LE

BOMS = (
    (BOM_UTF8, "UTF-8"),
    (BOM_UTF32_BE, "UTF-32-BE"),
    (BOM_UTF32_LE, "UTF-32-LE"),
    (BOM_UTF16_BE, "UTF-16-BE"),
    (BOM_UTF16_LE, "UTF-16-LE"),
)

def check_bom(data):
    return [encoding for bom, encoding in BOMS if data.startswith(bom)]
```

[Is UTF-8?](https://unicodebook.readthedocs.io/guess_encoding.html#is-utf-8)

``` python
def isUTF8(data):
    try:
        data.decode('UTF-8')
    except UnicodeDecodeError:
        return False
    else:
        return True
```

For any of these to work, you would need to read the file as binary and decode on a second step.
In any case you need to read the file in one go.

Alas, I do not see a clean way to guess the encoding without reading the file.
Is this a problem big enough to depend on a 3rd party library?

jgriesfeller: no I don't want some external library dependency for this. The BOM approach also reads the beginning of the file as does my code (the crash happens in the file header), so there's no real gain.
I therefore think I leave it as it is at least for now. Only the all points Aeronet file can be big, but not gig enough to really bother.

jgriesfeller:
rebased on current main-dev...
Not sure if I did it right...

jgriesfeller:
replaced by #538

PR #530, opened 2021-12-06
use plain dicts

### LINKED ISSUE #532: idiomatic use of dictionary keys
the pyearocom code base has plenty of unnecessary calls to the dictionary keys of the form

``` python
my_list = list(my_dict.keys())

for key in my_dict.keys():
    ...
```

which are equivalent to 

``` python
my_list = list(my_dict)

for key in my_dict:
    ...
```

and a few instances of the following hack

``` python
unique = list(dict.fromkeys(my_list))
```

which can be clearly written as

``` python
unique = list(set(my_list)))
```

### LINKED ISSUE #533: use dict instead of collections.OrderedDict
The pyaerocom makes extensive use of `OrderedDict` from the [collections module](https://docs.python.org/3/library/collections.html#collections.OrderedDict).

The Python 3.6 `dict` implementation had the side effect of preserving insertion order. This side effect became a language feature from Python 3.7.

As far as I can tell we can replace all the `OrderedDict` with plain `dict`

avaldebe:
- close #532
- close #533

### DIFF
--- a/pyaerocom/_lowlevel_helpers.py
+++ b/pyaerocom/_lowlevel_helpers.py
@@ -298,12 +298,12 @@ class ListOfStrings(FlexList):
 
 
 class DictStrKeysListVals(Validator):
-    def validate(self, val):
+    def validate(self, val: dict):
         if not isinstance(val, dict):
             raise ValueError(f"need dict, got {val}")
-        elif not all([isinstance(x, str) for x in val.keys()]):
+        if any(not isinstance(x, str) for x in val):
             raise ValueError(f"all keys need to be str type in {val}")
-        elif not all([isinstance(x, list) for x in val.values()]):
+        if any(not isinstance(x, list) for x in val.values()):
             raise ValueError(f"all values need to be list type in {val}")
         return val
 
--- a/pyaerocom/_lowlevel_helpers.py
+++ b/pyaerocom/_lowlevel_helpers.py
@@ -410,7 +410,7 @@ class BrowseDict(MutableMapping):
         return _class_name(self)
 
     def keys(self):
-        return list(self.__dict__.keys()) + self.ADD_GLOB
+        return list(self.__dict__) + self.ADD_GLOB
 
     def _get_glob_vals(self):
         return [getattr(self, x) for x in self.ADD_GLOB]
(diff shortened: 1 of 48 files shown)


pyaerocom/_lowlevel_helpers.py
```diff
@@ -410,7 +410,7 @@ def _class_name(self):
         return _class_name(self)
 
     def keys(self):
-        return list(self.__dict__.keys()) + self.ADD_GLOB
+        return list(self.__dict__) + self.ADD_GLOB
```
jgriesfeller, line 413: isn't `self.__dict__` a dictionary? [The documentation](https://docs.python.org/3/library/stdtypes.html#object.__dict__) is suggesting that. Just to be sure

pyaerocom/io/readgridded.py
```diff
@@ -364,15 +363,15 @@ def _get_vars_provided(self):
         _vars = []
         _vars.extend(self.vars_filename)
 
-        for aux_var in self.AUX_REQUIRES.keys():
+        for aux_var in self.AUX_REQUIRES:
             if "*" in aux_var:
                 continue
-            elif not aux_var in _vars and self.check_compute_var(aux_var):
+            if not aux_var in _vars and self.check_compute_var(aux_var):
```
jgriesfeller, line 369: Is there a special reasom you got rid of the `elif` here?

avaldebe: IMO, an `if` clause ending in a `continue`, `break`, `return` or `raise` is a guarding clause.
The same applies to the `elif` clause here.

Think about the following function:

``` python
def float_division(numerator: float, denominator: float) -> float:
    if not isinstance(numerator, float):
        raise TypeError("numerator should be a float")
    elif not isinstance(denominator, float):
        raise TypeError("denominator should be a float")       
    elif denominator == 0:
        raise ValueError("dude you can not divide by zero")
    else:
        return numerator/denominator
```

I find it more readable with the `if/elif` as a guarding clauses, indicating the anomalous cases,
and the "expected use" in the body of the function instead of an `else` clause

``` python
def float_division(numerator: float, denominator: float) -> float:
    if not isinstance(numerator, float):
        raise TypeError("numerator should be a float")
    if not isinstance(denominator, float):
        raise TypeError("denominator should be a float")       
    if denominator == 0:
        raise ValueError("dude you can not divide by zero")

    return numerator/denominator
```

jgriesfeller (approved):
good to go

PR #531, opened 2021-12-06
use importlib.resources/metadata instead of pyaeorocom.__dir__/__version__

avaldebe:
## replace `__dir__`

Currently `pyaerocom.__dir__` is used to provide the package installation path, which in turn is used to reach the configuration files in `pyaerocom.data`. [importlib.resources](https://docs.python.org/3/library/importlib.html#module-importlib.resources) was introduced on Python 3.7 for exactly this purpose.

With `importlib.resources.path` is possible to rewrite the following ini parsing 

https://github.com/metno/pyaerocom/blob/a4a35d49e031021fb855ef86a39e9b0383d02204/pyaerocom/io/fileconventions.py#L346-L354

as

``` python
from importlib import resources
[...]
    def import_default(self, name: str):
        """Checks and load default information from database"""
        
        conf_reader = ConfigParser()
        with resources.path(f"pyaerocom.data", "file_conventions.ini") as path:
            conf_reader.read(path)
```

## refactor `__version__`

[importlib.metadata](https://docs.python.org/3/library/importlib.metadata.html) was introduced on Python 3.8, and provides a simple access to the version of an installed package as follows:

In combination with the changes outlined on the previous section, one could replace the current `__dir__`/`__version__`
implementation

https://github.com/metno/pyaerocom/blob/dd482a64fea55d9d49e8714ccf3e84167b1f99dc/pyaerocom/_init_helpers.py#L67-L87

with

``` python
from importlib import metadata

__version__ = metadata.version("pyaerocom")
```

## Breaking changes

`search_names` and `get_all_names` from `pyaerocom.io.helpers` will be removed.
`search_names` saves all the available model IDs on `pyarocom_install_path/data/names.txt`
and `get_all_names` reads file and return the model IDs.

https://github.com/metno/pyaerocom/blob/dd482a64fea55d9d49e8714ccf3e84167b1f99dc/pyaerocom/io/helpers.py#L262-L300
https://github.com/metno/pyaerocom/blob/dd482a64fea55d9d49e8714ccf3e84167b1f99dc/pyaerocom/io/helpers.py#L303-L316

As far as I can tell this functions were introduced 4 years ago, and are currently not used anywhere on `pyaerocom`.
If such functionality is required, it should written to `~/MyPyaerocom`  or `~/.cache/pyaerocom`, not on the packages installation path.


jgriesfeller (approved):
looks good!

PR #535, opened 2021-12-07
one logger per module

### LINKED ISSUE #534: one logger per module
pyaerocom has 2 different loggers, `logger` and `print_log`. Both loggers print to the terminal, but with different formats and logging levels.

They are available from the root package and from the `const` configuration

``` python
from pyaerocom import const, logger, print_log

assert logger is const.logger
assert print_log is const.print_log
```

Some modules take the loggers from `pyaerocom.const`, others import directly from `pyaerocom`, and some do both.

These introduces unnecessary coupling between the modules.

Instead, I propose to configure the logger for the package one as follows:

``` python
# pyaerocom/_init_helpers.py or pyaerocom/__init__.py

import logging
from logging.handlers import TimedRotatingFileHandler

logger = logging.getLogger("pyaerocom")

# prints to screen INFO messages or higher
console_handler = logging.StreamHandler()
console_formatter = logging.Formatter("%(message)s")
console_handler.setFormatter(console_formatter)
console_handler.setLevel(logging.INFO)
logger.addHandler(console_handler)

# write all messages to pyaerocom.log
# old log files are keep the up to 14 days
file_handler = TimedRotatingFileHandler("pyaerocom.log", when="D", backupCount=14)
file_formatter = logging.Formatter(
     "%(asctime)s:%(name)s:%(levelname)s:%(message)s",
    datefmt="%F %T",
)
file_handler.setFormatter(file_formatter)
file_handler.setLevel(logging.DEBUG)
logger.addHandler(file_handler)
```

Then any module that needs a logger can get its own logger, as follows:

``` python
# pyaerocom/internal/module.py

import logging

logger = logging.getLogger(__name__)

logger.debug("this message will be written to file")
logger.info("this message will be written to file and printed to the console")
```

There the logger does not need its own configuration, because it inherits the configuration from the previous step

### LINKED ISSUE #537: do not log deprecation warnings, use warnings.warn instead
logging deprecation warnings, like the following, hides important information deep in the logs

https://github.com/metno/pyaerocom/blob/14d0257c9a39ac09038d9ce7b1de45e7b2638015/pyaerocom/stationdata.py#L1172-L1179

`pytest` can be configured to catch warnings like the following and treat them as errors,
so this kind of deprecation warnings can be acted upon.

``` python
        warnings.warn(
            "This method was renamed to resample_time as a means "
            "of harmonisation with GriddedData and ColocatedData",
            DeprecationWarning,
            stacklevel=2,
        )
```

avaldebe:
- close #534 
- close #537

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,3 +1,5 @@
+pyaerocom.log*
+
 # Byte-compiled / optimized / DLL files
 __pycache__/
 _cache/
--- a/pyaerocom/__init__.py
+++ b/pyaerocom/__init__.py
@@ -1,5 +1,5 @@
 # isort:skip_file
-from ._init_helpers import _init_supplemental, logger, change_verbosity
+from ._init_helpers import _init_supplemental, change_verbosity
 
 __version__, __dir__ = _init_supplemental()
 
--- a/pyaerocom/__init__.py
+++ b/pyaerocom/__init__.py
@@ -1,5 +1,5 @@
 # isort:skip_file
-from ._init_helpers import _init_supplemental, _init_logger, LOGLEVELS, change_verbosity
+from ._init_helpers import _init_supplemental, _init_logger, change_verbosity
 
 __version__, __dir__ = _init_supplemental()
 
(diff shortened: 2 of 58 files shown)


jgriesfeller (approved):
Good to be merged

Issue #518, opened 2021-11-11
3hourly as a basis for Aeronet colocation


MichaelSchulzMETNO:
The basis for the IFS evaluation is so far daily. However, 3hourly data from Aeronet and IFS are available. Also - "daily averages" can be difficult and from different days eg near 180 degree longitude, in particula East Asia & Pacific, since data are on UTC. SDS-WAS was also using 3 hourly data for the dust evaluation, and the copernicus evaluation server as well. 

In mid term it would be good to explore if co-location could be done on 3hourly basis.

jgriesfeller:
Higher temporal resolution than daily Aeronet data (called all point data) has been used before (averaged to daily though). So the data is there.

According to https://github.com/metno/pyaerocom/blob/a483626c5e710e1dff63f68a2448a2c19ffc6400/pyaerocom/io/read_aeronet_sunv3.py#L28-L31

the all point data is also supported by the reader. But I don't think that anybody has ever used it since it's not mentioned in https://github.com/metno/pyaerocom/blob/a483626c5e710e1dff63f68a2448a2c19ffc6400/pyaerocom/io/read_aeronet_sunv3.py#L36-L37

I will try to use it and see if I get something out of it

jgriesfeller:
Is there any 3hourly model data to compare btw?

jgriesfeller:
I found some suitable 3hourly data within the aerocom data base _(CHAM6-SALSA_GLOFIR2)_
judging from this partial test, I would say that using 3hourly data should be possible in aeroval already
![Screenshot from 2021-11-24 12-18-37](https://user-images.githubusercontent.com/10772065/143229177-44d0ea67-aa57-41d4-879a-61b51df3c991.png)

Unfortunately there are problems with in the Aeronet all points file's file encoding (not always compatible to Python's default file encoding of UTF-8). 
Examples:
```
(base) jang@pc5378:.../aerocom1/AEROCOM_OBSDATA$ file /lustre/storeA/project/aerocom/aerocom1/AEROCOM_OBSDATA/AeronetSunV3Lev2.0.AP/renamed/19930101_20211120_Aras_de_los_Olmos.lev20
/lustre/storeA/project/aerocom/aerocom1/AEROCOM_OBSDATA/AeronetSunV3Lev2.0.AP/renamed/19930101_20211120_Aras_de_los_Olmos.lev20: ISO-8859 text, with very long lines
[...]

PR #538, opened 2021-12-13
make the AERONET reading more fault tolerable; add possibility to read gzipped files

jgriesfeller:
Basically a new try for a fix of #518 done because I messed up the branch after @avaldebe merged the source run through black and isort.
No different functionality compared to #527

### DIFF
--- a/pyaerocom/exceptions.py
+++ b/pyaerocom/exceptions.py
@@ -21,9 +21,11 @@
 Module containing pyaerocom custom exceptions
 """
 
+
 class AeronetReadError(IOError):
     # Aeronet reading failed somehow
     pass
+
 
 class AeroValConfigError(ValueError):
     pass
--- a/pyaerocom/exceptions.py
+++ b/pyaerocom/exceptions.py
@@ -21,6 +21,9 @@
 Module containing pyaerocom custom exceptions
 """
 
+class AeronetReadError(IOError):
+    # Aeronet reading failed somehow
+    pass
 
 class AeroValConfigError(ValueError):
     pass
--- a/tests/io/test_read_aeronet_sunv3_all_points.py
+++ b/tests/io/test_read_aeronet_sunv3_all_points.py
@@ -1,6 +1,6 @@
 #!/usr/bin/env python3
 """
-Created on Mon Jul  9 14:14:29 2018
+Created on 20211214
 """
 import os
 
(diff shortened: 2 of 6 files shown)


tests/io/test_read_aeronet_sunv3_all_points.py
```diff
@@ -0,0 +1,52 @@
...
+def test_read_add_common_meta(reader):
+    files = reader.files
+    data = reader.read("od550aer", files=files, common_meta={"bla": 42})
+    assert all("bla" in x for x in data.metadata.values())
+
+
+if __name__ == "__main__":
+    import sys
+
+    pytest.main(sys.argv)
```
avaldebe, line 52: do not add unnecessary entry points, use the IDE to run specific tests

tests/io/test_read_aeronet_sunv3_all_points.py
```diff
@@ -0,0 +1,52 @@
+#!/usr/bin/env python3
+"""
+Created on 20211214
+"""
```
avaldebe, line 4: not needed, git knows when the file was created and who did it

tests/io/test_read_aeronet_sunv3_all_points.py
```diff
@@ -0,0 +1,52 @@
...
+import os
+
+import numpy as np
+import numpy.testing as npt
```
avaldebe, line 8: unused imports

pyaerocom/exceptions.py
```diff
@@ -22,6 +22,11 @@
 """
 
 
+class AeronetReadError(IOError):
+    # Aeronet reading failed somehow
+    pass
```
avaldebe, line 27: can you write a test that raises this exception?

avaldebe (commented):
I only have small suggestions, mostly a test that raises the new exception.
You can go ahead and merge if you think I'm nitpicking.

tests/io/test_read_aeronet_sunv3_all_points.py
jgriesfeller: I always thought pytest needed that. Apparently it does not. Removed.

avaldebe: `pytest` only need the file to be called `test_*.py` and the test functions to be called `test_something`.

This entry point it is possible to run all the tests on the file with

``` bash
python path/to/test_file.py
```

the same can be achieved without the entry point with

``` bash
pytest path/to/test_file.py
```

However, I think that those entry points were added to run the tests from an IDE which does not support running the tests directly. This way running the file from the IDE will run the tests.

avaldebe (approved):
:+1: ready to merge

PR #540, opened 2021-12-16
clean up entry points

avaldebe:
Many modules have an `if __name__ == "__main__":` clause at the end, and sometimes on the middle.
As far as I can they fall into one of the following cases:
- left over from the module's development, do not longer work or require specific user to run them
- are an aid to run all tests on a test file from an IDE with without `pytest` support
- offer an command line interface for the module in question

In the first 2 cases, the clauses can be removed. On the 3rd case, the command line interface would be better on a separate script, outside of the installable package.


pyaerocom/io/readgridded.py
```diff
@@ -2226,11 +2226,4 @@ def name(self):
...
 
 
-if __name__ == "__main__":
-    import pyaerocom as pya
-
-    reader = ReadGridded("NorESM2-NHIST_f19_tn14_20190710")
-    print(reader)
-
-    data = reader.read_var("od550aer", start=9999)
 is_3d = lambda var_name: True if "3d" in var_name.lower() else False
```
jgriesfeller, line 2229: not sure if that last line is ever called... But not important.

jgriesfeller (approved):
Looks good with a small comment

PR #547, opened 2022-01-10
Clean tests and organize fixtures

avaldebe:
- [x] split fixtures into small modules for easy re-use
- [x] use `tmp_path`/`tmp_path_factory` fixtures instead of `tmpdir`/`tmpdir_factory` fixtures
- [x] use `pathlib.Path` instead of `os.path`
- [ ] mark broken tests with `pytest.mark.xfail` instead if `pytest.mark.skip`
- [ ] run tests in parallel with `pytest-xdist`


dulte (approved):
I think it is easier to read with the fixtures. It is not much I can comment on. If all the tests still pass, I think it is ready to merge

PR #549, opened 2022-01-14
added variable name ratpm10pm25

jgriesfeller:
begin work on ratpm10pm25


jgriesfeller:
wrong way of doing things

Issue #519, opened 2021-11-11
New variable PM10/PM25 ratio


MichaelSchulzMETNO:
To harmonize dust and aerosol IFS model evaluation two new variables should be included in IFS CAMS84 evaluation: The ratio of PM10/PM25 and the coarse AOD from SDA. Best consult with Sara Basart on implementation.

MichaelSchulzMETNO:
Model output has to provide a  coarse mode od550gt1, to be compared to AERONET v3 SDA Level 1.5 coarse mode AOD.

avaldebe:
In principle, PM10/PM25 would require a new variable definition (with units) and some kind of alias.
Coarse mode should be easier.

jgriesfeller:
Coarse mode AOD is in pyaerocom already. Example: 
https://aeroval.met.no/evaluation.php?project=c3s&exp_name=IASI

It should also be possible to use level 1.5 of SDA. But I have not done that so far.

For the PM10/PM25 ratio I have to look into the code how derived variables are handled. Any suggestion for a name?
I guess you just want the maps? or also the same for observations and a comparison?

jgriesfeller:
This looks relatively easy at first glance. Examples are in the aeroval config repo on gitlab.met.no. I'll try it out...

jgriesfeller:
I could confirm that Aeronet SDA level 3 version 1.5 is usable in aeroval

The ratio of PM10/PM25 is doable for the model data via aeroval, but not for the obs data.

I will implement a proper variable in pyaerocom for both model and obs data.

jgriesfeller:
Unfortunately pm10 and pm25 are provided by a number of observation networks (EEA, GHOST, Airnow, EBAS). Because we don't have calculated observations from read observations (the devil is in the details here), the ratio needs to be implemented in every obs network we want to use for comparison.
I will start with EEA because that is what has been used by BSC earlier. I am not sure yet where exactly the implementation will start because for EEA, the data is in different files, but there's only one instrument per per species. For EBAS on the other hand there's potentially more than one instrument per station, so one also need to perform instrument matching

jgriesfeller:
EEA update: station IDs are not in the filenames unfortunately, so no station matching to avoid unnecessary file reads possible by just using the filenames. Considering a cache file with station codes per file name created at download time since the number of files is so large (3000 - 10000 per year and variable) at this point...TBD

lewisblake:
[...]

PR #550, opened 2022-01-20
calculate the new variable ratpm10pm25 using the calculation infrastructure of the ReadUngridded class

jgriesfeller:
I am trying to introduce the new variable ratpm10pm25 to solve #519 .
In principle that should be easy using 
https://github.com/metno/pyaerocom/blob/18a14f3d8c4740bec60d1acc1f2b2d651d7e8b73/pyaerocom/io/read_eea_aqerep_base.py#L171-L178

but somehow that is not used.

Skript to test:
```
# quick and dirty test for obs reading via the api

DATA_ID = 'EEAAQeRep.v2'
VAR_NAME = ['ratpm10pm25']
# works with the following
# VAR_NAME = ['concpm10']

def main():
    import pyaerocom.io as pio

    obs_obj = pio.ReadUngridded(DATA_ID)
    obs_data = obs_obj.read(vars_to_retrieve=VAR_NAME)
    print(obs_data1)


if __name__ == "__main__":
    main()
```

I could boil down the problem to this line:
https://github.com/metno/pyaerocom/blob/18a14f3d8c4740bec60d1acc1f2b2d651d7e8b73/pyaerocom/io/readungridded.py#L656
where `self.read_dataset` is called instead of `self.read_dataset_post`.
Unfortunately I can not find any code where `self.post_compute` is set to contain the fields used in https://github.com/metno/pyaerocom/blob/18a14f3d8c4740bec60d1acc1f2b2d651d7e8b73/pyaerocom/io/readungridded.py#L539
I deliberately do not get further to no bias your comments

I might not see the wood for the trees, but if you find where `self.read_dataset_post` is set, please tell me!
You have to use branch `griesie_fix519_v2` to test


jgriesfeller:
I reverse engineered the data structure. It's looking very weird:
```
DATA_ID = 'EEAAQeRep.v2'
VAR_NAME = ['ratpm10pm25']
TS_TYPE = 'daily'


def main():
    import pyaerocom.io as pio

    obs_obj = pio.ReadUngridded(DATA_ID)
    obs_obj.post_compute[DATA_ID] = {}
    obs_obj.post_compute[DATA_ID]["data_id"] = DATA_ID

    obs_obj.post_compute[DATA_ID]["aux_requires"] = {}
    obs_obj.post_compute[DATA_ID]["aux_merge_how"] = {}
    obs_obj.post_compute[DATA_ID]["aux_units"] = {}
    for var in VAR_NAME:
        obs_obj.post_compute[DATA_ID]["aux_requires"][var] = {}
        obs_obj.post_compute[DATA_ID]["aux_requires"][var][DATA_ID] = {}
        obs_obj.post_compute[DATA_ID]["aux_requires"][var][DATA_ID] = getattr(obs_obj._readers[DATA_ID],"AUX_REQUIRES")
        # Supported are 'combine', 'mean' and 'eval'
        obs_obj.post_compute[DATA_ID]["aux_merge_how"][var] = 'combine'
        obs_obj.post_compute[DATA_ID]["aux_units"][var] = '1'

    obs_obj.post_compute[DATA_ID]["aux_funs"] = {}
    obs_obj.post_compute[DATA_ID]["aux_funs"][DATA_ID] = {}
    obs_obj.post_compute[DATA_ID]["aux_funs"][DATA_ID] = getattr(obs_obj._readers[DATA_ID],"AUX_FUNS")
    obs_obj.post_compute[DATA_ID]["aux_flag"] = True
```
The reading starts with that at least after some small additional changes in the code
I have more and more the impression the calculated variables are not part of the API but only part of  aerocom-evaluation and/or aeroval. I think this functionality should also be part of the API, so I will add it (I am almost there)

jgriesfeller:
It turned out that the variable calculation infrastructure is not written to work with a calculation method (or just has not been tested). There's also too much testing regarding the units in the code given that calculation methods usually change the unit of the result. But I might not see the use case for the current implementation.

jgriesfeller:
Testing shows that RAM consumption in really bad: I cannot test my code with just 2 years of data the memory consumption is beyond 20GB. The process is just killed by the OS as soon as RAM runs out.

jgriesfeller:
adding some swap enabled calculating ratpm10pm25

jgriesfeller:
Looking into the results of my ratpm10pm25 reading it turns out that the cached reading for the second variable is wrong and provides pm10 data as pm25 data. So ratpm10pm25 is always 1 in that case.
The reading works when the data is read form disk directly though.

jgriesfeller:
Further analysis shows that the cache writing is wrong...

jgriesfeller:
trying to add ratpm10pm25 to the EBAS reader fails after reading the files:
```  
File "/home/jang/data/Python3/pyaerocom_new/pyaerocom/io/readungridded.py", line 723, in read
    self.read_dataset(
  File "/home/jang/data/Python3/pyaerocom_new/pyaerocom/io/readungridded.py", line 412, in read_dataset
    data_read = reader.read(vars_to_read, **kwargs)
  File "/home/jang/data/Python3/pyaerocom_new/pyaerocom/io/read_ebas.py", line 1693, in read
    data.clear_meta_no_data()
  File "/home/jang/data/Python3/pyaerocom_new/pyaerocom/ungriddeddata.py", line 2007, in clear_meta_no_data
    raise DataCoverageError("UngriddedData object appears to be empty")
pyaerocom.exceptions.DataCoverageError: UngriddedData object appears to be empty
```

because the depending variables are not read independently. The same system worked for EEA :( 

More work to do...

jgriesfeller:
postponed calculation ratpm10pm25 fro EBAS for now. This PR is now supposed to work with obs networks **EEAAQeRep.v2** and **EEAAQeRep.NRT** only

jgriesfeller:
Calculating the ratio between pm10 and pm25 actually works using aeroval if the input and the output units are the same. With some unit hacking, one can also make that work for this particular case (the unit changes from ug m-3 to 1)

jgriesfeller:
this PR is not the way to solve the problem. We should explore to run this the same way as aeroval does it.

PR #575, opened 2022-01-27
Fix logging configuration

### LINKED ISSUE #574: logging miss configuration
Logging.info no longer show when running aeroval. This is due to a new change to make all the loggers to one main logger.

This has removed a lot of the printing to console. 

We need to agree on if we should have this printing to console or not, and how to turn it on/off

avaldebe:
fix #574


jgriesfeller (approved):
all good

Issue #573, opened 2022-01-27
Sea Salt (concsspm10) calculation from EMEP results


avaldebe:
as far as I can tell from

https://github.com/metno/pyaerocom/blob/a8891208f8efc779ec6d69957f82869ec7939a8c/pyaerocom/data/emep_variables.ini#L72-L73

https://github.com/metno/pyaerocom/blob/a8891208f8efc779ec6d69957f82869ec7939a8c/pyaerocom/io/read_mscw_ctm.py#L69-L70

https://github.com/metno/pyaerocom/blob/a8891208f8efc779ec6d69957f82869ec7939a8c/pyaerocom/io/read_mscw_ctm.py#L92-L93

https://github.com/metno/pyaerocom/blob/a8891208f8efc779ec6d69957f82869ec7939a8c/pyaerocom/io/_read_mscw_ctm_helpers.py#L173

``` python
concsspm25 = SURF_ug_SeaSalt_f + 0.13 * SURF_ug_SeaSalt_c
concsspm10 = concsspm25 + SURF_ug_SEASALT_C
```

Why does `concsspm25` includes a fraction of `SeaSalt_c`?
And, `concsspm10` looks wrong to me.
I think it should just `SeaSalt_f + SeaSalt_c`, not `SeaSalt_f + 1.13 * SeaSalt_c`

PR #579, opened 2022-02-02
Changes sspm10 to be ssf+ssc

dulte:
Fix for issue #573


dulte:
The 0.13 should be there for sspm2.5, but I also think sspm10 is wrong. So I've made it so sspm10 = ssf + ssc now

avaldebe (approved):
🚀 ready to merge 💥

Issue #580, opened 2022-02-03
region(s) should stored as an array


AugustinMortier:
In AeroVal map json files (e.g: [AeronetL1.5-d-od550aer_Column_IFS-CTRL-od550aer.json](https://aeroval.met.no/data/cams84/eval/map/AeronetL1.5-d-od550aer_Column_IFS-CTRL-od550aer.json) ) each station has a "region" key whose value is a string. 

However, a station could belong to more than one region: e.g Ocean, and North-America.

This problem can be seen with the new [AeroVal minimap feature](https://aeroval.met.no/overall.php?project=cams84&exp_name=eval&tab=taylor). When selecting North-America, no station is highlighted since they all belong to "Oceans" (might be also a bug here).

![minimaps](https://user-images.githubusercontent.com/19610522/152380809-22cc4588-3302-4d03-88d4-1a34f9ae6ee7.png)

jgriesfeller:
Just for the record: due to expected running time of the linked experiment, I chose to re-run [the 2021 only experiment](https://aeroval.met.no/evaluation.php?project=cams84&exp_name=eval-2021) for now. I hope that suffices for testing.
The eval experiment might need more than a day to run because it's quite large. Please let me know if it has to be the eval experiment.

jgriesfeller:
I introduced a testing project on the web server that is only accessible if you know the URL:
https://aeroval.met.no/evaluation.php?project=testing

It has the regions in an array, but the error mentioned above persists.

AugustinMortier:
The AeroVal minimap now can highlight stations belonging to different regions. I think this issue is solved then. Thanks!

![Screenshot from 2022-03-03 16-47-57](https://user-images.githubusercontent.com/19610522/156600058-7969231c-42ac-4f2b-8b07-518819d1eabf.png)

PR #582, opened 2022-02-07
Adds list of regions instead of just the best

dulte:
Addressing #580


pyaerocom/region.py
```diff
@@ -367,7 +367,7 @@ def find_closest_region_coord(lat, lon, regions=None):
     matches = get_regions_coord(lat, lon, regions)
 
     if len(matches) == 1:
-        return matches[0]
+        return matches[0], matches
```
avaldebe, line 370: You're changing the type of the return from `str` to `tuple[str, list[str]]`.
I would create a new data structure for the result. Something based on `typing.NamedTuple`, like this:

``` python
from __future__ import annotations
[...]
from typing import NamedTuple
[...]

class ClosestRegion(NamedTuple):
    best: str
    matches: list[str]

def find_closest_region_coord(
    lat: float,
    lon: float,
    regions: dict[str, Region] | None = None
) -> ClosestRegion:
    [...]
    return ClosestRegion(best, matches)
```

This way you get attribute access to the `find_closest_region_coord` resuts, e.g.

``` python
reg = find_closest_region_coord(lon, lat, regions=regions)
# closest region ID
reg.best
# all regions applicable
reg.matches 
```

pyaerocom/region.py
```diff
@@ -367,7 +367,7 @@ def find_closest_region_coord(lat, lon, regions=None):
...
 
     min_dist = 1e6
```
avaldebe, line 372: this part could be replaced by

``` python
matches.sort(key=lambda id: regions[id].distance_to_center(lat, lon))
return ClosestRegion(matches[0], matches)
```

avaldebe (commented):
I have a couple of suggestions in order to improve the maintainability following up.

pyaerocom/aeroval/coldatatojson_helpers.py
```diff
@@ -628,8 +628,8 @@ def _init_site_coord_arrays(data):
 def _get_stat_regions(lats, lons, regions):
     regs = []
     for (lat, lon) in zip(lats, lons):
-        reg = find_closest_region_coord(lat, lon, regions=regions)
-        regs.append(reg)
+        best, reg = find_closest_region_coord(lat, lon, regions=regions)
+        regs.append([best, reg])
     return regs
```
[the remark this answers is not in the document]
avaldebe: or using the data structure introduced on my next comment

``` python
def _get_stat_regions(
    lats: list[float], 
    lons: list[float],
    regions: dict[str, Region] | None,
) -> list[ClosestRegion]:
    assert len(lats) == len(lons), "miss match number of lat/lon"
    return [
        find_closest_region_coord(lat, lon, regions=regions)
        for lat, lon in zip(lats, lons)
    ]
```

pyaerocom/region.py
```diff
@@ -343,8 +345,12 @@ def get_regions_coord(lat, lon, regions=None):
     return matches
 
 
-def find_closest_region_coord(lat, lon, regions=None):
-    """Find region that has it's center closest to input coordinate
+def find_closest_region_coord(
+    lat: float,
+    lon: float,
+    regions: Optional[dict] = None,
+) -> List[str]:
```
avaldebe, line 352: here you could use the `|` style of annotations, introduced on Python 3.10:

``` python
from __future__ import annotations # the very first import
...
def find_closest_region_coord(
    lat: float,
    lon: float,
    regions: dict | None = None,
) -> list[str]:
    ...
```

pyaerocom/region.py
```diff
@@ -358,23 +364,26 @@ def find_closest_region_coord(lat, lon, regions=None):
...
+    #     return matches[0], matches
+
+    # min_dist = 1e6
+    # best = None
+    # for match in matches:
+    #     region = regions[match]
+    #     dist = region.distance_to_center(lat, lon)
+    #     if dist < min_dist:
+    #         min_dist = dist
+    #         best = match
```
avaldebe, line 387: please remove dead code

avaldebe (changes requested):
small changes, mostly cleaning up commented code

AugustinMortier:
I believe this is connected to this issue, but it appears that some regions are never returned in the list of regions, such as 'N America': e.g: [AeronetL1.5-d-od550aer_Column_IFS-CTRL-od550aer.json](https://gitlab.met.no/aeroval/cams2-82/-/raw/master/IFS/map/AeronetL1.5-d-od550aer_Column_IFS-CTRL-od550aer.json)

As a result, the minimap is not highlighting any station in the selected region in the mini map.
Screenshot from 2022-06-07 11-14-56

Other regions also affected by this issue are:
- Rus,Bel,Ukr
- Pacific,Aust,NZ,

My suspicion is that those regions are probably crossing the -180/180 degrees longitude and that is maybe not well taken into account when fetching the closest region.

AugustinMortier:
In addition, it seems that the 'Oceans' region is added to almost all of the stations. This is not affecteing the statistics calculation though, as the number of stations shown in each region seems to be correct. So we should make sure that the regions in the region list is consistent with the regions used in the statistics calculation.

e.g: AERONET: 
- WORLD: 300 stations
- Oceans: 30 stations

Screenshot from 2022-06-07 11-27-49

Screenshot from 2022-06-07 11-28-15

PR #586, opened 2022-02-21
Exclude cartopy 0.20 from pip installs

avaldebe:
see #583 for details


jgriesfeller (approved):
looks good to me

avaldebe:
I do not know why the Python 3.10 venv crashed, but I think we can figure it out later

PR #589, opened 2022-02-23
make anaconda CI work with iris 3.2 as well

jgriesfeller:
for now just to see if that fixes CI

### DIFF
--- a/tests/test_griddeddata.py
+++ b/tests/test_griddeddata.py
@@ -98,8 +98,11 @@ def test_GriddedData_longitude(data_tm5):
     assert str(data_tm5.longitude.units) == "degrees"
 
     lons = data_tm5.longitude.points
-    assert_allclose(lons.min(), -181.5, rtol=TEST_RTOL)
-    assert_allclose(lons.max(), 175.5, rtol=TEST_RTOL)
+    # assert_allclose(lons.min(), -181.5, rtol=TEST_RTOL)
+    # assert_allclose(lons.max(), 175.5, rtol=TEST_RTOL)
+    # iris >= 3.2
+    assert_allclose(lons.min(), -178.5, rtol=TEST_RTOL)
+    assert_allclose(lons.max(), 178.5, rtol=TEST_RTOL)
 
 
 @data_unavail
--- a/tests/test_griddeddata.py
+++ b/tests/test_griddeddata.py
@@ -1,10 +1,10 @@
 import os
 from datetime import datetime
 
+import iris
 import numpy as np
 import pytest
 import xarray as xr
-import iris
 from iris.cube import Cube
 from numpy.testing import assert_allclose
 
@@ -105,15 +105,13 @@ def test_GriddedData_longitude(data_tm5):
     # import iris
     # version('iris')
     # using iris.__version__ instead
-    version = iris.__version__.split('.')
+    version = iris.__version__.split(".")
     if int(version[0]) >= 3 and int(version[1]) > 1:
         assert_allclose(lons.min(), -178.5, rtol=TEST_RTOL)
         assert_allclose(lons.max(), 178.5, rtol=TEST_RTOL)
     else:
         assert_allclose(lons.min(), -181.5, rtol=TEST_RTOL)
         assert_allclose(lons.max(), 175.5, rtol=TEST_RTOL)
-
-
 
 
 @data_unavail
--- a/tests/test_griddeddata.py
+++ b/tests/test_griddeddata.py
@@ -4,6 +4,7 @@ from datetime import datetime
 import numpy as np
 import pytest
 import xarray as xr
+import iris
 from iris.cube import Cube
 from numpy.testing import assert_allclose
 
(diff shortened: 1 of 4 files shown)


tests/test_griddeddata.py
```diff
@@ -98,8 +99,19 @@ def test_GriddedData_longitude(data_tm5):
...
+    # import iris
+    # version('iris')
+    # using iris.__version__ instead
+    version = iris.__version__.split(".")
+    if int(version[0]) >= 3 and int(version[1]) > 1:
+        assert_allclose(lons.min(), -178.5, rtol=TEST_RTOL)
+        assert_allclose(lons.max(), 178.5, rtol=TEST_RTOL)
+    else:
+        assert_allclose(lons.min(), -181.5, rtol=TEST_RTOL)
+        assert_allclose(lons.max(), 175.5, rtol=TEST_RTOL)
```
avaldebe, line 114: I would allocate the expected values on the if clause and call `assert_allclose`
after. Like this:

``` python
# the results depend on the version of iris
# https://github.com/metno/pyaerocom/issues/588
iris_version = tuple(map(int, iris.__version__.split(".")))
if iris_version > (3, 1):
    lon_min, lon_max = -178.5, 178.5 # correct values
else:
    lon_min, lon_max = -181.5, 175.5
assert_allclose(lons.min(), lon_min, rtol=TEST_RTOL)
assert_allclose(lons.max(), lon_max, rtol=TEST_RTOL)    
```

avaldebe: also could you check that the `assert_allclose` are really needed?
Maybe it would be enough with
``` Python
assert lons.min(), lons.max() == lon_min, lon_max
```

tests/test_griddeddata.py
```diff
@@ -98,8 +99,19 @@ def test_GriddedData_longitude(data_tm5):
...
 
     lons = data_tm5.longitude.points
-    assert_allclose(lons.min(), -181.5, rtol=TEST_RTOL)
-    assert_allclose(lons.max(), 175.5, rtol=TEST_RTOL)
+    # iris >= 3.2 corrected an error in iris.cube.Cube.intersection
+    # this does not work with iris:
+    # from importlib.metadata import version
+    # import iris
+    # version('iris')
+    # using iris.__version__ instead
```
avaldebe, line 107: the pip package is called `"scitools-iris"`
``` Python
>>> from importlib import metadata
>>> metadata.version("scitools-iris")
'3.1.0'
```

avaldebe:
there is at least one more change related to iris-3.2

https://github.com/metno/pyaerocom/runs/5357376124?check_suite_focus=true#step:6:178

the co-located model mean changed in one of tests

```
means = [np.nanmean(coldata.data.data[0]), np.nanmean(coldata.data.data[1])]
    
>       assert_allclose(means, [obsmean, modmean], rtol=TEST_RTOL)
E       AssertionError: 
E       Not equal to tolerance rtol=0.0001, atol=0
E       
E       Mismatched elements: 1 / 2 (50%)
E       Max absolute difference: 0.00980256
E       Max relative difference: 0.0545496
E        x: array([0.31593 , 0.169897])
E        y: array([0.31593, 0.[179])

**Comment by jgriesfeller** [comment]
@avaldebe that is the same error as mentioned in https://github.com/metno/pyaerocom/issues/588#issuecomment-1048771481 and to be expected since the longitudes changed due to the removed bug in iris. 
I agree with your changes.

**Comment by avaldebe** [comment]
I merged #590, so you should virtual env jobs should pass after you merge `main-dev` into your branch.

**Comment by avaldebe** [comment]
maybe we should mark the tests affected by the iris version with `@pytest.mark.xfail(iris_version < Version("3.2"))` and only provide the correct values.

something like this
``` python
# tests/conftest.py
from importlib import metadata
from packaging.version import Version
[...]

# iris >= 3.2 corrected an error in iris.cube.Cube.intersection
# see https://github.com/metno/pyaerocom/issues/588
iris_version = metadata.version("scitools-iris")
need_iris_32 = pytest.mark.xfail(Version(iris_version) < Version("3.2"), reason=f"results are different with iris {iris_version} < 3.2", strict=True)
```

``` python
# tests/test_griddeddata.py
from tests.conftest import need_iris_32
[...]

@need_iris_32
def test_GriddedData_longitude(data_tm5):
    """Test if longitudes are defined right"""
    assert str(data_tm5.longitude.units) == "degrees"

    lons = data_tm5.longitude.points
    assert (lons.min(), lons.max()) == (-178.5, 178.5)
```

avaldebe:
@jgriesfeller 

I tried the `pytest.mar.xfail` approach on 3f1ba0a (PR #547).
Maybe we should do the same here?

tests/test_griddeddata.py
```diff
@@ -98,8 +100,14 @@ def test_GriddedData_longitude(data_tm5):
...
-    assert_allclose(lons.min(), -181.5, rtol=TEST_RTOL)
-    assert_allclose(lons.max(), 175.5, rtol=TEST_RTOL)
+    # iris >= 3.2 corrected an error in iris.cube.Cube.intersection
+    # see https://github.com/metno/pyaerocom/issues/588
+    iris_version = Version(metadata.version("scitools-iris"))
+    if iris_version >= Version("3.2"):
+        lon_min, lon_max = -178.5, 178.5  # correct values
+    else:
+        lon_min, lon_max = -181.5, 175.5
+    assert (lons.min(), lons.max()) == (lon_min, lon_max)
```
avaldebe, line 110: why not to decorate the test with need_iris_32 and test only against the correct values?

``` python
@data_unavail
@need_iris_32
def test_GriddedData_longitude(data_tm5):
    """Test if longitudes are defined right"""
    assert str(data_tm5.longitude.units) == "degrees"

    lons = data_tm5.longitude.points
    assert (lons.min(), lons.max()) == (-178.5, 178.5)
```

avaldebe: done

avaldebe:
ready to merge 🚀 🎆

Issue #436, opened 2021-09-15
Only obs processing in AeroVal


jgliss:
Since the new trends interface is now also handled via the AeroVal interface (and pyaerocom backend processing), we need some functionality to handle a configuration setup (`EvalSetup`) such that it can process an experiment that contains only observations. In other words:

Suppose you want to run an experiment specified in an instance of `CFG=pya.aeroval.setupclasses.EvalSetup` via 
```python 
import pyaerocom as pya
processor = pya.aeroval.ExperimentProcessor(CFG) #with processor.cfg.model_cfg being empty.
processor.run()
```

Since AeroVal processing all goes via co-location and `ColocatedData` objects, we need a fake model to be used, thus, I suggest the following:

**On the fly generation of dummy model data in AeroCom format in a temporary location and add these dummy NetCDF files to `EvalSetup.model_cfg` (will be the only entry).** 

Steps:

1. In `processor.run()` check if `processor.cfg.model_cfg` is empty. If so, then:
2. Create a suitable temporary output location for the data, e.g. `tmpdir=pya.const.LOCAL_TMP_DIR` (which should be ~/MyPyaerocom/tmp
3. Register location of `tmpdir` in pyaerocom as search directory (so the dummy output can be found later on): `pya.const.add_data_search_dir(tmpdir)`.
4. Create dummy model output folder (and subfolder "renamed"), e.g.: `outdir="~/MyPyaerocom/tmp/dummy_model/renamed"` (**dummy_model** will be the model ID to be used).
5. For each obs variable (derived `EvalSetup.obs_cfg`) create a dummy iris Cube (called `cube` below) with low lat/lon resolution covering the whole globe (can be very low res, even only 2x2 grid points) and the timestamps according to the desired period and frequency specified in `EvalSetup`: should cover the total time span required based on the defined periods in the highest specified output resolution (e.g. 2000-2020, daily). Make sure the dummy cube has the correct variable name (`var_name`) and units (see pyaerocom/data/variables.ini for default units for each variable), `ts_type`, and var_name are assigned correctly).
6. Convert that `cube` to `GriddedData`: `dummy_vardata=pya.GriddedData(cube)`
7. Since AeroCom convention demands single year files: Split `dummy_vardata` into `GriddedData` into single year objects (`dummy_vardata_yr`) and for each of these single year GriddedData objects:
8. Save the dummy data in the output directory: `dummy_vardata_yr.to_netcdf(outdir)` (filename will be based on AeroCom convention).
[...]

PR #593, opened 2022-03-04
Obs only

dulte:
Way of handling dummy data to be able obs only evaluation. See #436


dulte:
The _make_dummy_model function is only partially annotated, since importing EvalSetup lead to a circular import

pyaerocom/aeroval/experiment_output.py
```diff
@@ -566,7 +566,14 @@ def _create_statistics_json(self):
         stats_info = statistics_defaults
         stats_info.update(extended_statistics)
         if self.cfg.statistics_opts.add_trends:
-            stats_info.update(statistics_trend)
+            if self.cfg.processing_opts.obs_only:
+                obs_statistics_trend = {}
+                for key in statistics_trend.keys():
```
avaldebe, line 571: You do not need the `.keys()` to iterate over the keys of a dictionary.
It is sufficient to write `for key in statistics_trend: ...`

pyaerocom/aeroval/experiment_output.py
```diff
@@ -566,7 +566,14 @@ def _create_statistics_json(self):
...
+                    if "mod" not in key:
+                        obs_statistics_trend[key] = statistics_trend[key]
+                stats_info.update(obs_statistics_trend)
```
avaldebe, line 574: you could write it as a dictionary comprehension
``` python
            if self.cfg.processing_opts.obs_only:
                obs_statistics_trend = {
                    key: val
                    key, val in statistics_trend.items()
                    if "mod" not in key
                }
                stats_info.update(obs_statistics_trend)                    
```

pyaerocom/aeroval/experiment_processor.py
```diff
@@ -101,8 +106,20 @@ def run(self, model_name=None, obs_name=None, var_list=None, update_interface=Tr
             var_list = [var_list]
 
         self.cfg._check_time_config()
-        model_list = self.cfg.model_cfg.keylist(model_name)
+
         obs_list = self.cfg.obs_cfg.keylist(obs_name)
+        if self.cfg.model_cfg == {}:
```
avaldebe, line 111: you could use the fact that an empty dictionary "is falsy"
``` python
        if not self.cfg.model_cfg:
            ...
```
this also covers the case when `self.cfg.model_cfg is None`

Is it possible for the `model_cfg` attribute to be missing?

pyaerocom/aeroval/helpers.py
```diff
@@ -130,3 +147,56 @@ def _get_min_max_year_periods(statistics_periods):
         if perstop > stopyr:
             stopyr = perstop
     return startyr, stopyr
+
+
+def _make_dummy_model(obs_list: list, cfg) -> str:
```
avaldebe, line 152: Are this functions private? If so, why they are been used on a different module?
Maybe the should regular functions on a private module.

``` python
# _dummy_model.py

def add_dummy_model_to_configuration(obs_list: list, cfg: TypeHintPlease) -> str:
    ...

def remove_dummy_model_files(model_id: str) -> None:
    ...

File: pyaerocom/griddeddata.py [hunk 5]
```diff
@@ -1940,7 +1940,10 @@ def crop(self, lon_range=None, lat_range=None, time_range=None, region=None):
         if time_range is None:
             return GriddedData(data, **suppl)
         else:
-            if all(isinstance(x, str) for x in time_range):
+
+            if all(isinstance(x, str) for x in time_range) or all(
+                isinstance(x, np.datetime64) for x in time_range
+            ):
```
avaldebe, line 1946: `isinstance` can take a tuple of classes to check against
``` python
            if all(isinstance(x, (str, np.datetime64)) for x in time_range):
                ...
```
or you really need all of the list elements to be of the same class?

pyaerocom/griddeddata.py
```diff
@@ -1940,7 +1940,10 @@ def crop(self, lon_range=None, lat_range=None, time_range=None, region=None):
...
                 time_range = (pd.Timestamp(time_range[0]), pd.Timestamp(time_range[1]))
```
avaldebe, line 1947: if you need exactly 2 elements on the list, maybe you should add an assert

pyaerocom/helpers.py
```diff
@@ -1682,3 +1684,75 @@ def get_time_rng_constraint(start, stop):
     t_upper = iris.time.PartialDateTime(year=stop.year, month=stop.month, day=stop.day)
 
     return iris.Constraint(time=lambda cell: t_lower <= cell <= t_upper)
+
+
+def get_max_period_range(periods):
+    start = min([int(per.split("-")[0]) for per in periods])
+    stop = max(
+        [int(per.split("-")[1]) if len(per.split("-")) > 1 else int(per) for per in periods]
+    )
```
avaldebe, line 1693: you do not need to pass a list to `min`/`max` a generator expression will work just as well
``` python
    start = min(int(per.split("-")[0]) for per in periods)
    stop = max(
        int(per.split("-")[1]) if "-" in per 1 else int(per) for per in periods
    )
```

pyaerocom/helpers.py
```diff
@@ -1682,3 +1684,75 @@ def get_time_rng_constraint(start, stop):
...
+
+    return start, stop
+
+
+def _make_dummy_cube(
+    var_name: str, start_yr: int = 2000, stop_yr: int = 2020, freq: str = "daily", dtype=float
+) -> iris.cube.Cube:
```
avaldebe, line 1700: same as before, if this is a private function?
if so, why it is not called on this module?

avaldebe (changes requested):
I have some minor comments about the code itself.
More importantly, I would appreciate the private functions to be collected on a private module with a descriptive name.

pyaerocom/aeroval/experiment_output.py
dulte: I initially did this, but pylance told me that I needed an else statement there as well

pyaerocom/aeroval/experiment_processor.py
dulte: I don't think it is possible. I think aeroval will crash if no model_cfg is given

pyaerocom/aeroval/helpers.py
dulte: Good point, this is just a remnant of an old implementation, where _make_dummy_model was in the experiment_processor. But since it has been moved to the helpers, I'll remove the _

dulte:
I think I've made all the changes requested

pyaerocom/aeroval/experiment_output.py
avaldebe: strange...

pyaerocom/aeroval/experiment_processor.py
avaldebe: and `model_cfg = None`?

pyaerocom/aeroval/helpers.py
```diff
@@ -130,3 +147,56 @@ def _get_min_max_year_periods(statistics_periods):
...
+
+def delete_dummy_model(model_id: str):
+    tmpdir = const.LOCAL_TMP_DIR
+    const.add_data_search_dir(tmpdir)
+
+    outdir = os.path.join(tmpdir, f"{model_id}/renamed")
+    dirs = glob.glob(outdir + "/*.nc")
+    for d in dirs:
+        print(f"Deleting dummy model {d}")
+        os.remove(d)
```
avaldebe, line 202: maybe you could use `pathlib.Path` here

``` python
from pathlib import Path
[...]

def delete_dummy_model(model_id: str) -> None:
    tmpdir = const.LOCAL_TMP_DIR
    const.add_data_search_dir(tmpdir)

    renamed = Path(tmpdir) / f"{model_id}/renamed"
    for path in renamed.glob("*.nc"):
        print(f"Deleting dummy model {path}")
        path.unlink()
```

pyaerocom/aeroval/helpers.py
```diff
@@ -130,3 +147,56 @@ def _get_min_max_year_periods(statistics_periods):
...
+def make_dummy_model(obs_list: list, cfg) -> str:
+
+    # Sets up variable for the model register
+    tmpdir = const.LOCAL_TMP_DIR
+    const.add_data_search_dir(tmpdir)
+
+    model_id = "dummy_model"
+    outdir = os.path.join(tmpdir, f"{model_id}/renamed")
+
+    os.makedirs(outdir, exist_ok=True)
```
avaldebe, line 161: not sure if `GriddedData.to_netcdf` accepts `pathlib.Path` objects but it might be cleaner to write this part with a Path object

``` python
    outdir = Path(tmpdir) / f"{model_id}/renamed"
    outdir.mkdir(parents=True, exist_ok=True)
```

avaldebe (changes requested):
Minor requests regarding the use of `pathlib.Path` instead of `os.path.*`.

Please, feel free to ignore them if you do not see the added value

pyaerocom/aeroval/helpers.py
dulte: `GriddedData.to_netcdf` uses `os.path.join`, so I think it must be str. If we are going to change this, then it should be its own PR

avaldebe: `os.path.join` accepts [path-like objects](https://docs.python.org/3/glossary.html#term-path-like-object) (see [docs](https://docs.python.org/3/library/os.path.html#os.path.join)),
which means that it will accept `pathlib.Path` objects

avaldebe: @dulte ?

pyaerocom/helpers.py
```diff
@@ -1682,3 +1684,73 @@ def get_time_rng_constraint(start, stop):
     t_upper = iris.time.PartialDateTime(year=stop.year, month=stop.month, day=stop.day)
 
     return iris.Constraint(time=lambda cell: t_lower <= cell <= t_upper)
+
+
+def get_max_period_range(periods):
+    start = min([int(per.split("-")[0]) for per in periods])
```
avaldebe, line 1690: you can pass the generator expression directly to `min` without creating a list, e.g.

``` python
    start = min(int(per.split("-")[0]) for per in periods)
```

pyaerocom/helpers.py
```diff
@@ -1682,3 +1684,73 @@ def get_time_rng_constraint(start, stop):
...
+    stop = max(int(per.split("-")[1]) if len(per.split("-")) > 1 else int(per) for per in periods)
```
avaldebe, line 1691: you can simplify the generator expression by processing the last element on the split,
if you are concerned about invalid periods you could add an assert

``` python
    assert all(per.count("-") <= 1 for per in periods), f"invalid period in {periods=}"
    stop = max(int(per.split("-")[-1]) for per in periods)
```

pyaerocom/helpers.py
```diff
@@ -1682,3 +1684,73 @@ def get_time_rng_constraint(start, stop):
...
+    return start, stop
+
+
+def make_dummy_cube(
+    var_name: str, start_yr: int = 2000, stop_yr: int = 2020, freq: str = "daily", dtype=float
+) -> iris.cube.Cube:
+    startstr = f"days since {start_yr}-01-01 00:00"
+
+    if freq not in TS_TYPE_TO_PANDAS_FREQ.keys():
+        raise ValueError(f"{freq} not a recognized frequency")
```
avaldebe, line 1702: you can test membership on a dictionary directly, without calling `.keys()`
``` python
    if freq not in TS_TYPE_TO_PANDAS_FREQ:
￼        raise ValueError(f"{freq} not a recognized frequency")
```

avaldebe (changes requested):
small nitpicks

Issue #551, opened 2022-01-24
Observation catalogue / overview


MichaelSchulzMETNO:
For users, production, but also for development: It would be useful to have a way to get an overview on observational data sets which can be read by pyaerocom. 

Which obs data sets are accessible via aerocom from lustre/metno?
Which obs data sets are accessible publicly?

What are the characteristics of the obs dataset that can be accessed via pyaerocom? (variables, temporal coverage&resolution, pyaerocom name,id,location of dataset, origin, data policy, additional documentation (papers, websites), version management on lustre, which pyaerocom reader is attached )

How can these characteristics be retrieved? How can this be stored best in pyaercom in the future? 

Can help on getting such an overview be included in the documentation?

avaldebe:
see #572

avaldebe:
We could start with enhancing the documentation of the reader, so the description is more than a list of acronyms.

MichaelSchulzMETNO:
And have a list of readers on the documentation page, (if not already somewhere ?)

PR #598, opened 2022-03-21
Docs551

lewisblake:
Addressing some of the desires in item #551 to enhance the documentation cataloging which observational data sets can be read by pyaercocom. Some of this information is in the "About" section of the documentation, however it can be easy to miss, so I made some modifications to docs > api.rst with relevant links. Want to discuss appropriate placement for this information and desired level of granularity (more, less).


krikarls (approved):
Looks good to me!

Issue #496, opened 2021-10-18
paerocom cli should have a --version argument


jgriesfeller:
In the Unix world it's a good practice for every program to provide an easy way of determining it's version number.
Usually that is done with a --version command line option.

The purpose for me is to determine easily on which version number I am.

avaldebe:
Should we add some additional info listing the installed `pya` dependencies?
This is quite easy to implement with [importlib.metadata.version](https://docs.python.org/3/library/importlib.metadata.html#distribution-versions)
Here is an example, from one of my CLI apps

```
$ utils4cwf -V

utils4cwf (0.4.0)
dependencies (installed version)

data formats
netCDF4 (1.5.5.1), eccodes (1.3.3), cfgrib (0.9.9.1), pyarrow (6.0.0)

data manipulation
dask (2021.11.1), xarray (0.20.1), scipy (1.6.1), numpy (1.19.5)

remote execution
fabric (2.6.0), invoke (1.6.0), paramiko (2.8.0)

configuration
python-dotenv (0.19.2), toml (0.10.2)

command line
typer (0.4.0), click (8.0.3)
```

jgriesfeller:
Sounds like a good idea
What I was after can be achieved like this
```
>>> import pyaerocom
>>> from importlib.metadata import version
>>> version('pyaerocom')
'0.12.2.dev1'
```
This is so small that I can implement it right away in https://github.com/metno/pyaerocom/blob/main-dev/pyaerocom/scripts/cli.py

avaldebe:
on the current version you can get `__version__` as follows

``` python
>>> from pyaerocom import __version__
>>> print(__version__)
0.12.2.dev1
```

which I implemented using `importlib.metadata.version`

https://github.com/metno/pyaerocom/blob/3e8048895bbd403e504a99e38967f43afc441dbf/pyaerocom/__init__.py#L6

> This is so small that I can implement it right away in https://github.com/metno/pyaerocom/blob/main-dev/pyaerocom/scripts/cli.py

Yes, you could just do that. However, I would take the opportunity to re-implement `pyaerocom.scripts.cli` using [typer](https://typer.tiangolo.com/tutorial).

jgriesfeller:
The do it, I am too used to argparse.

avaldebe:
addressed by #624

PR #602, opened 2022-03-21
Versionflag496

lewisblake:
Not to get too hasty, but I wanted to get a continue the conversation around this feature (#496), and comments, code reviews, etc., seem like a good way to go about it.

@avaldebe You mentioned using typer in your linked comment. I wanted to see if you still think we should go that route to re-implement `pyaerocom.scripts.cli` or whether something simple works for now. Currently it seems like pyaerocom doesn't depend on typer and so going that direction seems like a bigger modification. Also not sure if tests are run on CLIs / if possible


avaldebe:
> @avaldebe You mentioned using typer in your linked comment. I wanted to see if you still think we should go that route to re-implement `pyaerocom.scripts.cli` or whether something simple works for now. Currently it seems like pyaerocom doesn't depend on typer and so going that direction seems like a bigger modification. Also not sure if tests are run on CLIs / if possible

This CLI is very simple, so you can implement the extra option using `argparse.ArgumentParser`.

lewisblake:
Putting testing CLI on the back-burner for now. Can revisit if we think it's worth it and I have more time

lewisblake:
A simple implementation of the --version flag for the CLI. Made my Typer-base implementation a different branch and we can look at the separately.

lewisblake:
Typer-based CLI alternative #624

Issue #601, opened 2022-03-21
CAMS2_83 FAIRMODE metrics in pyaerocom


michaelgau:
implement FAIRMODE metrics in pyaerocom.
Daniel has already implemented MQI.
Next in line is: MQI90. That's the 90th percentile among all stations' MQIs.
90% of all stations have an MQI that is better (i.e. smaller) than MQI90.

After MQI90, we also need the three different parts of MQI. The bias indicator, standard deviation indicator, and correlation indicator.

Should have this in April.

PR #607, opened 2022-03-24
Fairmode Beta*MQI

lewisblake:
Beginning to address #601. When I create the draft pull request, some of the tests will fail (mostly formatting, IIRC), however that was the previous state of the branch and I'm not sure how to fix it now / if I should be the one to do so.

This implements Beta*MQI, where Augustin will be able to divide by Beta closer to the user-interface and test different values.

Some of key changes include moving the computation of FAIRMODE statistics and tests into their own module (i.e., out of `coldatatojson_helpers.py` and `test_coldatatojson_helpers2.py`), and writing a test of two cases. An assert is now in the newly named `compute_fairmode_stats()` function to ensure calculations are performed correctly within tolerance. More test cases will be needed to check this functionality.


avaldebe:
@dulte and @lewisblake 

Please merge `main-dev` into your respective branches so I can see only
the differences corresponding to this PR.

pyaerocom/aeroval/compute_fairmode_stats.py
```diff
@@ -0,0 +1,98 @@
...
+            UrRV=0.28,
+            RV=50,
+            alpha=0.13,
+        ),
+        concpm25=dict(
+            UrRV=0.28,
+            RV=25,
+            alpha=0.3,
+        ),
+    )
```
avaldebe, line 31: If `species_values` is a constant, it  could be defined as a CONSTANT at the top of the module, .e.g.

``` python
SPECIES = dict(
    concno2=dict(UrRV=0.24, RV=200, alpha=0.2),
    vmro3=dict(UrRV=0.18, RV=120, alpha=0.79),
    concpm10=dict(UrRV=0.28, RV=50, alpha=0.13),
    concpm25=dict(UrRV=0.28, RV=25, alpha=0.3),
)
```

pyaerocom/aeroval/compute_fairmode_stats.py
```diff
@@ -0,0 +1,98 @@
+"""
+Functions for computing the FAIRMODE statistics
+"""
+
+from math import sqrt, isclose
+from typing import Tuple
+
+
+def _get_RMSU(mean: float, std: float, species: str) -> Tuple[float, float, float, float]:
```
avaldebe, line 9: Why `_get_RMSU` instead of `_RMSU`?
I do not see any web request or any kind of `get` action.

Also, a docstring about what the acronym stands for would be useful for someone unfamiliar with the FAIRMODE indicators, like me

pyaerocom/aeroval/compute_fairmode_stats.py
```diff
@@ -0,0 +1,98 @@
...
+        ),
+        concpm25=dict(
+            UrRV=0.28,
+            RV=25,
+            alpha=0.3,
+        ),
+    )
+
+    if species not in species_values.keys():
+        raise KeyError(f"Species {species} not in list {species_values.keys()}")
```
avaldebe, line 34: You do not need `.keys()` to check if `species` is a key on the dict.
Also, from the point of view of the user of this function this is a `ValueError`.
I mean, the function was called with an unsupported value...

``` python
    if species not in species_values:
        raise ValueError(f"Unsupported {species=})
```

pyaerocom/aeroval/compute_fairmode_stats.py
```diff
@@ -0,0 +1,98 @@
...
+    alpha = species_values[species]["alpha"]
+
+    in_sqrt = (1 - alpha**2) * (mean**2 + std**2) + alpha**2 * RV**2
+
+    return UrRV * sqrt(in_sqrt), UrRV, RV, alpha
+
+
+def _get_fairmode_sign(mod_std: float, obs_std: float, R: float) -> float:
+    if obs_std * sqrt(2 * (1 - R)) == 0:
+        return 1
```
avaldebe, line 47: why `_get_fairmode_sign` instead of `_fairmode_sign`? ...
And docstring ....

Also, this is a good place for some asserts
``` python
def _get_fairmode_sign(mod_std: float, obs_std: float, R: float) -> int:
   assert obs_std >= 0, f"negative {obs_std=}"
   assert 0 <= R <= 1, f"out of range {R=}"
   if obs_std <= 0 or R <= 0 or R >= 1:
        return 1
```

pyaerocom/aeroval/compute_fairmode_stats.py
```diff
@@ -0,0 +1,98 @@
...
+    a = abs(mod_std - obs_std) / (obs_std * sqrt(2 * (1 - R)))
+    return 1 if a >= 1 else -1
+
+
+def _get_crms(mod_std: float, obs_std: float, R: float) -> float:
```
avaldebe, line 52: why `_get_crms` ...? docstring ...

avaldebe (changes requested):
many small nitpicks, otherwise a good first draft

tests/aeroval/test_fairmode_stats.py
```diff
@@ -0,0 +1,36 @@
...
+def test_fairmode_stats(obs_var: str, stats: dict, n_stats: int):
+    # Create a list with all the statistics returned by fairmode_stats(). If adding statistics there, will need to add here.
+    valid_stats = ["RMSU", "sign", "crms", "bias", "rms", "alpha", "UrRV", "RV", "beta_mqi"]
+    fairmode_stats_for_testing = compute_fairmode_stats(obs_var, stats)
+    # Check that the length of the returned dict has length all the fairmode stats keys if obs_var is a legitmate species, or returns an empty dict
+    assert len(fairmode_stats_for_testing) == n_stats
+    # If expected number of statistics is the same as the valid_stats
+    if n_stats == len(valid_stats):
+        # Check that all keys in dict returned by fairmode_stats() are in valid_stats.
+        assert all(key in fairmode_stats_for_testing for key in valid_stats)
```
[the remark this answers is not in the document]
lewisblake: Addressed

pyaerocom/aeroval/compute_fairmode_stats.py
lewisblake: Addressed

pyaerocom/aeroval/compute_fairmode_stats.py
lewisblake: Addressed

avaldebe (approved):
🚀 ready to merge 🎆

PR #614, opened 2022-03-31
Fix minor fairmode bug

lewisblake:
- Fixed small bug in `fairmode_stats.py::fairmode_stats()` so that it now checks for `NaN`s in the `stats` dictionary. 
- Fixed bug in `_fairmode_sign()` so that it is now  -1 <= R <= 1. Also check for NaNs here as well
- We were importing all of numpy and a few specific functions from cmath. Cleaned up to only import four functions from numpy

Note: we might want to eventually test for NaNs in `test_fairmode_stats.py`


dulte (approved):
Looks good

lewisblake:
Added more tests but still decreased our coverage a very small amount (one line) :shrug:

PR #620, opened 2022-04-04
pip install on machines with old/new versions of Proj

avaldebe:
Different versions for cartopy require different versions of Proj, see #583
This PR moves cartopy and iris (which depends on a specific version of cartopy)
to extras and updates the installation docs accordingly:

``` bash
# install pyaerocom on machines with Proj8 or newer
# e.g. Ubuntu 22.04 LTS (Jammy Jellyfish)
pip install pyaerocom[proj8]

# install pyaerocom on machiles with an older version of Proj
# e.g. Ubuntu 20.04 LTS (Focal Fossa)
pip install pyaerocom[proj-legacy]
```

### DIFF
--- a/.readthedocs.yml
+++ b/.readthedocs.yml
@@ -1,17 +1,17 @@
 # .readthedocs.yml
 # Read the Docs configuration file
 # See https://docs.readthedocs.io/en/stable/config-file/v2.html for details
-
+---
 version: 2
 
 sphinx:
-   configuration: docs/conf.py
+  configuration: docs/conf.py
 
 formats:
-   - pdf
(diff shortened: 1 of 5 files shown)


magnusuMET (commented):
It is quite unfortunate we can't solve this in a better way than to make the end-user specify which proj they are using. We should discuss whether `pip install` is used by end-users or just the developers. These flags should also be deprecated when 20.04 has exited mainstream (a years time??)

magnusuMET:
Any way we could use [PEP508](https://peps.python.org/pep-0508/) to specify the requirements based on the OS?

PR #622, opened 2022-04-06
Fix dates cams2 83

dulte:


pyaerocom/scripts/CAMS2_83/processer_engine.py
```diff
@@ -0,0 +1,138 @@
...
+
+        median_stats = {}
+        for key in stats_list.keys():
+            median_stats[key] = np.nanmedian(np.array(stats_list[key]))
+
+        return median_stats
+
+    def _sort_coldata(
+        self, coldata: list[ColocatedData]
+    ) -> Tuple[dict[str, list[ColocatedData]], set[str]]:
```
avaldebe, line 118: does this private method need to be a method at all? or could it be a helper function on the module?

pyaerocom/scripts/CAMS2_83/processer_engine.py
```diff
@@ -0,0 +1,138 @@
...
+        for station in station_list:
+            d = data.sel(station_name=[station])
+            arr = ColocatedData(d)
+            stats = arr.calc_statistics(use_area_weights=use_weights)
+
+            for key in stats_list.keys():
+                stats_list[key].append(stats[key])
+
+        median_stats = {}
+        for key in stats_list.keys():
```
avaldebe, line 111: do not need to iterate over a dick `.keys()`

pyaerocom/scripts/CAMS2_83/processer_engine.py
```diff
@@ -0,0 +1,138 @@
...
+
+    def _get_median_stats_point(self, data, use_weights) -> dict:
+
+        stats_list: dict[str, list[float]] = {
+            "rms": [],
+            "R": [],
+            "nmb": [],
+            "mnmb": [],
+            "fge": [],
+        }
```
avaldebe, line 99: this could be simplified using [`collections.defaultdict`](https://docs.python.org/3/library/collections.html#collections.defaultdict)

``` python
from collections import defaultdict
[...]

        stats_list: dict[str, list[float]] = defaultdict(list)
        [...]


```

avaldebe (commented):
improvements sugestions

PR #626, opened 2022-04-22
change URL to test new pyaerocom webserver with CI

jgriesfeller:
This is basically a pull request to test CI with a new pyaerocom.met.no web server in OSTACK.
For testing purposes only to make sure that the new server works with CI


jgriesfeller:
Thinking about it, I think it's better to put the new name into main-dev for a short period of time to make sure that CI always works

avaldebe (approved):
🚀 🎆

PR #627, opened 2022-04-22
Fix o3max

dulte:
Makes it so that the obs readers read o3max as o3. It is not a perfect fix, but at least *a* fix.

It works with EBAS


WillemVanCaspel (approved):
Very talented programmer!

WillemVanCaspel (approved):
This is even better!

avaldebe:
what is the plan for reading plain `O3` as `O3`?

PR #633, opened 2022-04-29
Remove extra c from concsspm10

dulte:


WillemVanCaspel (approved):
This fixes the reader! And this is why we need tests...

PR #641, opened 2022-05-04
enable variable ratpm10pm25 via aeroval

jgriesfeller:
This PR makes it possible to calculate the variable ratpm10pm25 via aeroval **for observation**.
The minimum snippet to use is this (example for AirNow obs network):
```Python
ODCSFUN = "AirNow;concpm10/AirNow;concpm25"
OBS_GROUNDBASED = {
    "AirNow": dict(
        obs_id="blablubb",
        obs_vars=["ratpm10pm25"],
        obs_type="ungridded",
        obs_vert_type="Surface",
        obs_merge_how={
            "ratpm10pm25": "eval",
        },
        obs_aux_requires={
            "ratpm10pm25": {
                "AirNow": ["concpm10", "concpm25"],
            }
        },
        obs_aux_funs={"ratpm10pm25": ODCSFUN},
        obs_aux_units={"ratpm10pm25": "1"},
    ),
}
```

Tests are still missing


pyaerocom/units_helpers.py
```diff
@@ -48,6 +48,10 @@
         ["concoa", "ug C/m3", "ug m-3", 1.0],
         ["concoc", "ug C/m3", "ug m-3", 1.0],
         ["conctc", "ug C/m3", "ug m-3", 1.0],
+        # a little hacky for ratpm10pm25...
```
jgriesfeller, line 51: Needed because pyaerocom always tries to convert evrerything into the target unit first

pyaerocom/units_helpers.py
```diff
@@ -48,6 +48,10 @@
...
+        # ["ratpm10pm25", "ug m-3", "1", 1.0],
+        ["concpm25", "ug m-3", "1", 1.0],
+        ["concpm10", "ug m-3", "1", 1.0],
```
avaldebe, line 54: could you add some tests to cover this helpers?

pyaerocom/ungriddeddata.py
```diff
@@ -1230,7 +1230,12 @@ def to_station_data_all(
                 out_data["stats"].append(data)
 
             # catch the exceptions that are acceptable
-            except (VarNotAvailableError, TimeMatchError, DataCoverageError) as e:
+            except (
+                VarNotAvailableError,
+                TimeMatchError,
+                DataCoverageError,
+                NotImplementedError,
+            ) as e:
```
avaldebe, line 1238: not sure if this lines are covered somewhere on the test suite, but it would be great to test that `NotImplementedError` can be used to filter obs

pyaerocom/data/variables.ini
```diff
@@ -5283,3 +5283,17 @@ minimum = -150
...
+minimum = -150
+maximum = 150
+dimensions = time,lat,lon
+
+[ratpm25pm10]
+description = ratio of pm25 and pm10
+unit = 1
+minimum = -150
+maximum = 150
+dimensions = time,lat,lon
```
avaldebe, line 5299: please add an `\n` ad the end of the file

avaldebe (changes requested):
I would appreciate some tests to cover the new unit helpers and the `NotImplementedError` obs filter

pyaerocom/data/variables.ini
jgriesfeller: done

pyaerocom/ungriddeddata.py
jgriesfeller: Not sure how to test these exception effectively. Any ideas?

pyaerocom/units_helpers.py
jgriesfeller: done

pyaerocom/ungriddeddata.py
jgriesfeller: The only ideas I have right now is a fake reader that throws these exceptions or a fabricated data set. Both is takes some time to create.

jgriesfeller: I will extend these exceptions for the next release (e.g. with the exception raised when a station moved more than a kilometer).
How about postponing that test until then?

jgriesfeller:
after speaking with @avaldebe : merging

Issue #639, opened 2022-05-04
Missing statistics for individual years [high priority]


AugustinMortier:
Since recently (don't know exactly when, but we noticed that first with a cams2-83 experiment), the statistics in the stations map files are empty for the individual years, even though data are available (see Screenshot).

![image](https://user-images.githubusercontent.com/19610522/166647935-86dc6761-cac7-451d-9bce-3e8dda99a356.png)

avaldebe:
we only run CAMS2_83 for late 2021 and early 2022,
so I do not see how we could have results for 2013..

AugustinMortier:
This was CAMS2-82 results, as an illustration.

jgriesfeller:
@AugustinMortier which file is that? I would like to look at other experiments. It seems that sometimes it's working and sometimes not (I had experiments where it actually worked)
That info will also help the debugging :-)

AugustinMortier:
It's the stations map files in map/ directory.
e.g: https://gitlab.met.no/aeroval/cams2-82/-/tree/master/IFS-AQ/map

lewisblake:
I ran an experiment to try and recreate this issue. I asked @jgriesfeller for the config file and he told me the config file in questions was [this one](https://gitlab.met.no/aeroval/config/-/blob/griesie_dev/config_files/cfg_cams2-82_eval_combined_only_2022.py). I pulled the branch, made a copy, and ran a stripped down version of the script:

- `periods=[
        "2020-2022",
        "2021",
        "2020",
    ], `

- `res = ana.run(obs_name = "EEA-NRT-rural", var_list = ["concpm10"])`

According to what I see [on Aeroval](https://aeroval.met.no/evaluation.php?project=cams2-82&exp_name=IFS-AQ&obs=EEA-NRT-rural#), assuming this bug is consistent and reproducible, I believe I should see json files in `aeroval/data/cams2-82/IFS-AQ-2022/map` with missing daily values for the yearly statisics (2021, 2020). However, from what I see, the json files do not display any signs of this bug. All the statistics look reasonable any the only `null` values I see correspond to instances where zero valid observations. 

It is possible I have not implemented my experiment in a way which matches the circumstances described in the issue and need to be more exhaustive. That said, I am surprised that this bug is so difficult to reproduce and that it has not appeared in many other experiments. Perhaps there was just an issue with the particular run (e.g., write-to-disk issue). I can show a bit of what I did on Monday at the pyaerocom meeting.

[...]

PR #653, opened 2022-05-27
Make aeroval cope with full NaN obs time series (solve #639:)

jgriesfeller:
1st try to solve #639

### DIFF
--- a/pyaerocom/aeroval/coldatatojson_helpers.py
+++ b/pyaerocom/aeroval/coldatatojson_helpers.py
@@ -635,7 +635,6 @@ def _get_stat_regions(lats, lons, regions):
 
 
 def _process_sites(data, regions, regions_how, meta_glob):
-
     freqs = list(data)
     (sites, lats, lons, alts, countries, jsdates) = _init_site_coord_arrays(data)
     if regions_how == "country":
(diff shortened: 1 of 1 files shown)


jgriesfeller:
A test experriment failing without this PR results now in a correcty experiment. Converting draft PR to PR .

pyaerocom/aeroval/coldatatojson_helpers.py
```diff
@@ -826,9 +826,9 @@ def _process_map_and_scat(
     scat_data = {}
     scat_dummy = [np.nan]
     for freq, cd in data.items():
-        use_dummy = True if cd is None else False
         for per in periods:
             for season in seasons:
+                use_dummy = True if cd is None else False
```
avaldebe, line 831:
``` python
        for per in periods:
            for season in seasons:
                use_dummy = cd is None
                ...
```

avaldebe (changes requested):
please write the boolean directly, instead of using a ternary operator

pyaerocom/aeroval/coldatatojson_helpers.py
jgriesfeller: I have to admit that rarely think of these short expressions because I don't find them very intuitive...
Done.

pyaerocom/aeroval/coldatatojson_helpers.py
```diff
@@ -821,14 +820,13 @@ def _process_map_and_scat(
...
 ):
-
     stats_dummy = _init_stats_dummy()
     scat_data = {}
     scat_dummy = [np.nan]
     for freq, cd in data.items():
-        use_dummy = True if cd is None else False
         for per in periods:
             for season in seasons:
+                use_dummy = cd is None
```
avaldebe, line 829: if `use_dummy = cd is None` is confusing maybe it would be more readable to replace
`if not use_dummy` with `if cd is not None`.

jgriesfeller: It's fine. This is me using several languages at once and therefore trying not to get too much into language specifics. After 2 seconds I got it.
I had just moved the existing line without much thinking about it.

avaldebe (approved):

PR #659, opened 2022-05-31
1st try to replace WORLD with ALL in used filter names

jgriesfeller:
and using only one constant to store the value


jgriesfeller:
@AugustinMortier as soon as you have finished your testing and can confirm that it's working for the web interface, I can start the review

AugustinMortier:
I confirm that the web interface now supports *ALL*.

Issue #658, opened 2022-05-31
Aeroval: heatmap data is also wrong if there's all NaN's in a season (related #639)


jgriesfeller:
Basically the same as #639 but for heatmap data:

The loops are basically the same at the heatmap data computation as at the map data computation:
https://github.com/metno/pyaerocom/blob/e8f0c45ec1adc9c43a3b2cd1ba81ce9797cb2dd7/pyaerocom/aeroval/coldatatojson_helpers.py#L1117-L1126

https://github.com/metno/pyaerocom/blob/e8f0c45ec1adc9c43a3b2cd1ba81ce9797cb2dd7/pyaerocom/aeroval/coldatatojson_helpers.py#L1119 needs to be moved into the seasons loop

PR #661, opened 2022-06-01
corrected heatmap data when entire season is full of NaNs

jgriesfeller:
Correct #658 
The heatmap data is basically computed using the same loop as the map data. The fix just moves the resetting of `use_dummy` into the season loop.


avaldebe (approved):
🚀 ready to merge 🎆

Issue #647, opened 2022-05-13
Split up overall ts files


AugustinMortier:
The overall timeseries (located in hm/ts/) should be separated into multiple files, as they can be very large (>100MB) when considering multiple models, observations, .... All models should be in the same file, but we could separate them at least per observation and perhaps per region.

jgriesfeller:
Maybe a common practice from SQL databases helps here: The principle there is make record sizes as small as possible (and as big as necessary).
In our case the principle should be that the files created should **not** contain more information than what is needed for the current visualisation.
Maybe we should go over how the data is organised. Is there anybody besides me with some database experience? Remember that we might want to put the whole data into a database at some point. So that effort is not lost. 
In addition, I need to know the data format to improve the parallelism.

jgriesfeller:
Since I just had a look at the structure of all the json files for my parallelization efforts, here's a little documentation:
The file `hm/ts/stats_ts.json` looks like this:
```
{
  "concpm25": {
    "AN-EEA-MP": {
      "Surface": {
        "IFS-OSUITE": {
          "sconcpm25": {
            "WORLD": {
              "1610668800000": {
                "totnum": 62403,
                "num_valid": 38742,
                "refdata_mean": 25.35994,
                "refdata_std": 29.8055,
                "data_mean": 28.44732,
                "data_std": 38.01789,
                "weighted": 0,
                "rms": 30.18373,
                "R": 0.63196,
                "R_spearman": 0.81003,
                "R_kendall": 0.60445,
                "nmb": 0.12174,
                "mnmb": -0.04012,
                "fge": 0.54547,
                "num_coords_tot": 2013,
                "num_coords_with_data": 1881
              },

```
So it should be easy to split that as done e.g. with the map files by obs network and variable

avaldebe:
nice, here is the `pydatic` model generated by [jsontopydantic](https://jsontopydantic.com/)

``` python
from __future__ import annotations

from pydantic import BaseModel, Field


class Field1610668800000(BaseModel):
    totnum: int
    num_valid: int
    refdata_mean: float
    refdata_std: float
    data_mean: float
    data_std: float
    weighted: int
    rms: float
    r: float = Field(..., alias='R')
    r_spearman: float = Field(..., alias='R_spearman')
    r_kendall: float = Field(..., alias='R_kendall')
    nmb: float
    mnmb: float
    fge: float
    num_coords_tot: int
    num_coords_with_data: int


class World(BaseModel):
[...]

PR #662, opened 2022-06-01
format with black

lewisblake:


pyaerocom/aeroval/coldatatojson_helpers.py
```diff
@@ -4,7 +4,8 @@
 import logging
 import os
 from datetime import datetime
-#from nbformat import current_nbformat_minor
+
+# from nbformat import current_nbformat_minor
```
avaldebe, line 8: clean up dead code

lewisblake: :heavy_check_mark:

lewisblake:
This should address the primary functionality and testing requested in #647. @AugustinMortier also requests a way to be able to determine the version of pyaerocom used in an experiment from outputted configuration json file (e.g., cfg_examples_emep-example.json). The current implementation does not do this yet.

PR #666, opened 2022-06-02
Split ts files 647

lewisblake:
This PR addresses #647. Changes include:

- When running Aeroval, in `hm/ts`, the outputted files are now into individual JSON files for each observation network, model variable, and vertical code combination, following a `obs-var-vert.json` format. This is accomplished through the creation on a `get_timeseries_filename()` function in `coldatatojson_helpers.py`. This function is also tested in `test_coldatajson_helpers2.py`.
- Additionally, a `pyaerocom_version` flag is included in the outputted configuration JSON file. This is implemented in the `ExperimentInfo` class in `setupclasses.py`. This can be used on the web interface to determine how the files in `hm/ts` are formatted.

### DIFF
--- a/pyaerocom/aeroval/coldatatojson_helpers.py
+++ b/pyaerocom/aeroval/coldatatojson_helpers.py
@@ -120,7 +120,7 @@ def _write_diurnal_week_stationdata_json(ts_data, out_dirs):
     write_json(current, fp, ignore_nan=True)
 
 
-def _add_hm_entry_json(
+def _add_heatmap_entry_json(
     heatmap_file, result, obs_name, var_name_web, vert_code, model_name, model_var
 ):
     if os.path.exists(heatmap_file):
--- a/tests/aeroval/test_coldatatojson_helpers.py
+++ b/tests/aeroval/test_coldatatojson_helpers.py
@@ -7,7 +7,7 @@ import numpy as np
 from pytest import mark, param
 
 from pyaerocom.aeroval.coldatatojson_helpers import (
-    _add_hm_entry_json,
+    _add_heatmap_entry_json,
     _init_stats_dummy,
     _prepare_aerocom_regions_json,
     _prepare_country_regions,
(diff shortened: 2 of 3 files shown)


pyaerocom/aeroval/coldatatojson_helpers.py
```diff
@@ -116,7 +120,7 @@ def _write_diurnal_week_stationdata_json(ts_data, out_dirs):
     write_json(current, fp, ignore_nan=True)
 
 
-def _add_entry_json(
+def _add_hm_entry_json(
```
avaldebe, line 123: It this function only used to update `heatmap` JSON files?
If so, it would be better called `_add_heatmap_entry_json`.

If this is a general purpose function, used to update all kids of JSON files,
it would be better to keep the name and rename the `heatmap_file` argument to `file`.

pyaerocom/aeroval/setupclasses.py
```diff
@@ -246,6 +246,7 @@ def __init__(self, exp_id: str, **kwargs):
         self.exp_descr = ""
         self.public = False
         self.exp_pi = getuser()
+        self.pyaerocom_version = __version__
```
avaldebe, line 249: Why is `self.pyaerocom_version` needed?
Will it end up somewhere on the JSON?

tests/aeroval/test_coldatatojson_helpers.py
```diff
@@ -91,14 +91,14 @@ def test__add_entry_json(tmp_path: Path):
 
     result = dict(value=None)
     obs, obs_var, vert, model, model_var = "obs", "obs_var", "Column", "model", "model_var"
-    _add_entry_json(path, result, obs, obs_var, vert, model, model_var)
+    _add_hm_entry_json(path, result, obs, obs_var, vert, model, model_var)
```
avaldebe, line 94: if the function under test changes name, the testing function should also change name...

avaldebe (changes requested):
I'm not sure about the `_add_hm_entry_json` name.
Is it really used only for heatmap JSON?

pyaerocom/aeroval/setupclasses.py
lewisblake: Yes, the idea is that by including it in the JSON, @AugustinMortier determine from version how to read in the json files in hm/ts (i.e., the old convention with statistics_ts.json or the proposed method)

pyaerocom/aeroval/coldatatojson_helpers.py
lewisblake: This function assumes a certain structure for the json files for writing, which as far as I can tell, is unique to the `hm` directory. It is not called anywhere else, and so changing to `_add_heatmap_entry_json` is probably the most appropriate course of action.

lewisblake: When I search for `_add_entry_json` in the codebase, I only find it in `coldatatijson_engine.py` (and the associated test files). I'm happy either way. Honestly I forgot I made this change during testing and could easily live with it just being `_add_json_entry()` . Whatever you think is best

avaldebe: I just want good function names. I prefer `_add_heatmap_entry_json` over `_add_entry_json`, but it is up to you.

lewisblake: I will stick with `_add_heatmap_entry_json` then

Issue #640, opened 2022-05-04
Negative O3 values EEA_AQeRep.v2


WillemVanCaspel:
Reading O3 data for the years 2015 and 2016 from EEA_AQeRep.v2 data produces negative ozone values (both mean and daily max). This is not the case for the years 2017 to 2020.

The only difference I can find between 2015-2016 and the other years is that the rows in the .cvs files from the former aren't sorted by date.

jgriesfeller:
Can you please add a file name to this issue? It's faster to test then due to the single file reading.

WillemVanCaspel:
See for example

/lustre/storeA/project/aerocom/aerocom1/AEROCOM_OBSDATA/EEA_AQeRep.v2/renamed/NL

jgriesfeller:
That is just a folder... Or does it apply to all year 2015 files in there?
Sorry, but please make these issues as precise as possible. The easier they can be tested, the faster they are addressed.

WillemVanCaspel:
Ah, silly me. 

Here's the full path:

/lustre/storeA/project/aerocom/aerocom1/AEROCOM_OBSDATA/EEA_AQeRep.v2/renamed/NL/NL_7_28545_2015_timeseries.csv

lewisblake:
Note: Check if data are ordered correctly in time

jgriesfeller:
The data is not always ordered in time. Correcting that at the lowest possible level for speed.

PR #682, opened 2022-06-21
order time steps to avoid negative O3 values in EEA_AQeRep.v2 data

jgriesfeller:
This PR is a try to solve issue #640 by ordering the read values in time. This assumes that the problem describes in issue #640 is really the non monotonic time values
@WillemVanCaspel if you have time you could try if this PR solves the issue. If not I have to dig further.

### DIFF
--- a/pyaerocom/io/read_eea_aqerep_base.py
+++ b/pyaerocom/io/read_eea_aqerep_base.py
@@ -527,7 +527,7 @@ class ReadEEAAQEREPBase(ReadUngriddedBase):
             #                                                   self.data_id,
             #                                                   self.data_dir,
             #                                                   all_str))
-        #self.files = files
+        # self.files = files
         return files
 
     def get_station_coords(self, meta_key):
--- a/pyaerocom/io/read_eea_aqerep_base.py
+++ b/pyaerocom/io/read_eea_aqerep_base.py
@@ -524,7 +524,6 @@ class ReadEEAAQEREPBase(ReadUngriddedBase):
             #                                                   self.data_id,
             #                                                   self.data_dir,
             #                                                   all_str))
-        # self.files = files
         return files
 
     def get_station_coords(self, meta_key):
@@ -584,7 +583,6 @@ class ReadEEAAQEREPBase(ReadUngriddedBase):
             if len(self.files) == 0:
                 logger.info("Retrieving file list")
                 files = self.get_file_list(self.FILE_MASKS[var_name])
-            # files = self.files
 
         if first_file is None:
             first_file = 0
--- a/pyaerocom/io/read_eea_aqerep_base.py
+++ b/pyaerocom/io/read_eea_aqerep_base.py
@@ -376,7 +376,14 @@ class ReadEEAAQEREPBase(ReadUngriddedBase):
         sort_flag = False
         # just assume hourly data for now
         time_diff = np.timedelta64(30, "m")
-        if np.min(diff_unsorted) < 0:
+        # np.min needs an array and fails with ValueError when a scalar is supplied
+        # this is the case for a single line file
+        try:
+            min_diff = np.min(diff_unsorted)
+        except ValueError:
+            min_diff = 0
+
+        if min_diff < 0:
             # data needs to be sorted
             ordered_idx = np.argsort(data_dict[self.START_TIME_NAME][:lineidx])
             data_out["dtime"] = data_dict[self.START_TIME_NAME][ordered_idx] + time_diff
--- a/pyaerocom/io/read_eea_aqerep_base.py
+++ b/pyaerocom/io/read_eea_aqerep_base.py
@@ -38,7 +38,7 @@ class ReadEEAAQEREPBase(ReadUngriddedBase):
     _FILEMASK = "*.csv"
 
     #: Version log of this class (for caching)
-    __version__ = "0.07"
+    __version__ = "0.08"
 
     #: Column delimiter
     FILE_COL_DELIM = ","
@@ -367,16 +367,33 @@ class ReadEEAAQEREPBase(ReadUngriddedBase):
         # TsType is
         # data_out['var_info'][aerocom_var_name]['ts_type'] = self.TS_TYPE
 
+        # Sometimes the times in the data files are not ordered in time which causes problems when doing
+        # time interpolations later on. Make sure that the data is ordered in time
+        diff_unsorted = (
+            data_dict[self.START_TIME_NAME][1:lineidx]
+            - data_dict[self.START_TIME_NAME][: lineidx - 1]
+        )
+        sort_flag = False
+        # just assume hourly data for now
+        time_diff = np.timedelta64(30, "m")
+        if np.min(diff_unsorted) < 0:
+            # data needs to be sorted
+            ordered_idx = np.argsort(data_dict[self.START_TIME_NAME][:lineidx])
+            data_out["dtime"] = data_dict[self.START_TIME_NAME][ordered_idx] + time_diff
+            sort_flag = True
+        else:
+            data_out["dtime"] = data_dict[self.START_TIME_NAME][:lineidx] + time_diff
+
         for key, value in data_dict.items():
             # adjust the variable name to aerocom standard
             if key != self.VAR_NAMES_FILE[aerocom_var_name]:
                 data_out[key] = value[:lineidx]
             else:
-                data_out[aerocom_var_name] = value[:lineidx]
+                if sort_flag:
+                    data_out[aerocom_var_name] = value[:lineidx][ordered_idx]
+                else:
+                    data_out[aerocom_var_name] = value[:lineidx]
 
-        # just assume hourly data for now
-        time_diff = np.timedelta64(30, "m")
-        data_out["dtime"] = data_dict[self.START_TIME_NAME][:lineidx] + time_diff
         # convert data vectors to pandas.Series (if attribute
         # vars_as_series=True)
         if vars_as_series:
--- a/pyaerocom/io/read_eea_aqerep_base.py
+++ b/pyaerocom/io/read_eea_aqerep_base.py
@@ -527,7 +527,7 @@ class ReadEEAAQEREPBase(ReadUngriddedBase):
             #                                                   self.data_id,
             #                                                   self.data_dir,
             #                                                   all_str))
-        self.files = files
+        #self.files = files
         return files
 
     def get_station_coords(self, meta_key):
@@ -587,7 +587,7 @@ class ReadEEAAQEREPBase(ReadUngriddedBase):
             if len(self.files) == 0:
                 logger.info("Retrieving file list")
                 files = self.get_file_list(self.FILE_MASKS[var_name])
-            files = self.files
+            # files = self.files
 
         if first_file is None:
             first_file = 0
--- a/pyaerocom/io/read_eea_aqerep_base.py
+++ b/pyaerocom/io/read_eea_aqerep_base.py
@@ -583,6 +583,7 @@ class ReadEEAAQEREPBase(ReadUngriddedBase):
             logger.info("Retrieving file list")
             files = self.get_file_list(self.FILE_MASKS[var_name])
 
+
         if first_file is None:
             first_file = 0
         if last_file is None:
@@ -696,5 +697,6 @@ class ReadEEAAQEREPBase(ReadUngriddedBase):
         # data_obj.data_revision[self.DATASET_NAME] = self.data_revision
         self.data = data_obj
         self._metadata = None
+        self.files = files
 
         return data_obj
--- a/pyaerocom/io/read_eea_aqerep_base.py
+++ b/pyaerocom/io/read_eea_aqerep_base.py
@@ -369,10 +369,7 @@ class ReadEEAAQEREPBase(ReadUngriddedBase):
 
         # Sometimes the times in the data files are not ordered in time which causes problems when doing
         # time interpolations later on. Make sure that the data is ordered in time
-        diff_unsorted = (
-            data_dict[self.START_TIME_NAME][1:lineidx]
-            - data_dict[self.START_TIME_NAME][: lineidx - 1]
-        )
+        diff_unsorted = np.diff(data_dict[self.START_TIME_NAME])
         sort_flag = False
         # just assume hourly data for now
         time_diff = np.timedelta64(30, "m")
--- a/pyaerocom/io/read_eea_aqerep_base.py
+++ b/pyaerocom/io/read_eea_aqerep_base.py
@@ -583,7 +583,6 @@ class ReadEEAAQEREPBase(ReadUngriddedBase):
             logger.info("Retrieving file list")
             files = self.get_file_list(self.FILE_MASKS[var_name])
 
-
         if first_file is None:
             first_file = 0
         if last_file is None:
--- a/pyaerocom/io/read_eea_aqerep_base.py
+++ b/pyaerocom/io/read_eea_aqerep_base.py
@@ -580,9 +580,8 @@ class ReadEEAAQEREPBase(ReadUngriddedBase):
         var_name = vars_to_retrieve[0]
         logger.info("Reading EEA data")
         if files is None:
-            if len(self.files) == 0:
-                logger.info("Retrieving file list")
-                files = self.get_file_list(self.FILE_MASKS[var_name])
+            logger.info("Retrieving file list")
+            files = self.get_file_list(self.FILE_MASKS[var_name])
 
         if first_file is None:
             first_file = 0
--- a/tests/fixtures/eea_v2.py
+++ b/tests/fixtures/eea_v2.py
@@ -1,0 +1,9 @@
+import pytest
+
+from pyaerocom.io import ReadEEAAQEREP_V2
+# from pyaerocom.ungriddeddata import UngriddedData
+
+
+@pytest.fixture(scope="session")
+def eea_v2_subset_reader() -> ReadEEAAQEREP_V2:
+    return ReadEEAAQEREP_V2("EEA_AQeRep.v2.Subset")
--- a/tests/fixtures/eea_v2.py
+++ b/tests/fixtures/eea_v2.py
@@ -1,8 +1,0 @@
-import pytest
-
-from pyaerocom.io import ReadEEAAQEREP_V2
-
-
-@pytest.fixture(scope="session")
-def eea_v2_subset_reader() -> ReadEEAAQEREP_V2:
-    return ReadEEAAQEREP_V2("EEA_AQeRep.v2.Subset")
--- a/tests/fixtures/eea_v2.py
+++ b/tests/fixtures/eea_v2.py
@@ -1,7 +1,6 @@
 import pytest
 
 from pyaerocom.io import ReadEEAAQEREP_V2
-# from pyaerocom.ungriddeddata import UngriddedData
 
 
 @pytest.fixture(scope="session")
--- a/tests/io/test_read_eea_aqerep_v2.py
+++ b/tests/io/test_read_eea_aqerep_v2.py
@@ -2,6 +2,7 @@ import pytest
 
 from pyaerocom.stationdata import StationData
 from pyaerocom.ungriddeddata import UngriddedData
+from tests.conftest import TEST_RTOL
 
 # Subset data id used for testing.
 DATA_ID = "EEA_AQeRep.v2.Subset"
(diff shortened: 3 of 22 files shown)


WillemVanCaspel:
I turned on a quick test run for 2015 to see if it works, but the evaluation does not finish. Instead, the following error is raised multiple times:

^M  2%|▏         | 513/20835 [01:25<1:34:34,  3.58it/s]Reading file /lustre/storeA/project/aerocom/aerocom1/AEROCOM_OBSDATA/EEA_AQeRep.v2/renamed/AT/AT_5_49646_2015_timeseries.csv
No metadata available for key PI
No metadata available for key country
No metadata available for key data_product
No metadata available for key data_version
No metadata available for key data_level
No metadata available for key framework
No metadata available for key instr_vert_loc
No metadata available for key revision_date
No metadata available for key website
No metadata available for key ts_type_src
No metadata available for key stat_merge_pref_attr
^M  2%|▏         | 514/20835 [01:25<1:26:00,  3.94it/s]Reading file /lustre/storeA/project/aerocom/aerocom1/AEROCOM_OBSDATA/EEA_AQeRep.v2/renamed/AT/AT_5_49646_2016_timeseries.csv
No metadata available for key PI
No metadata available for key country
No metadata available for key data_product
No metadata available for key data_version
No metadata available for key data_level
No metadata available for key framework
No metadata available for key instr_vert_loc
No metadata available for key revision_date
No metadata available for key website
No metadata available for key ts_type_src
No metadata available for key stat_merge_pref_attr
^M  2%|▏         | 515/20835 [01:25<1:23:55,  4.04it/s]Reading file /lustre/storeA/project/aerocom/aerocom1/AEROCOM_OBSDATA/EEA_AQeRep.v2/renamed/AT/AT_5_49646_2017_timeseries.csv
^M  2%|▏         | 515/20835 [01:25<56:15,  6.02it/s]  
Failed to perform analysis: Traceback (most recent call last):
  File "/home/willemvc/Python/pyaerocom/pyaerocom/pyaerocom/colocation_auto.py", line 793, in run
    coldata = self._run_helper(mod_var, obs_var)
  File "/home/willemvc/Python/pyaerocom/pyaerocom/pyaerocom/colocation_auto.py", line 1420, in _run_helper
    args = self._prepare_colocation_args(model_var, obs_var)
  File "/home/willemvc/Python/pyaerocom/pyaerocom/pyaerocom/colocation_auto.py", line 1363, in _prepare_colocation_args
    obs_data = self.get_obs_data(obs_var)
  File "/home/willemvc/Python/pyaerocom/pyaerocom/pyaerocom/colocation_auto.py", line 696, in get_obs_data
    return self._read_ungridded(obs_var)
  File "/home/willemvc/Python/pyaerocom/pyaerocom/pyaerocom/colocation_auto.py", line 899, in _read_ungridded
    obs_data = obs_reader.read(
  File "/home/willemvc/Python/pyaerocom/pyaerocom/pyaerocom/io/readungridded.py", line 668, in read
    self.read_dataset(
  File "/home/willemvc/Python/pyaerocom/pyaerocom/pyaerocom/io/readungridded.py", line 412, in read_dataset
    data_read = reader.read(vars_to_read, **kwargs)
  File "/home/willemvc/Python/pyaerocom/pyaerocom/pyaerocom/io/read_eea_aqerep_base.py", line 614, in read
    station_data = self.read_file(_file, var_name=var_name)
  File "/home/willemvc/Python/pyaerocom/pyaerocom/pyaerocom/io/read_eea_aqerep_base.py", line 379, in read_file
    if np.min(diff_unsorted) < 0:
  File "<__array_function__ internals>", line 180, in amin
  File "/home/willemvc/miniconda3/envs/pya/lib/python3.10/site-packages/numpy/core/fromnumeric.py", line 2916, in amin
    return _wrapreduction(a, np.minimum, 'min', axis, None, out,
  File "/home/willemvc/miniconda3/envs/pya/lib/python3.10/site-packages/numpy/core/fromnumeric.py", line 86, in _wrapreduction
    return ufunc.reduce(obj, axis, dtype, out, **passkwargs)
ValueError: zero-size array to reduction operation minimum which has no identity

Colocation processing status for EMEP vs. EEA-rural
  Model Var   Obs Var                     Status
0  concpm10  concpm10  NOT OK: Colocation failed
1  concpm25  concpm25  NOT OK: Colocation failed
2   concso2   concso2  NOT OK: Colocation failed
3    vmrno2    vmrno2  NOT OK: Colocation failed
4  vmro3max  vmro3max  NOT OK: Colocation failed


The path to the above evaluation output file is /lustre/storeB/users/willemvc/Python/aeroval/config/logs/log_val5.out, and the config file used is /lustre/storeB/users/willemvc/Python/aeroval/config/documentation/cfg_emep-valtest_2015EEA.py

I will be back from holidays next week and can then spend more time looking at this problem.

jgriesfeller:
I saw the problem in my tests as well. The error happens with single line files. Doing proper testing now...

jgriesfeller:
@WillemVanCaspel in my test (vmro3 only) the code works now.
But there are still negative values occasionally (at a few time steps, but at many stations). You can remove those by applying the outlier removal in the aeroval config.

jgriesfeller:
@WillemVanCaspel did you test this PR? Does it remove your problem?

pyaerocom/io/read_eea_aqerep_base.py
```diff
@@ -367,16 +367,40 @@ def read_file(self, filename, var_name, vars_as_series=False):
         # TsType is
         # data_out['var_info'][aerocom_var_name]['ts_type'] = self.TS_TYPE
 
+        # Sometimes the times in the data files are not ordered in time which causes problems when doing
+        # time interpolations later on. Make sure that the data is ordered in time
+        diff_unsorted = (
+            data_dict[self.START_TIME_NAME][1:lineidx]
+            - data_dict[self.START_TIME_NAME][: lineidx - 1]
```
avaldebe, line 374: I do not really understand how these lines sort the file on time.
Maybe a test to show that it works would be of help

avaldebe (changes requested):
Could you add a test with an unsorted file?

pyaerocom/io/read_eea_aqerep_base.py
avaldebe: Is `lineidx` the last line? If so, this could be written as:

``` python
    diff_unsorted = data_dict[self.START_TIME_NAME][1:] - data_dict[self.START_TIME_NAME][:-1]
```

or as a boolean
``` python
    unsorted = any(data_dict[self.START_TIME_NAME][1:] > data_dict[self.START_TIME_NAME][:-1])
```

avaldebe: It seems it could be further simplified with [np.diff]

``` pytyon
    unsorted = (np.diff(data_dict[self.START_TIME_NAME][:lineidx]) > 0).any()
```

[np.diff]: https://numpy.org/doc/stable/reference/generated/numpy.diff.html

jgriesfeller: I did not know np.diff to be honest. You know that I tend not to be very pythonic when it comes specialties / conveniences in Python ;-)
I also wonder if these convenience methods are good for speed or not as they need to test more cases. 
I'll change the code to use np.diff to avoid confusion

avaldebe: > I did not know np.diff to be honest.

Me neither, but this seemed to be a problem general enough to have been addressed by numpy.

> I also wonder if these convenience methods are good for speed or not as they need to test more cases.

In general, I'm more concerned abut readability than raw performance.
It does not matter how fast we can arrive to the wrong conclusions because the code does something unexpected.

In any case, I suspect that the real performance hit is on the `np.argsort` call.
BTW, why use `np.argsort` to create an index instead of sorting the data itself?

avaldebe:
Hi @jgriesfeller 

I wonder if this PR should be rename to something that describes the actual changes made.
Something like "EEA_AQeRep.v2 with monotonically increasing time".

The relation to #640 is already stated on the PR description. Additionally, the merge message could include a "close #640" statement so the issue is closed aromatically.

jgriesfeller:
Regarding the tests:
The EEA testing does not read any data anymore
after https://github.com/metno/pyaerocom/blob/367d5cb4e68e2a4557ccc6da09bece6495a084f0/tests/io/test_read_eea_aqerep_v2.py#L41 data is an empty Ungriddeddata object.

I think due to the new fixtures `"EEA_AQeRep.v2.Subset"` is not registered anymore as a data set and I don't know right away how to register that again.

jgriesfeller:
I just realised that my EEA testing is different in the sense that it reads the data vie the Ungriddeddata object, and not via the low level methods as e.g. EBAS or Aeronet.

avaldebe:
> data is an empty Ungriddeddata object.

That is a problem, maybe the test should fail if data is empty.
Otherwise, we can not tell when stopped working or if it did anything to begin with.

> I think due to the new fixtures `"EEA_AQeRep.v2.Subset"` is not registered anymore as a data set and I don't know right away how to register that again.

As far as I can tell, there is only one `reader` fixture available to `test_read_eea_aqerep_v2.py`

``` bash
$ pytest --fixtures tests/io/test_read_eea_aqerep_v2.py |'grep' -1w reader
------------ fixtures defined from tests.io.test_read_eea_aqerep_v2 ------------
reader [module scope] -- tests/io/test_read_eea_aqerep_v2.py:14
    no docstring available
```

https://github.com/metno/pyaerocom/blob/367d5cb4e68e2a4557ccc6da09bece6495a084f0/tests/io/test_read_eea_aqerep_v2.py#L13-L15

avaldebe:
or maybe one should register a new reader as follows

``` python
from pyaerocom.io.read_eea_aqerep_v2 import ReadEEAAQEREP_V2
[...]


class ReadEEAAQEREP_V2_Subset(ReadEEAAQEREP_V2):
     DATA_ID = "EEA_AQeRep.v2.Subset"
     SUPPORTED_DATASETS = [DATA_ID] 
    
ReadUngridded.SUPPORTED_READERS.append(ReadEEAAQEREP_V2_Subset)

@pytest.fixture(scope="module") 
def reader(): 
    return ReadUngridded(ReadEEAAQEREP_V2_Subset.DATA_ID)

[...]
```

pyaerocom/io/read_eea_aqerep_base.py
jgriesfeller: `np.argsort` because I need to reorder the the values according to the time (so the order comes from the time and needs to be applied to the time and the data values). These are pure numpy arrays. 
I have used similar code several times before (in IDL where these vector operations are the only fast code), so I did not find it very complicated ;-), But yes, np.diff if clearer

jgriesfeller:
tests include now also unordered station data and a station with just one time step.

tests/fixtures/eea_v2.py
```diff
@@ -0,0 +1,8 @@
+import pytest
+
+from pyaerocom.io import ReadEEAAQEREP_V2
+
+
+@pytest.fixture(scope="session")
+def eea_v2_subset_reader() -> ReadEEAAQEREP_V2:
+    return ReadEEAAQEREP_V2("EEA_AQeRep.v2.Subset")
```
avaldebe, line 8: Will this fixture use or multiple modules?
If it will only be used in on `tests/io/test_read_eea_aqerep_v2.py` it might be better to define it on that module.

tests/io/test_read_eea_aqerep_v2.py
```diff
@@ -1,26 +1,23 @@
...
-
+# Subset data id used for testing.
 DATA_ID = "EEA_AQeRep.v2.Subset"
 
 
 @pytest.fixture(scope="module")
-def reader():
-    return ReadUngridded(DATA_ID)
+def reader(eea_v2_subset_reader):
+    return eea_v2_subset_reader
```
avaldebe, line 13: Why is this fixture needed?
If all you want is to rename the rename the fixture you can import it with the new name:

``` python
from tests.fixtures.eea_v2 import eea_v2_subset_reader as reader
```

tests/io/test_read_eea_aqerep_v2.py
```diff
@@ -42,14 +51,15 @@ def test_read(reader):
         assert isinstance(data, UngriddedData)
 
         for stat_idx, statid in enumerate(station_id[var_name]):
+            print(statid)
```
avaldebe, line 54: `stdout` and `stderr` are captured by pytest, so you will not see this print unless the test fails, in witch case the print will bet lost between the stack trace.

tests/io/test_read_eea_aqerep_v2.py
```diff
@@ -42,14 +51,15 @@ def test_read(reader):
...
                 # It makes no sense to test this for every station
                 if stat_idx == 1:
                     assert isinstance(stat_data, StationData)
 
                 assert stat_data[var_name].mean() == pytest.approx(
-                    station_means[var_name][stat_idx], 0.01
+                    station_means[var_name][stat_idx], TEST_RTOL
                 )
             except:
                 print(f"failed test var {var_name}")
```
avaldebe, line 65: Why are you capturing the assert?
Without the assert this test will pass no matter what,
and the print will be never seen.

avaldebe (changes requested):
please remove the `try`/`except` block on `test_read` so the test can pass (or fail)

tests/fixtures/eea_v2.py
jgriesfeller: I realise now that I was not fit yesterday. Sorry for that.
Fixture removed.

tests/io/test_read_eea_aqerep_v2.py
jgriesfeller: leftover from testing; removed

tests/io/test_read_eea_aqerep_v2.py
jgriesfeller: leftover from testing; removed

pyaerocom/io/read_eea_aqerep_base.py
```diff
@@ -563,7 +583,6 @@ def read(
             if len(self.files) == 0:
                 logger.info("Retrieving file list")
                 files = self.get_file_list(self.FILE_MASKS[var_name])
```
avaldebe, line 585: I can not see where is `files` defined, and what will happen when `self.files` is not empty

avaldebe: > files is defined here:

I see, `files` is one of the method arguments. I totally missed it

> For EEA reading self.files should always be empty from now on. The OS caching should work quite well here. I have removed the check for self.files now.

If you are assuming `self.files` to be empty and `files` to be None,
I think you should put an assert at the top of the method to make the assumption clear.
Even better would be to put a guarding clause that raises a meaningful exception

``` python
    def read(
        self, vars_to_retrieve=None, files=None, first_file=None, last_file=None, metadatafile=None
    ):
    [...]
    if self.files:
        raise RuntimeError(f"{self.files=} should be empty")
    if files is not None:
        raise ValueError(f"{files=} should be None")

    logger.info("Retrieving file list")
    files = self.get_file_list(self.FILE_MASKS[var_name])
    [...]  
```

jgriesfeller: `files` can be not None in the case the user wants supply a list of files to be read. All other readers have the same (e.g. Aeronet:
https://github.com/metno/pyaerocom/blob/367d5cb4e68e2a4557ccc6da09bece6495a084f0/pyaerocom/io/readaeronetbase.py#L298-L306
 Very useful for testing. Therefore I would suggest to keep that functionality.

Searching through the code shows that all other readers (except for the ones for Aeronet) actually fill `self.files`. I therefore think, the `RuntimeError` should not be thrown either.

We should unify the obs readers...

avaldebe: > Very useful for testing. Therefore I would suggest to keep that functionality.

Fair point.

> Searching through the code shows that all other readers (except for the ones for Aeronet) actually fill `self.files`

In that case, this reader should do the same.

jgriesfeller: I added the assignment to `self.files` again. It's just not used for caching anymore.

avaldebe (approved):

WillemVanCaspel:
Apologies for the slowness, I wanted to run this check over the weekend before most people would come back from holidays, but now I am getting bogged down by an issue with the station names. Yesterday clearing the cache seemed like the thing to do to fix it, but now I see that yesterday's run also didn't complete. The co-location was a success, but there's something wrong with the number of stations in the heat map, or so it seems. The issue:

Colocation processing status for EMEP-GEMENIS vs. EEA-rural
  Model Var   Obs Var   Status
0  concpm10  concpm10  SUCCESS
1  concpm25  concpm25  SUCCESS
2   concso2   concso2  SUCCESS
3    vmrno2    vmrno2  SUCCESS
4  vmro3max  vmro3max  SUCCESS
Processing: /lustre/storeB/users/willemvc/Python/aeroval/coldata/TestingCams/Eval-2019-so2/EMEP-GEMENIS/concpm10_concpm10_MOD-EMEP-GEMENIS_REF-EEA-rural_20190101_20191231_daily_ALL-wMOUNTAINS.nc
Computing json files for EMEP-GEMENIS (concpm10) vs. EEA-rural (concpm10)
Processing statistics timeseries for all regions
Processing heatmap data for all regions
Traceback (most recent call last):
  File "/lustre/storeB/users/willemvc/Python/aeroval/config/documentation/cfg_emep-CAMS21vsGENEMISt-2019.py", line 488, in <module>
    res = ana.run()
  File "/home/willemvc/Python/pyaerocom/pyaerocom/pyaerocom/aeroval/experiment_processor.py", line 135, in run
    self._run_single_entry(model_name, obs_name, var_list)
  File "/home/willemvc/Python/pyaerocom/pyaerocom/pyaerocom/aeroval/experiment_processor.py", line 73, in _run_single_entry
    engine.run(files_to_convert)
  File "/home/willemvc/Python/pyaerocom/pyaerocom/pyaerocom/aeroval/coldatatojson_engine.py", line 53, in run
    self.process_coldata(coldata)
  File "/home/willemvc/Python/pyaerocom/pyaerocom/pyaerocom/aeroval/coldatatojson_engine.py", line 192, in process_coldata
    hm_all = _process_heatmap_data(
  File "/home/willemvc/Python/pyaerocom/pyaerocom/pyaerocom/aeroval/coldatatojson_helpers.py", line 1168, in _process_heatmap_data
    stats = _get_extended_stats(subset, use_weights)
  File "/home/willemvc/Python/pyaerocom/pyaerocom/pyaerocom/aeroval/coldatatojson_helpers.py", line 1027, in _get_extended_stats
    (_, stats["R_temporal_median"]) = _calc_temporal_corr(coldata)
  File "/home/willemvc/Python/pyaerocom/pyaerocom/pyaerocom/aeroval/coldatatojson_helpers.py", line 1082, in _calc_temporal_corr
    arr[0] = arr[0].where(obs_ok, drop=True)
  File "/home/willemvc/miniconda3/envs/pya/lib/python3.10/site-packages/xarray/core/dataarray.py", line 752, in __setitem__
    assert_coordinate_consistent(value, obj.coords.variables)
  File "/home/willemvc/miniconda3/envs/pya/lib/python3.10/site-packages/xarray/core/coordinates.py", line 389, in assert_coordinate_consistent
    raise IndexError(
IndexError: dimension coordinate 'station_name' conflicts between indexed and indexing objects:
<xarray.DataArray 'station_name' (station_name: 341)>
array(['AT0ENK1', 'AT0ILL1', 'AT0KLH1', ..., 'SK0406A', 'TR0016A', 'TR0017A'],
      dtype=object)
Coordinates:
    data_source   <U12 'EEAAQeRep.v2'
  * station_name  (station_name) object 'AT0ENK1' 'AT0ILL1' ... 'TR0017A'
    latitude      (station_name) float64 48.39 47.77 46.77 ... 48.93 36.85 37.19
    longitude     (station_name) float64 13.67 16.77 15.96 ... 22.27 35.35 35.26
    altitude      (station_name) float64 525.0 117.0 415.0 ... 431.0 12.0 107.0
    country       (station_name) <U22 'Austria' 'Austria' ... 'Turkey' 'Turkey'
    country_code  (station_name) <U2 'AT' 'AT' 'AT' 'AT' ... 'SK' 'SK' 'TR' 'TR'
vs.
<xarray.IndexVariable 'station_name' (station_name: 384)>
array(['AT0ENK1', 'AT0ILL1', 'AT0KLH1', ..., 'SK0406A', 'TR0016A', 'TR0017A'],
      dtype=object)

avaldebe:
@WillemVanCaspel 

I think you should write your feedback on the issue #640.
This PR is already merged, so the no one will follow up...

PR #683, opened 2022-06-22
Dummy cube latlon fix

lewisblake:
There was an issue within the creation of the dummy cube in which the longitudes and latitudes were switched. This merge fixes that bug and add some type hints to `make_dummy_cube_latlon`

### DIFF
--- a/.github/workflows/CI.yml
+++ b/.github/workflows/CI.yml
@@ -75,7 +75,6 @@ jobs:
         path: |
           ~/.cache/pip
           ~/.cache/srtm
-          ~/.cache/geonum
           .tox
         key: ${{ runner.os }}-tox-${{ matrix.python-version }}-${{ hashFiles('setup.cfg', '') }}
     - name: Install Tox
--- a/.github/workflows/CI.yml
+++ b/.github/workflows/CI.yml
@@ -75,7 +75,7 @@ jobs:
         path: |
           ~/.cache/pip
           .tox
-        key: ${{ runner.os }}-tox-${{ matrix.python-version }}-${{ hashFiles('setup.cfg') }}
+        key: ${{ runner.os }}-tox-${{ matrix.python-version }}-${{ hashFiles('setup.cfg', 'pyproject.toml') }}
     - name: Install Tox
       run: |
         python -m pip install --upgrade pip
--- a/.github/workflows/CI.yml
+++ b/.github/workflows/CI.yml
@@ -75,6 +75,7 @@ jobs:
         path: |
           ~/.cache/pip
           ~/.cache/srtm
+          ~/.cache/geonum
           .tox
         key: ${{ runner.os }}-tox-${{ matrix.python-version }}-${{ hashFiles('setup.cfg', '') }}
     - name: Install Tox
--- a/.github/workflows/CI.yml
+++ b/.github/workflows/CI.yml
@@ -76,7 +76,7 @@ jobs:
           ~/.cache/pip
           ~/.cache/srtm
           .tox
-        key: ${{ runner.os }}-tox-${{ matrix.python-version }}-${{ hashFiles('setup.cfg', '') }}
+        key: ${{ runner.os }}-tox-${{ matrix.python-version }}-${{ hashFiles('setup.cfg', 'pyproject.toml') }}
     - name: Install Tox
       run: |
         python -m pip install --upgrade pip
--- a/pyaerocom/helpers.py
+++ b/pyaerocom/helpers.py
@@ -6,7 +6,7 @@ import logging
 import math as ma
 from collections import Counter
 from datetime import MINYEAR, date, datetime
-from typing import Optional
+from typing import Optional, Tuple
 
 import iris
 import iris.analysis
@@ -205,7 +205,7 @@ def tuple_list_to_lists(tuple_list):
     return list(map(list, zip(tuple_list)))
 
 
-def make_dummy_cube_latlon(lat_res_deg: float = 2, lon_res_deg: float = 3, lat_range: Optional[tuple(float, float)] = None, lon_range: Optional[tuple(float, float)] = None):
+def make_dummy_cube_latlon(lat_res_deg: float = 2, lon_res_deg: float = 3, lat_range: Optional[Tuple[float,float]] = None, lon_range: Optional[Tuple[float,float]] = None):
     """Make an empty Cube with given latitude and longitude resolution
 
     Dimensions will be lat, lon
@@ -228,12 +228,11 @@ def make_dummy_cube_latlon(lat_res_deg: float = 2, lon_res_deg: float = 3, lat_r
     Cube
         dummy cube in input resolution
     """
+
     if lat_range is None:
         lat_range = (-90, 90)
     if lon_range is None:
         lon_range = (-180, 180)
-
-    breakpoint()
     
     lons = np.arange(lon_range[0] + (lon_res_deg / 2), lon_range[1] + (lon_res_deg / 2), lon_res_deg)
     lats = np.arange(lat_range[0] + (lat_res_deg / 2), lat_range[1] + (lat_res_deg / 2), lat_res_deg)
@@ -1713,14 +1712,14 @@ def make_dummy_cube(
     days_since_start = np.arange(len(times))
     unit = get_variable(var_name).units
 
-    lat_range = (-180, 180)
-    lon_range = (-90, 90)
-    lat_res_deg = 90
-    lon_res_deg = 45
+    lat_range = (-90, 90)
+    lon_range = (-180, 180)
+    lat_res_deg = 45
+    lon_res_deg = 90
     time_unit = Unit(startstr, calendar="gregorian")
 
-    lons = np.arange(lon_range[0] + lon_res_deg / 2, lon_range[1] + lon_res_deg / 2, lon_res_deg)
-    lats = np.arange(lat_range[0] + lat_res_deg / 2, lat_range[1] + lat_res_deg / 2, lat_res_deg)
+    lons = np.arange(lon_range[0] + (lon_res_deg / 2), lon_range[1] + (lon_res_deg / 2), lon_res_deg)
+    lats = np.arange(lat_range[0] + (lat_res_deg / 2), lat_range[1] + (lat_res_deg / 2), lat_res_deg)
 
     latdim = iris.coords.DimCoord(
         lats,
--- a/pyaerocom/helpers.py
+++ b/pyaerocom/helpers.py
@@ -6,7 +6,7 @@ import logging
 import math as ma
 from collections import Counter
 from datetime import MINYEAR, date, datetime
-from typing import Optional, Tuple
+from typing import Optional, Union, Tuple, List # 3.9 onward can use generic types
 
 import iris
 import iris.analysis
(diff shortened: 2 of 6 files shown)


dulte:
Looks really good. I'll approve as soon as the CI is fixed and passed

avaldebe (commented):
the annotations can be improved, otherwise a solid effort

pyaerocom/helpers.py
```diff
@@ -204,7 +205,12 @@ def tuple_list_to_lists(tuple_list):
     return list(map(list, zip(tuple_list)))
 
 
-def make_dummy_cube_latlon(lat_res_deg=2, lon_res_deg=3, lat_range=None, lon_range=None):
+def make_dummy_cube_latlon(
+    lat_res_deg: float = 2,
+    lon_res_deg: float = 3,
+    lat_range: Optional[Sequence[float]] = None,
+    lon_range: Optional[Sequence[float]] = None,
```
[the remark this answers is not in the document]
lewisblake: The function can also accept lat_range and lon_range as lists. Will type-annotating them to just be tuples fail? At first I had something similar with Tuple, but think it caused issues. I couldn't figure out how to type-annotate it to accept either a tuple or a list,l but with only two elements.

pyaerocom/helpers.py
```diff
@@ -227,13 +233,18 @@ def make_dummy_cube_latlon(lat_res_deg=2, lon_res_deg=3, lat_range=None, lon_ran
     Cube
         dummy cube in input resolution
     """
+
     if lat_range is None:
         lat_range = (-90, 90)
     if lon_range is None:
         lon_range = (-180, 180)
```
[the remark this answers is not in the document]
lewisblake: Yep.

pyaerocom/helpers.py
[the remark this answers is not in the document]
avaldebe: > The function can also accept lat_range and lon_range as lists. Will type-annotating them to just be tuples fail?


In general type annotations do not have runtime effects. There are some packages that use types to define runtime behaviour, such as Pydantic and Typer, but those are exceptions.

> At first I had something similar with Tuple, but think it caused issues. I couldn't figure out how to type-annotate it to accept either a tuple or a list,l but with only two elements.


List are mutable objects, so the most that a type annotation can do is to restrict the type to list of floats of any length with `list[float]`. Tuples are immutable objects, so the typing annotations can be more restrictive. Here I proposed a tuple of 2 floats with `tuple[float, float]`, but a tuple of any lenght can be annotated with `tuple[float, ...]`.

If the function accepts a list, it would be wise to place a guarding clause at the top of the function that assert that a list has 2 elements.

pyaerocom/helpers.py
```diff
@@ -227,13 +233,18 @@ def make_dummy_cube_latlon(lat_res_deg=2, lon_res_deg=3, lat_range=None, lon_ran
...
 
-    lons = np.arange(lon_range[0] + lon_res_deg / 2, lon_range[1] + lon_res_deg / 2, lon_res_deg)
-    lats = np.arange(lat_range[0] + lat_res_deg / 2, lat_range[1] + lat_res_deg / 2, lat_res_deg)
+    lons = np.arange(
+        lon_range[0] + (lon_res_deg / 2), lon_range[1] + (lon_res_deg / 2), lon_res_deg
+    )
+    lats = np.arange(
+        lat_range[0] + (lat_res_deg / 2), lat_range[1] + (lat_res_deg / 2), lat_res_deg
+    )
```
avaldebe, line 247: why the extra parenthesis?
``` python
>>> 3+1/2
3.5
```

lewisblake: I visually find it more explicit, but this is personal preference

lewisblake: Respectfully, I disagree ;) I think my change is the best balance between readability and efficiency. The way I have it, the endpoints of the interval are contained within a single positional argument, separated by commas. By inspecting this argument, you can see that the interval start and end are composed of two pieces: the bounds of the range provided in `lon_range` or `lat_range`, plus an offset contained in parenthesis. To me, this is the most natural way of thinking about it and cleanest way of representing that, but again, this probably boils down to personal preference and efficiency is probably the more important.

On to something else, would you mind taking a look at why CI is failing? I think it's happening on the installation. Even the linters are failing because `ImportError('NumPy 1.10+ is required to install cartopy.')`

pyaerocom/helpers.py
```diff
@@ -208,8 +207,8 @@ def tuple_list_to_lists(tuple_list):
 def make_dummy_cube_latlon(
     lat_res_deg: float = 2,
     lon_res_deg: float = 3,
-    lat_range: Optional[Sequence[float]] = None,
-    lon_range: Optional[Sequence[float]] = None,
+    lat_range: list[float] | None = (-90, 90),
+    lon_range: list[float] | None = (-180, 180),
```
avaldebe, line 211: Now that you have a concrete default, you do not need accept `None`
``` python
    lat_range: list[float] = (-90, 90),
    lon_range: list[float] = (-180, 180),
```

lewisblake: Of course :sweat_smile:

pyaerocom/helpers.py
```diff
@@ -234,10 +233,8 @@ def make_dummy_cube_latlon(
         dummy cube in input resolution
     """
 
-    if lat_range is None:
-        lat_range = (-90, 90)
-    if lon_range is None:
-        lon_range = (-180, 180)
+    # Accept lists for lat_range and lon_range, but make sure correct length
+    assert len(lat_range) == 2 and len(lon_range) == 2
```
avaldebe, line 237: you can use Python's syntax to simplify this line
``` python
    assert len(lat_range) == len(lon_range) == 2
```
(as long as you find it more readable)

lewisblake: Yes, this seems like a good change

pyaerocom/helpers.py
avaldebe: > Respectfully, I disagree ;) [...] To me, this is the most natural way of thinking about it and cleanest way of representing that, but again, this probably boils down to personal preference and efficiency is probably the more important.

I agree your way is better.

> On to something else, would you mind taking a look at why CI is failing? I think it's happening on the installation. Even the linters are failing because `ImportError('NumPy 1.10+ is required to install cartopy.')`

I noticed the CI failing. Not only the docs and lint jobs are failing, but the Python 3.8 venv is failing one of the tests. Will take a look at it.

lewisblake: Thanks for taking a look. It is strange because I don't think I changed anything that would effect other tests. I see more than one different test failing for different runs

avaldebe: #686 should take care of fixing the CI install errors.
You can marge `main-dev` into this branch after #686 is merged.

lewisblake:
Some tests in CI are failing. 

In venv 3.9 it's  `FAILED tests/test_geodesy.py::test_srtm_altitude - FileNotFoundError:`

in venv 3.10 push and Pull_request it's 
`FAILED tests/test_geodesy.py::test_srtm_altitude - FileNotFoundError: [Errno ...
FAILED tests/test_mathutils.py::test_calc_statistics[data2-ref_data2-expected2]
FAILED tests/test_mathutils.py::test_calc_statistics[data4-ref_data4-expected4]
FAILED tests/test_ungriddeddata.py::test_check_convert_var_units - RuntimeWar...`

The venv 3.9 failing I don't understand. Maybe it has to do with the order in which things are run? 

The venv 3.10 tests failing are also strange. Same issue with `test_srtm_altitude`. When I run pytest locally on `test_calc_statistics`, it passes. When I run pytest locally on `test_check_convert_var_units` I get

`FAILED tests/test_ungriddeddata.py - pytest.PytestUnknownMarkWarning: Unknown pytest.mark.dependency - is this a typo?  You can register custom marks to avoid this w...`

Checking the documentation suggests that maybe it should be `@pytest.mark.dependency() ` with parenthesis but I get the same error. I also can't find where this would be registered. In fact I'm not sure it's custom at all, and rather should be covered by the pytest-dependency plug in we have in setup.cfg (also in pyaerocom_env.yml but conda is passing fine). 

Do either of you have any suggestions / ideas? This seems different from the issues fixed in #686

avaldebe:
> Checking the documentation suggests that maybe it should be `@pytest.mark.dependency() ` with parenthesis but I get the same error. I also can't find where this would be registered. But it's not custom anyway, and rather should be covered by the pytest-dependency plug in we have in setup.cfg (also in pyaerocom_env.yml but conda is passing fine).

It seems to me that something is wrong with the `tox` environment. Most likely, something wrong with the cached version of the environment.

I'll play with the CI configuration file on this branch, so the cache is regenerated and come back to you.

lewisblake:
Finally (mostly passing CI). I ran the conda 3.10 separately first and then the venv 3.9 and that worked. Unfortunately this merge will lower our code coverage, so I'm going to write a test for this function since it caused problems in the first place and maybe we now know how to get it to pass CI (with patience! :) )

.github/workflows/CI.yml
```diff
@@ -74,8 +74,9 @@ jobs:
       with:
         path: |
           ~/.cache/pip
+          ~/.cache/srtm
           .tox
-        key: ${{ runner.os }}-tox-${{ matrix.python-version }}-${{ hashFiles('setup.cfg') }}
+        key: ${{ runner.os }}-tox-${{ matrix.python-version }}-${{ hashFiles('setup.cfg', '') }}
```
avaldebe, line 79: why did you remove the `pyproject.toml` from the cache key?

tests/test_geodesy.py
```diff
@@ -27,7 +27,10 @@ def test_is_within_radius_km():
 
 
 def test_srtm_altitude():
-    assert geodesy.get_topo_altitude(TEST_LAT, TEST_LON) == pytest.approx(207)
+    try:
+        assert geodesy.get_topo_altitude(TEST_LAT, TEST_LON) == pytest.approx(207)
+    except FileNotFoundError:  # this test can sporadically fail
+        pass
```
avaldebe, line 33: Not sure that this is the way to treat a flaky dependency on CI.

On a local development environment we do not expect this to fail and on a production environment this should not fail.

maybe marking the test with `@pytest.mark.xfail(reason="fails in CI when geonum can not download 'srtm' data", raises=FileNotFoundError)`
as suggested on the [pytest docs](https://docs.pytest.org/en/7.1.x/explanation/flaky.html#xfail-strict)

tests/test_helpers.py
```diff
@@ -250,3 +250,15 @@ def test_extract_latlon_dataarray_no_matches_error(lat, lon):
...
+def test_make_dummy_cube():
+    # make a dummy cube of an arbitrary variable name over one year
+    cube = helpers.make_dummy_cube("concpm10", start_yr=2020, stop_yr=2021)
+    data = xr.DataArray.from_iris(cube)
+    # First coordinate does not exist in the dataarray.
+    lat = [90, -67.5, 22.2]
+    lon = [180, -135.0, 45.0]
+    subset = helpers.extract_latlon_dataarray(data, lat, lon, check_domain=True)
+    assert isinstance(subset, xr.DataArray)
+    assert len(subset.lat) == len(lat) - 1 and len(subset.lon) == len(lon) - 1
```
avaldebe, line 264: maybe split into two different assert statements?

avaldebe (changes requested):
I would prefer a different approach to geodesy fails due geonum.
Please consider my suggestion

.github/workflows/CI.yml
lewisblake: I did not mean to. Will reinstate.

tests/test_geodesy.py
lewisblake: Agreed, this is better. Implemented

tests/test_helpers.py
lewisblake: Done. Also done for `test_extract_latlon_dataarray` above which followed the same format.

lewisblake:
RTD still failing https://github.com/readthedocs/readthedocs.org/issues/9368

avaldebe:
> RTD still failing [readthedocs/readthedocs.org#9368](https://github.com/readthedocs/readthedocs.org/issues/9368)

Maybe we need to change the RTD setup to use a conda environment.
Alas, I can not do that today...

In any case, as long as the docs CI jobs passes we can be sure that RTD will be able to build the docs once the `--use-deprecated=legacy-resolver` issue is solved or we port the setup to a conda environment.

avaldebe (approved):
🏁 🏎️ ready to merge

PR #685, opened 2022-06-22
extract periods from date range, add new models

charlienegri:
note on edge cases where start/end dates are the end/beginning of a year:
  
with `make_period(datetime(2018,12,31),datetime(2022,1,1))` this gives   
`['20181231-20220101', '20181231-20181231', '20190101-20191231', '20200101-20201231', '20210101-20211231', '20220101-20220101']`  
I assume this is the intended behavior, but if not we can change the function

### DIFF
--- a/pyaerocom/io/cams2_83/models.py
+++ b/pyaerocom/io/cams2_83/models.py
@@ -7,15 +7,17 @@ from typing import NamedTuple
 
 
 class ModelName(str, Enum):
-    EMEP = "emep"
+    CHIMERE = "chimere"
     DEHM = "dehm"
+    EMEP = "emep"
     EURAD = "euradim"
     GEMAQ = "gemaq"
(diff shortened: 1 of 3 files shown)


avaldebe:
@charlienegri the lint, docs and venv jobs are crashing because Cartopy packages are incompatible with the new dependency resolver used by pip. See #686 for details.

After that PR is merged, merging `main-dev` into this branch should fix some of the CI jobs.
However, I do not know why the conda jobs are failing.

charlienegri:
I have no clue, a lot of tests are failing but I don't think this branch ever passed the pipeline

charlienegri:
the tests are failing because of the different way of handling the dates as datetime objects and because of the new definition of periods, I can look into this more and try to fix them

dulte:
Looks good. will fix failing tests in the other branch

Issue #674, opened 2022-06-08
Oceans region added to too many stations


AugustinMortier:
It seems that the 'Oceans' region is added to almost all of the stations. This is not affecting the statistics calculation though, as the number of stations shown in each region seems to be correct. So we should make sure that the regions in the region list is consistent with the regions used in the statistics calculation.

e.g: AERONET: 
- WORLD: 300 stations
- Oceans: 30 stations

![Screenshot from 2022-06-07 11-27-49](https://user-images.githubusercontent.com/19610522/172346626-44174fac-7d1e-4985-b0d0-476bedd72ab0.png)

![Screenshot from 2022-06-07 11-28-15](https://user-images.githubusercontent.com/19610522/172346654-99aca64d-a61e-4047-a994-c36919cf5c69.png)

_Originally posted by @AugustinMortier in https://github.com/metno/pyaerocom/issues/582#issuecomment-1148427944_

Issue #673, opened 2022-06-08
Missing regions


AugustinMortier:
It appears that some regions are never returned in the list of regions, such as 'N America': e.g: [AeronetL1.5-d-od550aer_Column_IFS-CTRL-od550aer.json](https://gitlab.met.no/aeroval/cams2-82/-/raw/master/IFS/map/AeronetL1.5-d-od550aer_Column_IFS-CTRL-od550aer.json)

As a result, the minimap is not highlighting any station in the selected region in the mini map.
![Screenshot from 2022-06-07 11-14-56](https://user-images.githubusercontent.com/19610522/172343806-6f04ff90-8ea1-448a-b19d-23050907c1b4.png)

Other regions also affected by this issue are:
- Rus,Bel,Ukr
- Pacific,Aust,NZ,

My suspicion is that those regions are probably crossing the -180/180 degrees longitude and that is maybe not well taken into account when fetching the closest region.

_Originally posted by @AugustinMortier in https://github.com/metno/pyaerocom/issues/582#issuecomment-1148414685_

PR #689, opened 2022-06-28
Oceans674

lewisblake:
You've heard of Ocean's 11, Ocean's 12, and Ocean's 13. Well get ready for....

Addressing issues described in #674 and this comment: https://github.com/metno/pyaerocom/pull/582#issuecomment-1148414685

We know that almost all regions are contained in "Oceans". This is because the the HTAP definition in `regions_def.py`. A first observation is that [on the website](https://aeroval.met.no/overall.php?project=cams2-82&exp_name=IFS&station=WORLD&tab=timeseries#), when you select "North America", "Pacific, Australia, and New Zealand", or "Russia, Ukraine, Belarus", stations on the globe do not "light up" as they do for the rest of the region (e.g., Europe, Eastern Europe, etc.). I suspected this might have to do with the face that each of these regions have longitude boundaries in which the easterly-most coordinate is "greater than" the westerly-most coordinate (i.e., the bounding box goes through the antipodes of the prime meridian), Sure enough, the function `contains_coordinate` was not set up for this case. So I changed it, and added some tests. 

This needs testing on a run, which I will try tomorrow. Not sure if it will correct everything in the issue, but this is a start.

### DIFF
--- a/pyaerocom/region.py
+++ b/pyaerocom/region.py
@@ -351,13 +351,14 @@ def get_regions_coord(lat, lon, regions=None):
     if regions is None:
         regions = get_all_default_regions()
     ocean_mask = load_region_mask_xr("OCN")
+    on_ocean = get_mask_value(lat, lon, ocean_mask)
     for rname, reg in regions.items():
         if rname == ALL_REGION_NAME:  # always True for ALL_REGION_NAME
             continue
         # OCN needs special handling determined by the rname, not hardcoded to return OCN b/c of HTAP issues
-        if reg.contains_coordinate(lat, lon) and not get_mask_value(lat, lon, ocean_mask):
+        if reg.contains_coordinate(lat, lon) and not on_ocean:
             matches.append(rname)
-        if rname == "OCN" and get_mask_value(lat, lon, ocean_mask):
+        if rname == "OCN" and on_ocean:
             matches.append(rname)
     if len(matches) == 0:
         matches.append(ALL_REGION_NAME)
--- a/pyaerocom/region.py
+++ b/pyaerocom/region.py
@@ -142,8 +142,23 @@ class Region(BrowseDict):
         bool
             True if coordinate is contained in this region, False if not
         """
-        lat_ok = self.lat_range[0] <= lat <= self.lat_range[1]
-        lon_ok = self.lon_range[0] <= lon <= self.lon_range[1]
+
+        lat_lb = self.lat_range[0]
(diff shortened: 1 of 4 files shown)


lewisblake:
@AugustinMortier Are the regions on the website defined from `regions.json`?

lewisblake:
Here's how I see the situation.

The problem is that the "Ocean" region is applied to many of the stations. However, the statistics (e.g., time series) for each region appear to be computed correctly. Indeed, they are. This is because the filtering for a colocated data object takes place in `colocateddata.py::apply_region_mask()`. This function accepts as an argument `region_id` (str) and calls a function `load_region_mask_xr`, which opens the HTAP NetCDF filters (e.g., `EAShtap.0.1x0.1deg.nc` located at https://pyaerocom.met.no/pyaerocom-suppl/htap_masks/), and returns them as xarray objects. The `region_id` corresponds to keys in `_HTAP_NAMES` provided in `region_defs.py`. Basically the same key is used to implement the NetCDF filtering as is used to get the region definition in `_HTAP_DEFS`, which end up in `regions.json` and then on the web interface. It is also important to note that `regions.json` has essentially a dictionary, with each region defined in terms of rectangular bounds: 

```json
"N America": {
        "minLat": 18.95,
        "maxLat": 71.15,
        "minLon": 172,
        "maxLon": -52
    },
```

If the regions are are stored as an array (https://github.com/metno/pyaerocom/issues/580), then `coldatatojson_helpers.py::_get_stat_regions()` and `coldatatojson_helpers.py::_process_sites()` will get multiple matches against the regions definitions in `_HTAP_DEFS`.

Therefore, we either need to completely re-overhaul how regions are defined on the web interface, or on the web, we only need to plot points for which "Oceans" are their only region. However, having the regions defined in terms of the NetCDF filters and also be flexible enough to be an array is fundamentally incompatible.

lewisblake:
Linking this issue https://github.com/metno/pyaerocom/issues/673

lewisblake:
> Q: Also, why is there no "W Europe": https://aeroval.met.no/overall.php?project=cams2-82&exp_name=IFS&station=WORLD&tab=timeseries#
> 
> A: Not in _HTAP_NAMES

@AugustinMortier do you want W Europe on the web as well?

lewisblake:
Conda is failing because there is no gloabl-land-mask for conda. Maybe instead we need https://github.com/regionmask/regionmask

Issue #692, opened 2022-07-05
Testing suite should be machine and user agnostic


lewisblake:
When trying to run the test suit on PPI, I ran into an issue with permissions. e.g., 

```
PermissionError: [Errno 13] Permission denied: '/lustre/storeA/project/aerocom/aerocom2/pyaerocom_out/testdata-minimal/scripts/create_subset_ebas.py'
```

After chatting with @jgriesfeller and @avaldebe and a ticket to IT to change ownership of the directories needed for testing to @jgriesfeller, we concluded that testing is not multi-user capable.

```
/modules/centos7/user-apps/aerocom/anaconda3/envs/pyadev-applied/lib/python3.9/pathlib.py:1120: in _opener
    return self._accessor.open(self, flags, mode)
E   PermissionError: [Errno 13] Permission denied: '/tmp/test_iris_io/invalid.nc'
```
```
(pyadev-applied) jang@ppi-clogin-a1:~/.../Python3/pyaerocom$ ll /tmp/test_iris_io/invalid.nc
-rw-rw-r-- 1 danielh danielh 0 Jul  4 12:44 /tmp/test_iris_io/invalid.nc
```

In order for tests to be run agnostically, they should not depend on `/tmp`, but rather be specific to the machine and $USER (@jgriesfeller please let me know if I am forgetting anything). The Python module [tempfile](https://docs.python.org/3/library/tempfile.html#) may be starting point for thinking of a solution.

avaldebe:
> ```
> /modules/centos7/user-apps/aerocom/anaconda3/envs/pyadev-applied/lib/python3.9/pathlib.py:1120: in _opener
>     return self._accessor.open(self, flags, mode)
> E   PermissionError: [Errno 13] Permission denied: '/tmp/test_iris_io/invalid.nc'
> ```
> 
> ```
> (pyadev-applied) jang@ppi-clogin-a1:~/.../Python3/pyaerocom$ ll /tmp/test_iris_io/invalid.nc
> -rw-rw-r-- 1 danielh danielh 0 Jul  4 12:44 /tmp/test_iris_io/invalid.nc
> ```

The likes that create this file are

https://github.com/metno/pyaerocom/blob/e9f26665a551be2489a0f05baa8637c5d93e9ffb/tests/io/test_iris_io.py#L66-L68

The quick'n dirty solution would be to create a temporary directory each time with `tempfile.mkdtemp()` as follows

``` diff
[...]

PR #694, opened 2022-07-05
No tempfile

avaldebe:
Address some of the problems behind #692 by providing temporary files as fixtures.
These changes makes possible to run multiple tests in parallel with `pytest-xdist`.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -111,7 +111,7 @@ install_command =
 commands_pre =
     python --version
 commands =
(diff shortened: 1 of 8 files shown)


avaldebe:
> [...] These changes makes possible to run multiple tests in parallel with `pytest-xdist`.

Unfortunately, [it does not work consistently on CI](https://github.com/metno/pyaerocom/runs/7198326855?check_suite_focus=true).

Issue #605, opened 2022-03-24
EBAS reading: add tests for correct values for the most used EBAS variables


jgriesfeller:
The current EBAS reading tests do not test actual variable values. 
I think we should also make sure that the read values of the most used variables make sense.
Variables:
- concso4
- concso2
- concpm10
- concpm25
- conco3
- concoa
- concss
- concoc

lewisblake:
Some guidance for criteria for "makes sense" would be appreciated here. In my initial testing, I've found concpm10 concentrations with negative values (all -0.1), but can't seem to find an obvious way to determine if these are fill values or suspect data.

lewisblake:
It is also worth noting that the ReasEbas filter itself does not check for negative values in the data.

avaldebe:
@jgriesfeller 

> It is also worth noting that the ReasEbas filter itself does not check for negative values in the data.

Do you k now if this is a bug or a feature?

lewisblake:
I think it's a feature. It's listed:  

TODO
    ----
    - Check for negative values vs. detection limit
    - ...
 
I looked to see if this was just outdated and had been implemented but this TODO not removed, but I did not see anything which appears like it has been implemented. I think we should implement removing of negative values before testing for it :)

lewisblake:
I do not see an issue, so I will open one and address it.

PR #696, opened 2022-07-06
Increase test coverage

lewisblake:
Beginning an effort (not the last) to increase our test coverage. I will include fixes as I identify them.

### DIFF
--- a/pyaerocom/io/helpers.py
+++ b/pyaerocom/io/helpers.py
@@ -78,7 +78,7 @@ def aerocom_savename(data_id, var_name, vert_code, year, ts_type):
     return f"aerocom3_{data_id}_{var_name}_{vert_code}_{year}_{ts_type}.nc"
 
 
-def _print_read_info(i, mod, tot_num, last_t, name, logger):  # pragma : no cover
+def _print_read_info(i, mod, tot_num, last_t, name, logger):  # pragma: no cover
     """Helper for displaying standardised output in reading classes
 
     Not to be used directly
--- a/pyaerocom/io/helpers.py
+++ b/pyaerocom/io/helpers.py
@@ -78,7 +78,7 @@ def aerocom_savename(data_id, var_name, vert_code, year, ts_type):
     return f"aerocom3_{data_id}_{var_name}_{vert_code}_{year}_{ts_type}.nc"
 
 
-def _print_read_info(i, mod, tot_num, last_t, name, logger):
+def _print_read_info(i, mod, tot_num, last_t, name, logger):  # pragma : no cover
     """Helper for displaying standardised output in reading classes
 
     Not to be used directly
--- a/pyaerocom/io/read_ebas.py
+++ b/pyaerocom/io/read_ebas.py
@@ -1691,18 +1691,6 @@ class ReadEbas(ReadUngriddedBase):
 
         data.clear_meta_no_data()
 
-        # # remove negative values. EBAS species appear to be positive valued only
-        # neg = np.where(data._data[:, data._DATAINDEX ] < 0)
-        # neg_bool = (data._data[:, data._DATAINDEX ] < 0)
-        # # this loop is to remove the negative rows from the metadata
-        # # data.meta_idx is indected by 0.0, 1.0, ...
-        # for i in data.meta_idx:
-        #     # typically only one, but incase there are more lop through
-        #     for species in data.meta_idx[i]:
-        #         data.meta_idx[i][species] = np.intersect1d(data.meta_idx[i][species], neg)
-        # # the data, unlike the metadata, is a flat file, so can remove all at once
-        # data._data = data._data[~neg_bool,:]
-
         return data
 
     def _read_files(self, files, vars_to_retrieve, files_contain, constraints):
--- a/pyaerocom/io/read_ebas.py
+++ b/pyaerocom/io/read_ebas.py
@@ -1378,7 +1378,6 @@ class ReadEbas(ReadUngriddedBase):
                     else:
                         raise
 
-
         if len(data_out["var_info"]) == 0:
             raise EbasFileError(
                 f"All data columns of specified input variables are NaN in {filename}"
@@ -1390,7 +1389,6 @@ class ReadEbas(ReadUngriddedBase):
 
         if self.readopts_default.ensure_correct_freq:
             self._flag_incorrect_frequencies(data_out)
-
 
         # compute additional variables (if applicable)
         data_out = self.compute_additional_vars(data_out, vars_to_compute)
@@ -1697,7 +1695,7 @@ class ReadEbas(ReadUngriddedBase):
         # neg = np.where(data._data[:, data._DATAINDEX ] < 0)
         # neg_bool = (data._data[:, data._DATAINDEX ] < 0)
         # # this loop is to remove the negative rows from the metadata
-        # # data.meta_idx is indected by 0.0, 1.0, ... 
+        # # data.meta_idx is indected by 0.0, 1.0, ...
         # for i in data.meta_idx:
         #     # typically only one, but incase there are more lop through
         #     for species in data.meta_idx[i]:
--- a/pyaerocom/io/read_ebas.py
+++ b/pyaerocom/io/read_ebas.py
@@ -1378,6 +1378,7 @@ class ReadEbas(ReadUngriddedBase):
                     else:
                         raise
 
+
         if len(data_out["var_info"]) == 0:
             raise EbasFileError(
                 f"All data columns of specified input variables are NaN in {filename}"
@@ -1389,6 +1390,7 @@ class ReadEbas(ReadUngriddedBase):
 
         if self.readopts_default.ensure_correct_freq:
             self._flag_incorrect_frequencies(data_out)
+
 
         # compute additional variables (if applicable)
         data_out = self.compute_additional_vars(data_out, vars_to_compute)
@@ -1690,6 +1692,18 @@ class ReadEbas(ReadUngriddedBase):
         data = self._read_files(files, vars_to_retrieve, files_contain, constraints)
 
         data.clear_meta_no_data()
+
+        # # remove negative values. EBAS species appear to be positive valued only
+        # neg = np.where(data._data[:, data._DATAINDEX ] < 0)
+        # neg_bool = (data._data[:, data._DATAINDEX ] < 0)
+        # # this loop is to remove the negative rows from the metadata
+        # # data.meta_idx is indected by 0.0, 1.0, ... 
+        # for i in data.meta_idx:
+        #     # typically only one, but incase there are more lop through
+        #     for species in data.meta_idx[i]:
+        #         data.meta_idx[i][species] = np.intersect1d(data.meta_idx[i][species], neg)
+        # # the data, unlike the metadata, is a flat file, so can remove all at once
+        # data._data = data._data[~neg_bool,:]
 
         return data
 
--- a/pyaerocom/ungriddeddata.py
+++ b/pyaerocom/ungriddeddata.py
@@ -2672,7 +2672,7 @@ class UngriddedData:
         insert_nans=True,
         ax=None,
         **kwargs,
-    ):
+    ):  # pragma: no cover
         """Plot time series of station and variable
 
         Parameters
--- a/pyaerocom/ungriddeddata.py
+++ b/pyaerocom/ungriddeddata.py
@@ -370,20 +370,21 @@ class UngriddedData:
 
         return data_obj
 
-    def add_station_data(self, stat, meta_idx=None, data_idx=None, check_index=False):
+    def add_station_data(
+        self, stat, meta_idx=None, data_idx=None, check_index=False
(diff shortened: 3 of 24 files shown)


lewisblake:
Note to self: possibly include #605

lewisblake:
Address #708

lewisblake:
Think this is at a good place to merge for now. Can always return to this effort, but I think we've hit a milestone with over 80% coverage

pyaerocom/io/helpers.py
```diff
@@ -78,7 +78,7 @@ def aerocom_savename(data_id, var_name, vert_code, year, ts_type):
     return f"aerocom3_{data_id}_{var_name}_{vert_code}_{year}_{ts_type}.nc"
 
 
-def _print_read_info(i, mod, tot_num, last_t, name, logger):
+def _print_read_info(i, mod, tot_num, last_t, name, logger):  # pragma : no cover
```
avaldebe, line 81: should be `pragma: no cover` without the extra spece before the `:`

pyaerocom/io/read_ebas.py
```diff
@@ -1691,6 +1691,18 @@ def read(
...
+        # neg = np.where(data._data[:, data._DATAINDEX ] < 0)
+        # neg_bool = (data._data[:, data._DATAINDEX ] < 0)
+        # # this loop is to remove the negative rows from the metadata
+        # # data.meta_idx is indected by 0.0, 1.0, ...
+        # for i in data.meta_idx:
+        #     # typically only one, but incase there are more lop through
+        #     for species in data.meta_idx[i]:
+        #         data.meta_idx[i][species] = np.intersect1d(data.meta_idx[i][species], neg)
+        # # the data, unlike the metadata, is a flat file, so can remove all at once
+        # data._data = data._data[~neg_bool,:]
```
avaldebe, line 1704: please remove dead code

pyaerocom/ungriddeddata.py
```diff
@@ -372,17 +372,18 @@ def from_station_data(stats, add_meta_keys=None):
...
+        #     meta_idx = self.last_meta_idx + 1
+        # elif meta_idx in self.meta_idx:
+        #     raise ValueError(
+        #         f"Cannot add data at meta block index {meta_idx}, index already exists"
+        #     )
+
+        # if data_idx is None:
+        #     data_idx = self._data.shape[0]
+        # elif not np.all(np.isnan(self._data[data_idx, :])):
+        #     raise ValueError(f"Cannot add data at data index {data_idx}, index already exists")
```
avaldebe, line 386: please remove dead code

pyaerocom/varnameinfo.py
```diff
@@ -134,8 +134,9 @@ def __str__(self):
         s = (
             f"\nVariable {self.var_name}\n"
             f"is_wavelength_dependent: {self.is_wavelength_dependent}\n"
-            f"is_optical_density: {self.is_optical_density}"
         )
+        if hasattr(self, "is_optical_density"):
+            s += f"is_optical_density: {self.is_optical_density}\n"  # pragma: no cover # can't find situation where this happens however not sure if depricated
```
avaldebe, line 139: the `pragma: no cover` should be on the `if` line
``` python
        if hasattr(self, "is_optical_density"): # pragma: no cover 
            s += f"is_optical_density: {self.is_optical_density}\n"  # can't find situation where this happens however not sure if deprecated
```
that way it is clear that the whole if branch wont be covered

tests/aeroval/test_coldatatojson_helpers2.py
```diff
@@ -1,23 +1,36 @@
 # ToDo: merge with test_coldatatojson_helpers.py
 from __future__ import annotations
 
+from importlib.metadata import metadata
```
avaldebe, line 4: I can not see where this import is used

tests/test_trends_helpers.py
```diff
@@ -0,0 +1,122 @@
...
+)
+def test__end_season(season, year):
+    try:
+        res = _end_season(season, year)
+        assert isinstance(res, str)
+        assert "01" in res
+    except:
+        with pytest.raises(ValueError) as e:
+            res = _end_season(season, year)
+        assert e.type is ValueError
```
avaldebe, line 95: The way this test is structured it is not clear what cases will raise an exception and which test cases wil produce a valid result.

could you put the tests-cases that raise an exception in a separate test?

tests/test_trends_helpers.py
```diff
@@ -0,0 +1,122 @@
...
+)
+def test__mid_season(season, year):
+    try:
+        res = _mid_season(season, year)
+        assert isinstance(res, np.datetime64)
+        assert "15" in str(res)
+    except:
+        with pytest.raises(ValueError) as e:
+            res = _mid_season(season, year)
+        assert e.type is ValueError
```
avaldebe, line 51: no clear which tests will create a valid result and winch ones will raise an exception.

Please slit into 2 test functions, one for valid test cases and another for the test cases that will raise an exception

tests/test_ungriddeddata.py
```diff
@@ -1,11 +1,12 @@
 import string
+from asyncio.proactor_events import _ProactorBaseWritePipeTransport
```
avaldebe, line 2: not sure if this import is ever used

tests/test_ungriddeddata.py
```diff
@@ -239,3 +240,56 @@ def test_from_single_station_data():
...
+    assert isinstance(aeronetsunv3lev2_subset.has_flag_data, (np.bool_, bool))
+
+
+def test_is_filtered(aeronetsunv3lev2_subset: UngriddedData):
+    assert isinstance(aeronetsunv3lev2_subset.is_filtered, (np.bool_, bool))
+
+
+def test_available_meta_keys(aeronetsunv3lev2_subset: UngriddedData):
+    assert isinstance(aeronetsunv3lev2_subset.available_meta_keys, list)
+    assert np.all(isinstance(key, str) for key in aeronetsunv3lev2_subset.available_meta_keys)
```
avaldebe, line 259: Why `np.all`? A plain `all`should be enough

tests/test_ungriddeddata.py
```diff
@@ -239,3 +240,56 @@ def test_from_single_station_data():
...
+
+def test_nonunique_station_names(aeronetsunv3lev2_subset: UngriddedData):
+    assert isinstance(aeronetsunv3lev2_subset.nonunique_station_names, list)
+
+
+def test_set_flags_nan_error(aeronetsunv3lev2_subset: UngriddedData):
+    data = aeronetsunv3lev2_subset.copy()
+    with pytest.raises(AttributeError) as e:
+        data = data.data.set_flags_nan(inplace=True)
+    assert e.type is AttributeError
```
avaldebe, line 270: This line is redundant. The `with pytest.raises(AttributeError)` will fail if a different exception is raised.
Instead, you could assert on the exception message. e.g.
``` python
    assert str(e.value).startswith("Ups....")
```

tests/test_ungriddeddata.py
```diff
@@ -239,3 +240,56 @@ def test_from_single_station_data():
...
+
+def test_extract_var(aeronetsunv3lev2_subset: UngriddedData):
+    data = aeronetsunv3lev2_subset.copy()
+    od = data.extract_var("od550aer")
+    assert not data.is_filtered
+    assert od.is_filtered
+    assert od.shape[0] < data.shape[0]
+    with pytest.raises(VariableDefinitionError) as e:
+        data.extract_var("nope")
+    assert e.type is VariableDefinitionError
```
avaldebe, line 288: this line is redundant, see previous comment

tests/test_ungriddeddata.py
```diff
@@ -239,3 +240,56 @@ def test_from_single_station_data():
...
+
+
+def test_find_common_stations(aeronetsunv3lev2_subset: UngriddedData):
+    data1 = aeronetsunv3lev2_subset.copy()
+    data2 = aeronetsunv3lev2_subset.copy()
+    station_map = data1.find_common_stations(other=data2)
+    assert np.all(key == station_map[key] for key in station_map)
```
avaldebe, line 295: why `np.all` instead of plain `all`?
maybe it could be written as:
``` python
    assert station_map == {key: key for key in station_map)
```
or 
``` python
    assert station_map.keys() == station_map.values()
```

tests/test_varnameinfo.py
```diff
@@ -0,0 +1,87 @@
...
+        ("od550gt1aer", True, True, True),
+    ],
+)
+def test_varnameinfo(var_name, contains_numbers, contains_wavelength_nm, is_wavelength_dependent):
+    var_name_info = VarNameInfo(var_name)
+    assert isinstance(var_name_info, VarNameInfo)
+    assert var_name_info.contains_numbers == contains_numbers
+    assert var_name_info.contains_wavelength_nm == contains_wavelength_nm
+    assert var_name_info.is_wavelength_dependent == is_wavelength_dependent
+    assert isinstance(var_name_info.__str__(), str)
```
avaldebe, line 21: `__str__()` is called when converting a class into a `str`, so maybe this would be better 
``` python
    assert isinstance(str(var_name_info), str)
```
or 
``` python
    assert hasattr(var_name_info, "__str__")
```
but I'm not sure....

tests/test_varnameinfo.py
```diff
@@ -0,0 +1,87 @@
...
+    [("concpm10", None), ("wetoxn", "Surface"), ("od550aer", "Column")],
+)
+def test_get_default_vert_code(var_name, vert_code):
+    var_name_info = VarNameInfo(var_name)
+    try:
+        assert var_name_info.get_default_vert_code() == vert_code
+    except:
+        with pytest.raises(ValueError) as e:
+            var_name_info.get_default_vert_code()
+        assert e.type is ValueError
```
avaldebe, line 35: please put the test cases that will raise an exception into a separate test function, see previous comments

tests/test_varnameinfo.py
```diff
@@ -0,0 +1,87 @@
...
+    [
+        ("concpm10"),
+        ("wetoxs"),
+    ],
+)
+def test_wavelength_nm_errors(var_name):
+    var_name_info = VarNameInfo(var_name)
+    with pytest.raises(VariableDefinitionError) as e:
+        var_name_info.wavelength_nm()
+    assert e.type is VariableDefinitionError
```
avaldebe, line 74: this line is redundant, see previous....

avaldebe (changes requested):
Very good.
Please remove dead code and split valid/invalid test cases.

pyaerocom/ungriddeddata.py
lewisblake: I specifically wrote "Do not remove below", because at this point, I do not know if this can be removed from the codebase. This section originlly appeared after `raises(NotImplementedError("Coming at some point"))`, but was counting against our code coverage. Until it is functional in the code, I do not believe it should count against our coverage. I can always leave it as I found it.

avaldebe: sorry, I did not see the `Do not remove` part.
Can it be moved to a separate function that raises NotImplementedError as the 1st line?
That way the whole function can me marked with `pragma: no cover`

tests/test_varnameinfo.py
lewisblake: I want to actually execute the `__str__()` function though. I don't think these alternatives will do that

pyaerocom/ungriddeddata.py
lewisblake: That essentially what it originally was, except without the `pragma: no cover`. I added one and uncommented out the code again. That seems cleaner

tests/test_varnameinfo.py
avaldebe: `str(object)` will call `object.__str__()`, but it is up to you

lewisblake: Okay, first option it is then :)

tests/test_varnameinfo.py
```diff
@@ -23,16 +23,21 @@ def test_varnameinfo(var_name, contains_numbers, contains_wavelength_nm, is_wave
...
+
+
+@pytest.mark.parametrize(
+    "var_name",
+    [("concpm10")],
+)
+def test_get_default_vert_code_error(var_name):
+    var_name_info = VarNameInfo(var_name)
+    with pytest.raises(ValueError):
+        var_name_info.get_default_vert_code()
```
avaldebe, line 40: if there is only one test case, maybe you could "skip" the test parameterization and write
``` python
def test_get_default_vert_code_error():
    var_name_info = VarNameInfo("concpm10")
    with pytest.raises(ValueError):
        var_name_info.get_default_vert_code()
```

avaldebe: also, I'm not sure if the test is meant to provide a single string or a tuple of strings.

avaldebe (changes requested):
only one small nitpick regarding one of the tests

avaldebe (approved):
🚀 🎆 ready to merge

PR #702, opened 2022-07-13
set log filename from environment variable

### LINKED ISSUE #687: logging needs to use a more specific file name
The future paralleisation will start several aeroval jobs from the same work directory. The current logging to the file `./pyaerocom.log` ends up in a scambled log file.
Because my qsub scripts log the console output to a specific file like `/lustre/storeA/project/aerocom/logs/aeroval_logs/jang.20220621_101003.cfg_file_IASI.ULB_AeronetSDAV3L2.1657189_log.txt` the general log file in in principle not needed at all (if the console output works). So we might want to get rid of that somehow with an option.

For discussion.

avaldebe:
close #687

### DIFF
--- a/pyaerocom/_logging.py
+++ b/pyaerocom/_logging.py
@@ -18,7 +18,7 @@ from logging.config import fileConfig
 
 LOGGING_CONFIG = dict(
     # root logger
-    file_name=os.getenv("PYAEROCOM_LOG_FILE", "pyaerocom.log"),
+    file_name=os.getenv("PYAEROCOM_LOG_FILE", default="pyaerocom.log"),
     file_days="14",
     file_level="DEBUG",
     # pyaerocom logger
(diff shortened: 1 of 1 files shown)


pyaerocom/_logging.py
```diff
@@ -12,12 +12,13 @@
...
 
 import logging
+import os
 from importlib import resources
 from logging.config import fileConfig
 
 LOGGING_CONFIG = dict(
     # root logger
-    file_name="pyaerocom.log",
+    file_name=os.getenv("PYAEROCOM_LOG_FILE", "pyaerocom.log"),
```
jgriesfeller, line 21: personally I would prefer to set the entire log file name via the env variable. Just like many command line tools support.

avaldebe: you can provide full path on `PYAEROCOM_LOG_FILE`

jgriesfeller: ups, I just realised that I did not understand the `os.getenv` method. On the other hand the Python documentation mentions this syntax (which I find clearer): `os.getenv(key, default=None)`
Would you mind adding the default keyword?

avaldebe: done

PR #703, opened 2022-07-13
scaffolding for plugin architecture 

avaldebe:
Provide some readers via entry_points


pyaerocom/io/readungridded.py
```diff
@@ -52,14 +56,15 @@ class ReadUngridded:
         ReadAeronetSunV3,
         ReadEarlinet,
         ReadEbas,
-        ReadGAW,
         ReadAasEtal,
-        ReadGhost,
         ReadAirNow,
```
lewisblake, line 60: It's not clear to me why only GAW, Ghost, and MSCW CTM have their own plugins. Is there something special about these or do we just need to begin moving the other readers into this format later?

pyproject.toml
```diff
@@ -1,6 +1,99 @@
...
+    "Development Status :: 3 - Alpha",
+    "Intended Audience :: Science/Research",
+    "Intended Audience :: Education",
+    "Topic :: Scientific/Engineering :: Atmospheric Science",
+]
+requires-python = ">=3.8"
+dependencies = [
+    'importlib-metadata>=3.6; python_version < "3.10"',
+    'tomli>=2.0.1; python_version < "3.11"',
+    # scitools-iris>=3.1.0,!=3.2.*
```
lewisblake, line 26: Consider removing these commented out lines. Looks like some are covered by `[project.optional-dependencies]` below

lewisblake (commented):
Little things. I think this architecture makes a lot more sense and helps split things into more manageable pieces. Godt jobbet

pyaerocom/io/readungridded.py
```diff
@@ -52,14 +56,15 @@ class ReadUngridded:
...
         ReadMarcoPolo,
         ReadEEAAQEREP,
         ReadEEAAQEREP_V2,
     ]
+    SUPPORTED_READERS.extend(
+        ep.load() for ep in metadata.entry_points(group="pyaerocom.ungridded")
+    )
```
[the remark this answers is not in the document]
avaldebe: A broken module would raise a `SyntaxError`, which should cause the test suite to fail.

A more subtle fail would be a missing dependency, which would raise a `ModuleNotFoundError`.
We could try/except this exception and issue a `warnings.warn`, but any evaluation relying on the missing reader would crash at a later point.
I think that it is best to let it fail at this point.

pyaerocom/io/readungridded.py
avaldebe: These were easy to port. Other readers are more complicated.
We can port more readers on subsequent PRs.

pyproject.toml
avaldebe: god point, will do

pyaerocom/plugins/ghost/additional_variables.py
```diff
@@ -0,0 +1,54 @@
...
+    data : list
+        list of modified :class:`StationData` objects that include computed
+        mass concentrations in addition to VMR data.
+
+    """
+    for stat in data:
+        vmrdata = stat[vmrvar]
+        meta = stat["meta"]
+        p = meta["network_provided_volume_standard_pressure"]
+        T = meta["network_provided_volume_standard_temperature"]
```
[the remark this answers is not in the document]
avaldebe: So far, I only moved the code around and renamed some modules.
I will rename this variables following your suggestion

jgriesfeller (approved):
Looks good to me

Issue #713, opened 2022-08-15
Daily data is average out for station with other ts_types


dulte:
If a station has data in two different ts_types, the function Ungriddeddata.to_station_data will resample everything to the coarsest ts_type. Meaning that if we look at daily data, then some stations might simply be removed if there exist data in weekly or monthly resolution. This is the case for two stations in Norway when looking at precipitation.

dulte:
See #714

dulte:
#714 provided a fix. So I'll close this issue for now

PR #714, opened 2022-08-15
Fixes bug where daily data is avg out when lower freq is present for …

dulte:
…same station in same time span


dulte:
See #713

dulte:
Lewis is working on a fix for CI. Might be that we need to update the expected value for the colocated data in the CI

jgriesfeller (approved):
good to go

PR #715, opened 2022-08-18
make calulation of temporal correlation work with xarray 2022.6.0

jgriesfeller:
and optimise code by avoiding an unnecessary data copy operation.

Can likely be further optimised by directly using numpy


lewisblake:
Ready to merge :racing_car:

PR #721, opened 2022-09-05
2022 reporting

dulte:


dulte:
Two most important failing test:

- test_read_emep_multiple_dirs: Due to change in reading of multiple years. Time dimension in fake_MSCWCtm_data does not seem to be real years, and therefore fails when xarray tries to concatinate
- test_read_emep_dummy_data: Fails due to change in definition of concNtno3 in reader

lewisblake (approved):
:1st_place_medal:

PR #727, opened 2022-09-07
Splitting ColocatedData with the goal of accounting for vertical profiles

lewisblake:
This is a (mostly failed) attempt at splitting up `ColocatedData` for the eventual inclusion of 3D spatial fields and 4D spatio-temporal fields. Hence **this is not a real draft PR**, but mostly keeping this here for posterity until we decide how to proceed. 

The goal was to make `ColocatedData` a parent class and then create derived classes `ColocatedData2D` and `ColocatedData3D`. Trying to put everything agnostic to the number of spatial dimensions into the parent class `ColocatedData` left very little in `ColcatedData2D` (about 350 lines of code).  

Daniel and I realized that this is not a trivial task, and discussed what we would want out of "an ideal colocated data class". We for the time being agreed that it would be ideal to only have colocation done on the tuple (lon, lat, alt, time), instead "dimensions" such as the station name. This is geared towards accounting for the vertical profile data and is basically the point-cloud idea Jan has brought up before. We want to discuss this in our next meeting, because creating and entirely new colocated data class is a huge overhaul.

### DIFF
--- a/pyaerocom/__init__.py
+++ b/pyaerocom/__init__.py
@@ -45,7 +45,10 @@ from .stationdata import StationData
 from .griddeddata import GriddedData
 from .ungriddeddata import UngriddedData
 from .filter import Filter
-from .colocateddata import ColocatedData
+
+# from .colocateddata import ColocatedData
+from .colocateddata2d import ColocatedData2D
+from .colocateddata3d import ColocatedData3D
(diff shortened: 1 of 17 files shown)


lewisblake:
Note to self: `PointCloudData` class

PR #737, opened 2022-09-16
add StationCoordinateError to exceptions to skip over during obs data reading

### LINKED ISSUE #652: Obs data: skip stations that raise StationCoordinateError
At the moment the following exceptions raised during obs data reading do **not** make the reading fail:
https://github.com/metno/pyaerocom/blob/0df5099865a6a0ce684722a237cf92da9f6be621/pyaerocom/ungriddeddata.py#L1233-L1238

I think we should add `StationCoordinateError` to that list. 

Explanation:
`StationCoordinateError` is raised here
https://github.com/metno/pyaerocom/blob/0df5099865a6a0ce684722a237cf92da9f6be621/pyaerocom/stationdata.py#L368-L374 
and here 
https://github.com/metno/pyaerocom/blob/0df5099865a6a0ce684722a237cf92da9f6be621/pyaerocom/io/readaeronetbase.py#L382-L388

In the latter case, the station is skipped already.

The current behavior is problematic, because the user has to remove the station raising this exception by hand to get an experiment through aeroval. 
**Skipping a station should nevertheless logged properly so that it's easily traceable**

To be discussed at the user meeting

jgriesfeller:
fixes #652

### DIFF
--- a/pyaerocom/trends_engine.py
+++ b/pyaerocom/trends_engine.py
@@ -1,6 +1,8 @@
 import matplotlib.pyplot as plt
 import numpy as np
-from matplotlib.cm import get_cmap
+from matplotlib import colormaps
+
+# from matplotlib.cm import get_cmap
 from matplotlib.colors import Normalize
 from scipy.stats import kendalltau
 from scipy.stats.mstats import theilslopes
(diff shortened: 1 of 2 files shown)


avaldebe (commented):
🚀 🎆 ready to merge

Issue #736, opened 2022-09-15
Missing model data is interpolated when colocate_time=True


dulte:
When colocate_time=True one expects the time stamp to be removed if either obs data or model data is missing. When obs data is missing, the model data is removed at that point. But in the reverse case, model data is interpolated instead of the obs data being removed. This means that you will end up with phantom model data

This is most probably done because model data is assumed to be complete

(sorry for the bad explanation, I will elaborate more during the meeting)

jgriesfeller:
In the case of satellite data (as model data) model data is also not complete. So this is a bug that needs to be resolved.

dulte:
#755

PR #755, opened 2022-10-11
First pass on removing nan values in model from obs with colocate time

dulte:
Then colocate_time=True, model values are removed where you have nan obs. The reverse is not the case. This is to fix that. See #736


lewisblake:
@dulte would you like help with this?

Issue #756, opened 2022-10-24
Split map files by year


lewisblake:
Similar to #707 

In the obs-only project, the map files can be very large (upwards of 200MB), so we need to split them by year as well to reduce their size.

PR #758, opened 2022-10-24
Split map files by year

lewisblake:
Closing #756


dulte:
Have you tested this for periods of more than one year? E.g. 2010-2020.

lewisblake:
On the Aeroval side, @AugustinMortier will be testing the results from a run from this afternoon over 2018-2020 using this set up shortly. Perhaps it would be best to wait and see how this change effects the website before merging.

lewisblake:
@AugustinMortier says it works on the Aeroval side

dulte (approved):
Then you can merge :)

PR #760, opened 2022-10-25
let Python 3.11 tests crash on CI

avaldebe:
- builds netCDF4 package on CI, as there is no wheel for Python 3.11
- fails to build cartopy on CI
- CI won't fail if/when tests for Python 3.11 fail


dulte (approved):
Looks good. Let's quickly look at it during the meeting, but if it passes the test, then I think you can merge

PR #761, opened 2022-10-26
obs_only_stats flag in config

lewisblake:
In obs-only experiments, the statistics which make sense only related to the observations (i.e., refdata), and we only want to display these on the web. To avoid having to manually edit `statistics.json` each time remove the meaningless statisics, this proposes adding a flag, `obs_only_stats`, which can be set in the config if the Obs-Mean stats are the only things wanted from an obs-only run.


lewisblake:
Tested on a local run

lewisblake:
@dulte thoughts on this?

dulte (approved):
Looks good

PR #783, opened 2022-11-07
conda-forge compatibility

avaldebe:
- [X] exclude xarray 2022.11.0 from dependencies
seems that xarray 2022.11.0 crashes tests for the conda-forge package creation https://github.com/metno/pyaerocom/discussions/782#discussioncomment-4076584
- [x] address deprecation warnings from `importlib.resources` on Python 3.11 https://github.com/metno/pyaerocom/discussions/782#discussioncomment-4077214
- [x] add Python 3.11 conda env, let test fail without failing CI

### DIFF
--- a/.github/workflows/CI.yml
+++ b/.github/workflows/CI.yml
@@ -116,8 +116,10 @@ jobs:
     - name: Install pyaerocom
       run: python -m pip install . --no-deps
     - name: Run pytest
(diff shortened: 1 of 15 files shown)


lewisblake (approved):
`pyaerocom/data/resources.py`: is this standard approach? Will we need to keep this around so long as we maintain older versions of Python below 3.11?

PR #787, opened 2022-11-15
Update plotting default ranges on Aeroval

lewisblake:
@AugustinMortier and I went through the obs-only experiment and defined better global defaults for the plotting on many variables. This contains those updates.


dulte (approved):
Looks good

PR #796, opened 2022-11-29
run CI for Python 3.11 on Ubuntu 22.04

avaldebe:
There is no Cartopy wheel for Python 3.11. Creating the wheel requires Proj8, which is only available on Ubuntu 22.04.


lewisblake (approved):
:unicorn:

PR #797, opened 2022-11-30
Fix PPI CI

lewisblake:
Fixes for tests which crash CI on PPI. These tests previously did not crash CI on GitHub because there is no lustre there


avaldebe (approved):
ready to merge 🚀 🎆

PR #794, opened 2022-11-28
temperature from ebas

avaldebe:
close #788

Issue #788, opened 2022-11-15
Temperature data


AugustinMortier:
As part of the [observations explorer](https://aeroval.met.no/evaluation.php?project=obs) project, it would be great to have temperature data, in addition to the already supported precipitations. Those two variables could be part of the 'meteorology' experiment.
EBAS provide temperature data over Europe.

dulte:
Surface temp. is already defined as ts in variables.ini, so you only need to add it to the ebas_config. EBAS has a variable called temperature, which is found as "air", "instrument", "met", or "aerosol" or "PM X"

lewisblake:
Currently implemented in main and working in test observation meteorology experiments

PR #798, opened 2022-12-01
Surface temperature cont.

lewisblake:
Updates to include surface temperature on the web. Currently providing the variable a more readable name.

Realted to #788, #794

Need to:
1. Fix the classification of TS as Surface, no Column
2. Figure out why the data is not showing on the web side
Screenshot from 2022-12-01 11-16-01


lewisblake:
@AugustinMortier and I tried to implement use of the temperature variable in an experiment, but had no luck. The data possibly might be invalid for this variable, or the way the reader works can not yet handle this case. Need to figure out if it's ever been used before. If not, why was it partially implemented in pyaerocom?

lewisblake:
Try reading the data using base pyaerocom

lewisblake:
Reading with the base API seems to work

lewisblake:
the coldata nc files seem to have valid coords

lewisblake:
check out `_process_regional_timeseries`

lewisblake:
Issue seems to be going from UngriddedData to ColocatedData for ts

lewisblake:
dummy data for prmm are 1 but for ts they are NaN

lewisblake:
The issue was that in variables.ini, the minimum for ts was 170, and so in GriddedData, the dummy values which are set to all be 1 were being masked out. This resulted in all the dummy data being NaN, which led to inappropraite application of  `merged.loc[merged.data.isnull()] = np.nan` in `_colocate_site_data_helper_timecol`, and the resulting colocated data objects being full of NaNs. Probably need to bring this up at a pyaerocom meeting, since my solution of just changing the accepted minimum value is variables.ini is pretty hacky. A more robust solution is probably required. Maybe for dummy data, we multiple the data (np.ones) by the mean of the min and max values provided in variables.ini

lewisblake:
Working locally, although may nee to change the scale on the y axis of the timeseries

lewisblake:
My final fix was to only do filtering on the data in the case of not doing on observation only experiment (i.e., if the model is a dummy model)

jgriesfeller (approved):
Looks good to me

PR #800, opened 2022-12-05
Aeroval path fixture

avaldebe:
When running the test suite on PPI, from `~\pyaerocom`, the test suite fails with the following error message

```
 tests/aeroval/test_setup.py::test_evalsetup_cams84[IO paths]
...
E           PermissionError: [Errno 13] Permission denied: '../../data'
```

This is because `../../data` defined on the `CAMS84` configuration is resolved as `/home/data`.

This PR adds a fixture that changes the current work directory to a temporary path, where `../../data` can be safely resolved.

### DIFF
--- a/tests/aeroval/test_setup.py
+++ b/tests/aeroval/test_setup.py
@@ -2,7 +2,7 @@ from __future__ import annotations
 
 from typing import Iterable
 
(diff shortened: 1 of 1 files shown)


lewisblake (approved):
Looks good

PR #801, opened 2022-12-05
clean up MEP reader

avaldebe:
follow up from #784


lewisblake (approved):
Ser veldig bra ut!

lewisblake (approved):
Looks even better. Did you notice any performance changes?

avaldebe:
> Looks even better. Did you notice any performance changes?

It is a bit faster, but nothing considerable

avaldebe:
@dulte 
I would appreciate if you can have a look at the `MEP` (formerly `HARP`) reader.
In particular, I would like your opinion on the `dtime` settings.
Currently it uses the `datetime_start` field, but maybe it should use the `datetime_end` or the the average of both (time at the middle of the period).

lewisblake (approved):

PR #805, opened 2022-12-14
Adding MEP data

lewisblake:


lewisblake:
The issue from running a config was that in the __init__ of the MEP reader, Alvaro had (with the best of intentions) a clause to raise a ValueError is the data_dir was None. This cause the Collocator to fail because "pyaeroval" is set up in such a way that it initializes the readers, and then uses the AerocomBrowser to determine the data_dir for the obs_id. So the code as failing before it even got to read the paths I was placing in paths.ini. 


To get around this Add the MEP path to const, import it in the reader. From this point testing possibilities of setting info from const as defaults vs checking is lustre is available and then setting the data_dir

lewisblake:
It looks like cartopy is crashing CI in virtual environments

pyaerocom/plugins/mep/reader.py
```diff
@@ -55,7 +56,7 @@ class ReadMEP(ReadUngriddedBase):
     __version__ = "0.01"
```
jgriesfeller, line 56: please increase version number

jgriesfeller (changes requested):
The way you implemented that reader seems to be a hybrid between the plugin readers from @avaldebe and the readers in the `io` folder.
I have only experience with the latter and therefore propose to discuss the details during the next pyaerocom meeting when we all are back from holidays. At least there's a reader now one can incorporate in a branch to use it.
In addition, @avaldebe would ask: Where are the tests ;-)

pyaerocom/config.py
```diff
@@ -88,6 +88,9 @@ class Config:
     #: DMS
     DMS_AMS_CVO_NAME = "DMS_AMS_CVO"
 
+    #: MEP name
+    MEP_NAME = "MEP"
+
```
avaldebe, line 93: Why is this needed?
You are coupling back the plugin with something configured on the core.
This breaks the core/plugin separation,

pyaerocom/data/paths.ini
```diff
@@ -100,6 +100,7 @@ EEA_V2 = ${BASEDIR}/aerocom/aerocom1/AEROCOM_OBSDATA/EEA_AQeRep.v2/renamed/
 
 AIR_NOW = ${BASEDIR}/aerocom/aerocom1/AEROCOM_OBSDATA/MACC_INSITU_AirNow
 MARCO_POLO = ${BASEDIR}/aerocom/aerocom1/AEROCOM_OBSDATA/CHINA_MP_NRT
+MEP = ${BASEDIR}/aerocom/aerocom1/AEROCOM_OBSDATA/MEP/download/
```
avaldebe, line 103: Does this need to be here?
It would be better if this could be defined on the plugin.

avaldebe (commented):
Is it possible to keep all MEP definitions inside the the plugin?

pyaerocom/config.py
lewisblake: If I don't include it, the code just hangs... maybe we can discuss Monday

jgriesfeller:
I tested the branch for the entire data set. I stopped after ~ 2 hours and a RAM usage of 20GB since I could not see which percentage of the data has been read (due to no logging setting to debug). I could not get a qlogin with more RAM, so I stopped there for now.
Just as comment: I am pretty sure that `xr.open_mfdataset` is a quite slow way of reading the data for 100+ files per station. We could just create one file per station containing the entire time series and make that work without changing the code.
I might just write a script for that...

jgriesfeller:
Script to convert to one file per station is done, the calculation is running (StoreA).

jgriesfeller:
The current implementation is not what we need because it can't be pickled (and therefore not cached):
```
Writing cache file: /home/jang/MyPyaerocom/_cache/jang/MEP_concpm10.pkl
Failed to write cache: TypeError("cannot pickle 'property' object")
Traceback (most recent call last):
  File "/home/jang/anaconda.2018/envs/pya_para_mep/lib/python3.10/site-packages/pyaerocom/io/cachehandler_ungridded.py", line 347, in write
    pickle.dump(data, out_handle, pickle.HIGHEST_PROTOCOL)
TypeError: cannot pickle 'property' object
```

jgriesfeller:
We need this working (**not** necessarily merged) as soon as possible. The next report is due at the end of the month. Shall I take over the work for now?

lewisblake:
If you can, please. I've been told to finish up other projects with deadlines we've had to push back

jgriesfeller:
OK, I hereby take over this PR

lewisblake:
Is there a status update on this PR? Are we waiting to sort out the plugin architecture issue before merging this (assuming checks pass)?

jgriesfeller:
Trying to finish this PR this week. But we should discuss the plugin structure. Do we even need that given that we want to limit pyaerocom to just read one file format for observations.
And which file format should that be (text based (e.g. Parquet) or binary (e.g. netcdf / HARP)).

lewisblake:
Note from meeting: We are reading the new and old version of the station files because the recursive reader finds both. Move old files or improve reader? @avaldebe

lewisblake:
Perhaps the `ExperimentProcessor`, not the `Colocator`, should be initializing readers

jgriesfeller:
The reader does not find old files. I changed the path it searches:
https://github.com/metno/pyaerocom/blob/dcaf8a623e2a3456cf98010534b642c89673e645/pyaerocom/data/paths.ini#L103

It's just capable to also use the old file structure (with the price of taking forever to read the data)

jgriesfeller:
Why the fuck are obs data paths not put into the path list (`const.OBSLOCS_UNGRIDDED`) list when they are not accessible?
https://github.com/metno/pyaerocom/blob/1fc161b4f41d297695aac783b61a406d79ba6d55/pyaerocom/config.py#L665-L693 makes it impossible to use the const module in tests that are not run on CI.
This will fail:
https://github.com/metno/pyaerocom/blob/1fc161b4f41d297695aac783b61a406d79ba6d55/tests/plugins/mep/test_reader.py#L11

jgriesfeller (approved):
All good

PR #808, opened 2023-01-12
changed methods check_write_access and check_dir_access to just rely …

### LINKED ISSUE #759: Random change of cache folder
If the user has set the CACHEDIR manually, then it might change randomly during an evaluation.

If the check_write_access function in _lowlevel_helpers fails, then the CACHEDIR getter will silently change the cachedir. Since CACHEDIR is called each time the cache is read, this has a non-zero to high probability to happen for longer evaluations, e.g. the long evaluation in CAMS2_83

jgriesfeller:
…on the OS to determine accessibility

The earlier implementation tried to solve a problem that is up to the OS to solve

closes #759

### DIFF
--- a/pyaerocom/_lowlevel_helpers.py
+++ b/pyaerocom/_lowlevel_helpers.py
@@ -5,7 +5,6 @@ import abc
 import logging
 import os
 from collections.abc import MutableMapping
-from concurrent.futures import ThreadPoolExecutor
 from pathlib import Path
 
 import numpy as np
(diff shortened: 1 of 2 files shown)


avaldebe (commented):
The code changes are fine, however CI is failing with the following error:
```
ImportError: cannot import name lgeos
```

As far as I can tell the error message comes from SciTools/cartopy#2076, and the solution is to add add a requirement for `shapely<2`.

I'll try this on a separate PR and came back to you after that.

avaldebe:
> The code changes are fine, however CI is failing with the following error:
> 
> ```
> ImportError: cannot import name lgeos
> ```
> 
> As far as I can tell the error message comes from [SciTools/cartopy#2076](https://github.com/SciTools/cartopy/issues/2076), and the solution is to add add a requirement for `shapely<2`.
> 
> I'll try this on a separate PR and came back to you after that.

PR #809 should fix this...

PR #809, opened 2023-01-12
pin shapely version

avaldebe:
CI is failing with the following error:
```
ImportError: cannot import name lgeos
```

As far as I can tell the error message comes from SciTools/cartopy#2076, and the solution is to add a requirement for `shapely<2`.


lewisblake:
Thank you. This was a rather pervasive issue

lewisblake (approved):

PR #816, opened 2023-01-25
make map engine use the defaults for variables colormap setting

charlienegri:
draft. still testing this.

### DIFF
--- a/pyaerocom/aeroval/modelmaps_engine.py
+++ b/pyaerocom/aeroval/modelmaps_engine.py
@@ -16,6 +16,8 @@ from pyaerocom.exceptions import (
 )
 from pyaerocom.helpers import isnumeric
 
+from pyaerocom.aeroval.glob_defaults import var_ranges_defaults
+
 logger = logging.getLogger(__name__)
 
 
@@ -122,7 +124,12 @@ class ModelMapsEngine(ProcessingEngine, DataImporter):
         """
         data = self.read_model_data(model_name, var)
         check_var_ranges_avail(data, var)
-        varinfo = VarinfoWeb(var)
+        
+        if var in var_ranges_defaults.keys():
+            cmapinfo = var_ranges_defaults[var]
+            varinfo = VarinfoWeb(var,cmap=cmapinfo['colmap'],cmap_bins=cmapinfo['scale']) 
+        else:
+            varinfo = VarinfoWeb(var)
 
         data = self._check_dimensions(data)
 
--- a/pyaerocom/aeroval/modelmaps_engine.py
+++ b/pyaerocom/aeroval/modelmaps_engine.py
@@ -150,7 +150,7 @@ class ModelMapsEngine(ProcessingEngine, DataImporter):
         if tst < freq:
             raise TemporalResolutionError(f"need {freq} or higher, got{tst}")
         elif tst > freq:
-            data = data.resample_time(freq)
+            data = data.resample_time(str(freq))
 
         data.check_unit()
         # first calcualate and save geojson with contour levels
--- a/pyaerocom/aeroval/modelmaps_engine.py
+++ b/pyaerocom/aeroval/modelmaps_engine.py
@@ -143,18 +143,13 @@ class ModelMapsEngine(ProcessingEngine, DataImporter):
                 logger.info(f"Skipping processing of {outname}: data already exists.")
                 return []
 
-        #mainfreq = self.cfg.time_cfg.main_freq
-        allfreqs = self.cfg.time_cfg.freqs
+        freq = min(TsType(fq) for fq in self.cfg.time_cfg.freqs)
+        freq = min(freq, self.cfg.time_cfg.main_freq)
         tst = TsType(data.ts_type)
-        allfreqs = [TsType(fq) for fq in freqs]  # convert them into TsType objects
-        allfreqs.sort(
-            reverse=True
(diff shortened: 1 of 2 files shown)


pyaerocom/aeroval/modelmaps_engine.py
```diff
@@ -122,7 +123,12 @@ def _process_map_var(self, model_name, var, reanalyse_existing):
         """
         data = self.read_model_data(model_name, var)
         check_var_ranges_avail(data, var)
-        varinfo = VarinfoWeb(var)
+
+        if var in var_ranges_defaults.keys():
```
avaldebe, line 127: the `keys` method is not needed to check if a particular key is in a dictionary.
You could write something like

``` python
        if var in var_ranges_defaults:
```

avaldebe (changes requested):
Is this all?

pyaerocom/aeroval/modelmaps_engine.py
charlienegri: noted

charlienegri:
> Is this all?

yeah, not a big change this one

avaldebe (approved):

charlienegri:
added tweaking of the frequency, needs another review

pyaerocom/aeroval/modelmaps_engine.py
```diff
@@ -137,8 +143,20 @@ def _process_map_var(self, model_name, var, reanalyse_existing):
                 logger.info(f"Skipping processing of {outname}: data already exists.")
                 return []
 
-        freq = self.cfg.time_cfg.main_freq
+        mainfreq = self.cfg.time_cfg.main_freq
+        allfreqs = self.cfg.time_cfg.freqs
         tst = TsType(data.ts_type)
+        allfreqs = [TsType(fq) for fq in freqs]  # convert them into TsType objects
```
avaldebe, line 149: I can not see where `freqs` is defined. Do you meant `allfreqs`?

charlienegri:
@avaldebe am done I think, if you wanna take a look

pyaerocom/aeroval/modelmaps_engine.py
```diff
@@ -137,8 +143,20 @@ def _process_map_var(self, model_name, var, reanalyse_existing):
...
+        allfreqs.sort(
+            reverse=True
+        )  # sort them from least coarse to coarsest, e.g. [hourly, daily, monthly]
+        coarsestfreq = allfreqs[-1]
+
+        if mainfreq < coarsestfreq:
+            freq = mainfreq
+        else:
+            freq = coarsestfreq
```
avaldebe: It looks to me that  `allfreqs` is only defined to calculate `coarsestfreq`.
If this is the case you could write as follows:
``` python
        coarsestfreq = min(TsType(fq) for fq in self.cfg.time_cfg.freqs)
```

Also, is it possible for `self.cfg.time_cfg.main_freq` not to be included in `self.cfg.time_cfg.freqs`? In this case, this calculation could be rewritten as:
``` python
        freq = min(TsType(fq) for fq in self.cfg.time_cfg.freqs)
        freq = min(freq, self.cfg.time_cfg.main_freq)
```

pyaerocom/aeroval/modelmaps_engine.py
```diff
@@ -137,11 +143,18 @@ def _process_map_var(self, model_name, var, reanalyse_existing):
...
+        allfreqs = [TsType(fq) for fq in freqs]  # convert them into TsType objects
+        allfreqs.sort(
+            reverse=True
+        )  # sort them from least coarse to coarsest, e.g. [hourly, daily, monthly]
+        coarsestfreq = allfreqs[-1]
+
+        if tst < coarsestfreq:
             raise TemporalResolutionError(f"need {freq} or higher, got{tst}")
-        elif tst > freq:
+        elif tst > coarsestfreq:
```
avaldebe, line 157: what happens when `tst == coarsestfreq`?

avaldebe (changes requested):
it looks to me that `coarsestfreq` can be defined as a `min` of a generator expression, skipping the list / sort steps.

pyaerocom/aeroval/modelmaps_engine.py
charlienegri: I do

pyaerocom/aeroval/modelmaps_engine.py
charlienegri: good point, I assumed not, but it's safer maybe to assume it could

pyaerocom/aeroval/modelmaps_engine.py
avaldebe: also what is the difference between `coarsestfreq` and `freq`?
Don't you want to resample as follows?
``` python
            data = data.resample_time(coarsestfreq)
```

charlienegri: there's no need for resampling I guess

charlienegri: I did want to have `data = data.resample_time(coarsestfreq)` in the previous version, typo 
with your edits, that code is not needed anymore

avaldebe (approved):

charlienegri:
this has been tested and is ready to be merged, only the lint is failing for some reason

PR #818, opened 2023-02-03
Implement color scales from Hilde

lewisblake:
Implementation of the color scale changes from Hilde.


jgriesfeller (approved):
I am taking this as it is. I suspect we also want some color scales for fakedryo3 in addition, but I have to get that working first

PR #824, opened 2023-02-15
pick coarsest freq for scatter plot

charlienegri:
this should be consistent with what the map engine does now (after #816 was merged)

### DIFF
--- a/pyaerocom/aeroval/coldatatojson_engine.py
+++ b/pyaerocom/aeroval/coldatatojson_engine.py
@@ -226,7 +226,7 @@ class ColdataToJsonEngine(ProcessingEngine):
             _write_site_data(ts_objs, out_dirs["ts"])
 
             scatter_freq = min(TsType(fq) for fq in self.cfg.time_cfg.freqs)
-            scatter_freq = str(min(scatter_freq, main_freq))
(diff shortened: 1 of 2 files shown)


pyaerocom/aeroval/coldatatojson_engine.py
```diff
@@ -225,6 +225,9 @@ def process_coldata(self, coldata: ColocatedData):
 
             _write_site_data(ts_objs, out_dirs["ts"])
 
+            scatter_freq = min(TsType(fq) for fq in self.cfg.time_cfg.freqs)
+            scatter_freq = str(min(scatter_freq, main_freq))
```
avaldebe, line 229: please keep the `scatter_freq` type to `TsType` here and convert to string on the `_process_map_and_scat` call (if needed).

avaldebe (changes requested):
small nitpick regarding types/typing, otherwise good

pyaerocom/aeroval/coldatatojson_engine.py
charlienegri: 👍

avaldebe (approved):
🦄 🔥 🚒

PR #827, opened 2023-02-17
Use mean of accepted values as data with dummy_model

lewisblake:
While working on the obs-only experiment, I noticed my previous attempt at a fix was not working for surface temperature (#798). I believe this fix is more robust.

The idea is to not apply the outlier range filters in `GriddedData.remove_outliers()` if the model is `dummy_model`. May be relevant to issues @jgriesfeller was experiencing with the fake deposition species.

Proof it works:

Screenshot from 2023-02-17 17-57-40

### DIFF
--- a/pyaerocom/aeroval/helpers.py
+++ b/pyaerocom/aeroval/helpers.py
@@ -175,6 +175,9 @@ def make_dummy_model(obs_list: list, cfg) -> str:
             # Converts cube to GriddedData
             dummy_grid = GriddedData(dummy_cube)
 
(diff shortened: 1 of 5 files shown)


lewisblake:
The approach is actually slightly different, and maybe should be reflected in the title. 

Instead, use the mean of the min and max values allowed for the variable in variables.ini to fill the dummy model data, instead of using ones. This ensures the values will not be filtered out.

jgriesfeller (approved):
Looks good

PR #834, opened 2023-03-14
Adding regular bias and absolute bias

lewisblake:
Closing #823. Born out of CAMS2_82 this PR is to add the bias and absolute bias statistics. I have some questions about the formulas for other stats. Namely the factor of 2 in MNMB and FGE. Looking [here](https://aaqr.org/articles/aaqr-17-05-lrt-0185_suppl.pdf) I don't see them, but I have seen them in other descriptions. Want to make sure everything is consistent.

Eventually need to add tests.


avaldebe (approved):
ready to merge 🚀 🎆

lewisblake:
Do the statistics look right to you? Still not sure where the factor of 2 comes from in some of the others

avaldebe:
> Do the statistics look right to you? Still not sure where the factor of 2 comes from in some of the others

I would try to avoid calling `sum` directly, as it might end up iterating over `difference[mask]`, e.g:

``` python
        mb = difference[mask].sum() / num_points
        mab = np.abs(difference[mask]).sum() / num_points
```

Maybe numpy has some syntactic sugar where `sum(array)` is actually `array.sum()` without the iteration? Other that it looks OK to me.

Maybe you want to divide by `num_points - 1`, in order to respect the degrees of freedom, but I wont make much difference for the number of obs that we usually dealt with.

lewisblake:
Thanks for the input. It looks like `np.sum` will be faster in this case because we're using numpy arrays. I also don't think we need to substract by 1 in the denominator since we're not estimating any parameters (e.g., a mean such as in the sample variance formula). I do think your syntax is cleaner though, so I changed it to that.

PR #835, opened 2023-03-15
Bias and absolute bias cont.

lewisblake:
Continued work of #834 to make sure the new stats are handled properly on aeroval


jgriesfeller (approved):
approved.

PR #838, opened 2023-03-24
Pin xarray to smaller 2023.3.0

jgriesfeller:
to make CI work again


jgriesfeller:
Background:
This was actually introduced in xarray 2023.03.0 and breaks atm only CI for Python 3.9 and 3.10. But since that change is from 22 March, it's just a matter of time until all CI will break.
This PR that pins xarray to smaller 2023.03.0 right now until we have removed the deprecated code from pyaerocom. The latter was unfortunately not straight forward (and I want finish 2 PRs before my Easter holiday)

lewisblake (approved):
Good catch. We will probably need to talk about this in a meeting

PR #844, opened 2023-04-11
Only set data dummy to mean if defined

lewisblake:
Bug fix from #827. If the min and max aren't defined in variables.ini, then the are set tho to `VMIN_DEFAULT` and `VMAX_DEFAULT` in the `Variable` class, which are currently `-inf` and `inf`. Hence, the mean is not defined, the dummy data all set to `NaN`, and erroneously filtered out.


jgriesfeller (approved):
All good

PR #853, opened 2023-04-24
Tweak bias stats

lewisblake:


lewisblake:
This PR applied weights similar to what is done with mnmb, which the calculation is based on. The current values we're getting out don't really make sense, and so I think these weights need to be applied.

AugustinMortier:
The formulas in the information page are manual work. So I can add/edit a formula in this page if you give me the information

lewisblake:
Remove the weights part and just correct the unit. Run a test to see if weights are needed (what they are)

lewisblake:
Gives reasonable results with just the included changes

jgriesfeller (approved):
Looks good

PR #854, opened 2023-04-25
update coverage configuration

avaldebe:


tests/plugins/test_entry_points.py
lewisblake, line 1: Note to self 7 question: Could something similar be done with the MEP plugin to get the entry points?

lewisblake (approved):

PR #866, opened 2023-05-23
adjusted some color bars to match those from the old idl tools

### LINKED ISSUE #865: adjust aeroval color scales
Some of aeroval's map color scales are not nice:
image

or
image

These should be adjusted.

The PR for this will use some the color bars we had in the idl based aerocom-tools

jgriesfeller:
This PR just adjusts some color bars


jgriesfeller:
some colorbars are too large for the map plots, but fine for the contour plots. We might want to split these up?

lewisblake:
Would like some input from @AugustinMortier regarding how to use two different colorbars for maps and mean-mod in Evaluation on Aeroval

PR #871, opened 2023-06-05
fix test for 3.11

charlienegri:
this is to account for new error message in python 3.11 when attribute cannot be set


lewisblake (approved):
:rocket:

PR #873, opened 2023-06-07
Give proxy variables better name and write only_use_in to menu.json

### LINKED ISSUE #872: proxyod550s need better names on AeroVal and maps=only switch
Non-observables, proxyod550*, could probably be named without "proxy" on AeroVal and should only appear in Maps tab as they do not make sense to have elsewhere for evaluation.

lewisblake:
Designed to close #872

### DIFF
--- a/pyaerocom/aeroval/experiment_output.py
+++ b/pyaerocom/aeroval/experiment_output.py
@@ -623,11 +623,18 @@ class ExperimentOutput(ProjectOutput):
             lname = "UNDEFINED"
         # if a variable only comes from the model, set model_only=True in variables.ini. Picked up here and written to menu.json
         model_only = False
-        try: 
+        try:
             model_only = const.VARS[var].model_only
         except AttributeError:
             pass
-        return {"type": tp, "cat": cat, "name": name, "longname": lname, "obs": {}, "model_only": model_only}
+        return {
+            "type": tp,
+            "cat": cat,
+            "name": name,
+            "longname": lname,
+            "obs": {},
+            "model_only": model_only,
+        }
 
     def _check_ovar_mvar_entry(self, mcfg, mod_var, ocfg, obs_var):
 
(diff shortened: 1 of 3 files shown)


AugustinMortier:
This could work, but this implies that when there is a "model_only" thing, the web interface knows in which pages it is supposed to display the variables. If the pages were set explicitly instead, this would remove one decision layer from the web-side, so I suggest having a key like "use_in": ["maps.php"] (or similar). If there is no "use_in", then we do not filter the variable at all.

lewisblake:
I see. I changed it to a flag `only_use_in`. To be honest, I like this solution a little less because it blurs the line between pyaerocom and "pyaeroval" more. _Maybe_ there is another way to do it, but as it's written, we have to go into `pyaerocom/data/variables.ini` and specify that this variable will only be used in maps.php. This mean that things outside of `pyaerocom/aeroval` have to be aware of what is on aeroval. On the pyaerocom side of things though, it's fine to say that a particular variable comes only from a model. In my mind, the pyaeroval side would handle what to do with that information on the website, but there isn't the infrastructure in place to do that now.

AugustinMortier:
I understand it's not ideal, but the two parts already must know about each other until we have an intermediate system linking the backend and the frontend. I have updated the web-interface so it's using this `use_only_in` key. I have updated [the CAMS2-82 IFS menu](https://aeroval.met.no/data/cams2-82/IFS/menu.json) to test it: [https://aeroval.met.no/evaluation.php?project=cams2-82&exp_name=IFS&station=ALL](https://aeroval.met.no/evaluation.php?project=cams2-82&exp_name=IFS&station=ALL)

Because it's not a menu item, this did not clean up the heatmap, so I'm filtering out, for this chart, the variables starting with "proxy": [https://aeroval.met.no/overall.php?project=cams2-82&exp_name=IFS&station=ALL](https://aeroval.met.no/overall.php?project=cams2-82&exp_name=IFS&station=ALL)

PR #876, opened 2023-06-09
Modifications to 873

lewisblake:
- puts maps.php (and other potential `only_use_in` flags from variables.ini) into a list
- gives Proxy NH4 and NO3 better names
- changes all proxy names from AOD to OD broken down by speciation


lewisblake:
Note: Also only returns "only_use_in" if key exists

PR #895, opened 2023-08-14
try to fix airnow files with strange file encodings

jgriesfeller:
Another approach:
determine file encoding beforehand and provide pandas with that


jgriesfeller:
This is till not the final solution because it takes far too much RAM reading the entire dataset since the file reader reads all variables and keeps all that data in RAM. On the other hand the variables are read by aeroval variable by variable so the multi variable reading ability of the current reader is entirely useless.
This needs to be optimised so that only one variable is read at a time.
As it is now for reading reading vmro3 60GB of RAM are not enough.

jgriesfeller:
Just for documentation: Not even 75GB is enough to read `vmro3`

jgriesfeller:
Just as comment: The earlier years will be added as utf-8 encoded files. Producing them is external work.

pyaerocom/io/read_airnow.py
```diff
@@ -445,3 +559,139 @@ def read(self, vars_to_retrieve=None, first_file=None, last_file=None):
...
+                return "utf_16"
+            if line[2:18] == str(codecs.BOM_UTF32_BE).split("'")[1]:
+                return "utf_32"
+            if line[2:18] == str(codecs.BOM_UTF32_LE).split("'")[1]:
+                return "utf_32"
+        return ""
+
+    def get_all_file_encodings(self, filename):
+        encoding_list = []
+        encodings = (
```
lewisblake, line 582: Maybe we could put this in a different file and read it in?

pyaerocom/io/read_airnow.py
```diff
@@ -445,3 +559,139 @@ def read(self, vars_to_retrieve=None, first_file=None, last_file=None):
...
+                return "utf_8"
+            if line[2:10] == str(codecs.BOM_UTF16_BE).split("'")[1]:
+                return "utf_16"
+            if line[2:10] == str(codecs.BOM_UTF16_LE).split("'")[1]:
+                return "utf_16"
+            if line[2:18] == str(codecs.BOM_UTF32_BE).split("'")[1]:
+                return "utf_32"
+            if line[2:18] == str(codecs.BOM_UTF32_LE).split("'")[1]:
+                return "utf_32"
+        return ""
```
lewisblake, line 578: Does returning `""` yeild the desired behavior for functions which call this? Should a warning be thrown if no encoding can be determined?

pyaerocom/io/read_airnow.py
```diff
@@ -395,15 +442,82 @@ def _filedata_to_statlist(self, arrs, vars_to_retrieve):
                 stats.append(stat)
         return stats
 
-    def read_file(self):
+    def read_file(self, filename, vars_to_retrieve=None):
```
lewisblake, line 445: Could use type hints

pyaerocom/io/read_airnow.py
```diff
@@ -308,29 +330,36 @@ def _read_files(self, files, vars_to_retrieve):
...
+                        raise DataRetrievalError(
+                            f"file {fp}: error in creating unique stationlist"
+                        )
         if len(arrs) == 0:
             raise DataRetrievalError("None of the input variables could be found in input list")
-        return self._filedata_to_statlist(arrs, vars_to_retrieve)
+        return self._filedata_to_statlist(arrs, vars_to_retrieve, unique_stat_ids=unique_stat_ids)
 
-    def _filedata_to_statlist(self, arrs, vars_to_retrieve):
+    def _filedata_to_statlist(self, arrs, vars_to_retrieve, unique_stat_ids=None):
```
lewisblake, line 362: consider adding type hints

pyaerocom/io/read_airnow.py
```diff
@@ -348,9 +377,12 @@ def _filedata_to_statlist(self, arrs, vars_to_retrieve):
             list of :class:`StationData` objects, one for each var and station.
 
         """
+        # doubling of RAM usage!
```
lewisblake, line 380: Worth reconsidering?

lewisblake (changes requested):

pyaerocom/io/read_airnow.py
jgriesfeller: There are more tests after these (in `get_file_encoding`) to we must not raise an exception here

pyaerocom/io/read_airnow.py
jgriesfeller: The whole reader is without type hints. I don't have much time to burn for this right now...

pyaerocom/io/read_airnow.py
jgriesfeller: It's really only used here. I don't see the point really.

pyaerocom/io/read_airnow.py
jgriesfeller: not now, would implicate a major rewrite

jgriesfeller: Just to clarify: Polishing should be done when we convert this reader to the new plugin structure

pyaerocom/io/read_airnow.py
jgriesfeller: added

lewisblake (approved):
:jack_o_lantern:

PR #908, opened 2023-09-25
Drop Proj legacy

avaldebe:
only provide packages compatible with Proj8


tests/io/test_read_aeronet_invv3.py
lewisblake, line 1: Not sure if strictly in the scope of this PR or if these changes are just needed to get it through CI. I modified this test in my Earlinet PR as well. Fine if done here though

lewisblake (commented):
Most looks good. Thanks for the clean up. One question regarding the changes to `test_load_berlin()`

tests/io/test_read_aeronet_invv3.py
lewisblake: Separate PR not need imo

lewisblake (approved):
:rocket:

PR #918, opened 2023-10-05
Fix Regions 901

### LINKED ISSUE #901: Too many regions assigned to stations
In CAMS2_82 we noticed that too many regions are assigned to stations. It can be that a station is assigned one region and oceans, but not a third. 

https://cams2-82.aeroval.met.no/evaluation.php?project=cams2-82&exp_name=IFS&station=SE+Asia&mod=IFS-Reanalysis#

Ctrl+F e.g., N Africa in the station list. Also check the overall evaluation, for a region, number of stations. Then click on one of them and look at the linear model for pred vs obs. More stations appear in this linear model than are reported in the evaluation table.

lewisblake:
Designed to close #901 

Still not currently a perfect solution because regions are sorted according to their distance to the center of that region. A radius determination of containment can be problematic. Some stations which I believe should be part of, for example, North America are only assigned to ALL. Not sure how to get around this without coming up with a new way of assigning regions.

### DIFF
--- a/pyaerocom/aeroval/coldatatojson_helpers.py
+++ b/pyaerocom/aeroval/coldatatojson_helpers.py
@@ -632,10 +632,11 @@ def _init_site_coord_arrays(data):
     return (sites, lats, lons, alts, countries, jsdates)
 
 
-def _get_stat_regions(lats, lons, regions):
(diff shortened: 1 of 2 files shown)


jgriesfeller (approved):
all good

PR #921, opened 2023-10-09
Fix plugins for ICOS and mmr vs vmr issue

### LINKED ISSUE #920: Plugins not picking up ReadICOS
Plugins must be registered in pyproject.toml and then the package reinstalled

lewisblake:
Close #920 and set up tools for running an aeroval experiment. The model data will need to be named in terms of vmr. In a config file, one will need to specify:
-  `io_aux_file` (typically `=os.path.abspath("../eval_py/gridded_io_aux.py")`)
- MODEL_AUX_VARS = {
    "vmrco2": dict(vars_required=["mmrco2"], fun="mmr_to_vmr"),
}
- `model_read_aux=MODEL_AUX_VARS` in the `MODELS` dict


jgriesfeller (approved):
looking good

Issue #926, opened 2023-10-13
add test for working aeroval diurnal cycle analysis to CI


jgriesfeller:
the diurnal cycle analysis has just been accidentially removed from main-dev during a bigger than ususal PR. This slipped throuigh testing because there is no test for it.
Because it's a rarely used feature but also a feature that is important when it's needed, we should have a test if it's still working in CI. 
Needs hourly model and obs data (example `vmrno2` from `EBAS` and EMEP data from here: 
`/lustre/storeB/project/fou/kl/CAMS2_40/task4041/EMEP.cams2.40.p2/`

The model data needs to be much coarser to be small enough for CI

aeroval option to set: `use_diurnal=True`

lewisblake:
One idea is that if it is too big in memory due to hourly data would be to test that this code exists. We could ping it, and mark the test as expecting to fail for no data (or something else)

lewisblake:
General comment: It could be worth thinking about having (at least) two types of tests: unit tests, and and tests which take more time and should be run locally (not on CI) at the frequency closer to the release schedule, not necessarily during PRs. Would be useful for this at PM ratio tests.

PR #927, opened 2023-10-13
Refactor and reintroduce diurnal processing

lewisblake:
In the chat @jgriesfeller mentioned "the diurnal cycle analysis does not seem to work anymore (no files in ts/diurnal although it was switched on and hourly data was there as well... ) and also no mentioning of the word diurnal in the log". Some how this code got dropped during refactoring so I am reintroducing it, with some modification to align with the new styling


jgriesfeller:
issue #926

jgriesfeller (approved):
All good, but why is linting failing?

PR #934, opened 2023-10-24
ReadICOS: CH4 and CO

### LINKED ISSUE #930: Change ICOS station names from station codes and implement CH4
In the ICOS CO2 data (and probably CH4), we need to get the site_name attribute from the netcdf files instead of the site code from the filename. Slightly different behavior than the MEP reader, so likely need to overwrite that method in this derived class.

lewisblake:
Close #930 

Stations codes were fixed on the data preprocessing to create station files. There is still one problematic station which we remove - should bring this up with the data providers


jgriesfeller (approved):
all good

PR #957, opened 2023-12-12
restore default colormaps

### LINKED ISSUE #953: Restore default colormaps
The [new default colormaps](pyaerocom/aeroval/glob_defaults.py) include too many steps in the colorscales which cause several display problems. The maximum number of levels should be around 10 in order for the legend to be properly displayed, and for the maps to not be too heavy with too many polygons. This should be high priority since this affects all projects.

lewisblake:
close #953


AugustinMortier (approved):
Looks good. The absolute numbers for od550dust look a little bit small to me, but I guess it depends on the first projects that were looking at it (which region..). We could adjust it later if needed.

PR #993, opened 2024-02-12
Make computed statistics configurable by allowing dropping of default statistics and specification of the number of decimals shown 

### LINKED ISSUE #885: Make computed statistics configurable
The lately introduced Mean Bias and Mean Absolute Bias seem to be set as default statistics since they show up in freshly processed experiments (like the ones in CAMS2_83). I understand some project requires it, but it's not the case for most of them. I would suggest to remove those statistics from the list of the default used statistics. This would make the menus less overwhelming for most of the users, and mostly, the files smaller by default.

lewisblake:
Designed to close #885 

Similar to what we do for obs-only or model-only experiments, `statistics.json` will be configurable to only show statistics in [`statistics_default`](https://github.com/metno/pyaerocom/blob/f8596b2cca36bf53226d1f224c95a01564888b2b/pyaerocom/aeroval/glob_defaults.py#L329) which are **not** included in the `drop_stats` tuple in the configuration file. This means that if a project does not require Mean Bias or Mean Absolute Bias, for example, then setting `drop_stats = ("mb", "mab")` will result in an experiment which does not show these statistics. The strings provided in `drop_stats` must match the keys in `statistics_default`. `drop_stats` is implemented as a tuple of strings to imply to programmers that this object is immutable at runtime.

### DIFF
--- a/pyaerocom/aeroval/setupclasses.py
+++ b/pyaerocom/aeroval/setupclasses.py
@@ -136,7 +136,12 @@ class StatisticsSetup(ConstrainedContainer):
         days, obviously).
     drop_stats: tuple, optional
         tuple of strings with names of statistics (as determined by keys in
-        aeroval.glob_defaults.py's statistics_defaults) to not compute
+        aeroval.glob_defaults.py's statistics_defaults) to not compute. For example,
+        setting drop_stats = ("mb", "mab"), results in json files in hm/ts with
+        entries which do not contain the mean bias and mean absolute bias,
+        but the other statistics are preserved.
+
+
 
     Parameters
     ----------
--- a/pyaerocom/aeroval/setupclasses.py
+++ b/pyaerocom/aeroval/setupclasses.py
@@ -157,6 +157,7 @@ class StatisticsSetup(ConstrainedContainer):
         self.obs_only_stats = False
         self.model_only_stats = False
         self.drop_stats = ()
+        self.stats_decimals = None
         self.update(**kwargs)
 
 
@@ -237,6 +238,7 @@ class EvalRunOptions(ConstrainedContainer):
         self.only_model_maps = False
         self.obs_only = False
         self.drop_stats = ()
+        self.stats_decimals = None
         self.update(**kwargs)
 
 
--- a/pyaerocom/aeroval/setupclasses.py
+++ b/pyaerocom/aeroval/setupclasses.py
@@ -156,6 +156,7 @@ class StatisticsSetup(ConstrainedContainer):
         self.use_diurnal = False
         self.obs_only_stats = False
         self.model_only_stats = False
+        self.drop_stats = ()
         self.update(**kwargs)
 
 
@@ -235,6 +236,7 @@ class EvalRunOptions(ConstrainedContainer):
         #: If True, process only maps (skip obs evaluation)
         self.only_model_maps = False
         self.obs_only = False
+        self.drop_stats = ()
         self.update(**kwargs)
 
 
--- a/pyaerocom/aeroval/setupclasses.py
+++ b/pyaerocom/aeroval/setupclasses.py
@@ -135,7 +135,7 @@ class StatisticsSetup(ConstrainedContainer):
         compute the statistics for a given month (in this case, a month with 31
         days, obviously).
     drop_stats: tuple, optional
-        tuple of strings with names of statistics (as determined by keys in 
+        tuple of strings with names of statistics (as determined by keys in
         aeroval.glob_defaults.py's statistics_defaults) to not compute
 
     Parameters
--- a/pyaerocom/aeroval/setupclasses.py
+++ b/pyaerocom/aeroval/setupclasses.py
@@ -134,7 +134,9 @@ class StatisticsSetup(ConstrainedContainer):
         the above example 310 values would be used - 31 for each site - to
         compute the statistics for a given month (in this case, a month with 31
         days, obviously).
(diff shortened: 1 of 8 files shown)


lewisblake:
This PR introduces two features that are set in a pyaeroval configuration file:

1.  `drop_stats`: this is a tuple which acts as described above with respect to `statistics.json` but **also** drops the provided statistics from the computed json files. This should hopefully make them a little lighter in some cases. For example, setting `drop_stats = ("mb", "mab"),` results in json files in `hm/ts` with entries which do not contain the mean bias and mean absolute bias, but the other statistics are preserved (forgive the sub-optimal example).
```json
"1389744000000": {
    "totnum": 423.0,
    "num_valid": 0.0,
    "refdata_mean": null,
    "refdata_std": null,
    "data_mean": null,
    "data_std": null,
    "weighted": 0.0,
    "rms": null,
    "nmb": null,
    "mnmb": null,
    "fge": null,
    "R": null,
    "R_spearman": null,
    "num_coords_tot": 423.0,
    "num_coords_with_data": 0.0
}
```
2.  `stats_decimals`: This is an integer which, if provided, overwrites the `"decimals"` key for each statistic in `statistics_default` with the provided value. Some experiments may want to show more decimals than others.

AugustinMortier:
A general thought on the "drop_stats" idea, which, I think is similar to the "hide_pages" issue for AeroVal and from which I'm trying to get away from. Having this requires to have a predefined list of default statistics (/pages) , and excluding some statistics (/pages) out of it. For AeroVal, I suggested to replace the "hide_pages" variable by a "pages" variable which would simply define the list of pages to be shown. This means that the user does not need to know what the default list is.

Why wouldn't you have similarly a "stats" variable, instead of the "drop_stats"? This would probably make a little bit easier to follow what is computed, and for the user to be sure which statistics will actually be computed.

lewisblake:
> Having this requires to have a predefined list of default statistics

Does Aeroval have an internal representation of the statistics computed? I was under the impression Aeroval just picks up what is available in `statistics.json`. We already have a predefined list of statistics in `pyaerocom/aeroval` through [`statistics_default`](https://github.com/metno/pyaerocom/blob/f8596b2cca36bf53226d1f224c95a01564888b2b/pyaerocom/aeroval/glob_defaults.py#L329). In this sense I think it's a little different than `hide_pages`, because `drop_stats` makes no assumption of what is on the web.

> Why wouldn't you have similarly a "stats" variable, instead of the "drop_stats"? This would probably make a little bit easier to follow what is computed, and for the user to be sure which statistics will actually be computed.

One could define a `stats` variable, but I would argue this doesn't have any clear advantages. The modeler still needs to know what statistics are available in `statistics_default` (documentation could be improved here). We don't want such a variable showing up in all of our configuration files either, so making the default behavior to include all the stats in `statistics_default` and only drop those the modeler doesn't care about reduces the potential for misuse without making the code overly complicated. The simplest implementation of a `stats` variable would basically be the same, except it would just remove all computed statistics which aren't provided in `stats`. (The actual compute time of the statistics is quite fast, and so it's not really worth overly complicating the code to only computed the statistics which would be provided in a hypothetical `stats` variable.)

lewisblake:
Long-term goal based on our conversation 20.02.24.

Statistics could be defined in a config, with a (key, function) pair.  The function

https://github.com/metno/pyaerocom/blob/8a1a4a16466e82ead958194764e493c0422adef7/pyaerocom/mathutils.py#L192

could be broken up such that each statistic is computed using its own function. Custom new stats could be implmented with user-provided functions following a defined interface. `statistics.json` would then be based on what actually gets used based on this config file.

AugustinMortier (approved):

pyaerocom/aeroval/setupclasses.py
```diff
@@ -134,7 +134,9 @@ class StatisticsSetup(ConstrainedContainer):
         the above example 310 values would be used - 31 for each site - to
         compute the statistics for a given month (in this case, a month with 31
         days, obviously).
-
+    drop_stats: tuple, optional
+        tuple of strings with names of statistics (as determined by keys in
+        aeroval.glob_defaults.py's statistics_defaults) to not compute
```
heikoklein: - example for drop_stats as in PR comment https://github.com/metno/pyaerocom/pull/993#issuecomment-1938606477
- drop_stats and stats_decimals added, but only drop_stats documented?

pyaerocom/mathutils.py
```diff
@@ -189,7 +189,9 @@ def _nanmean_and_std(data):
 @ignore_warnings(
     RuntimeWarning, "An input array is constant", "invalid value encountered in .*divide"
 )
-def calc_statistics(data, ref_data, lowlim=None, highlim=None, min_num_valid=1, weights=None):
+def calc_statistics(
+    data, ref_data, lowlim=None, highlim=None, min_num_valid=1, weights=None, drop_stats=None
+):
```
heikoklein, line 195: new parameter drop_stats without any documentation

heikoklein: on a public function

heikoklein (approved):
The implemented functionality looks good keeping in mind that is not the long-term solution. There is quite a bit of documentation given in this PR discussion https://github.com/metno/pyaerocom/pull/993#issuecomment-1938606477 on how to use the functionality. This documentation will get lost when the PR is closed and should be put into the online documentation. Approve this PR since documentation could be done directly on rtd.

PR #998, opened 2024-02-16
run AeroVal experiment from CLI

avaldebe:


lewisblake (commented):
We have tests for the CLI. What do you think about adding a test for this?

avaldebe:
I do not think that running an experiment would work as a test.
Instead I would create a fixture that prevents `proc.run()` for doing anything.
Would this could work as a test?

lewisblake:
We do have configs which we currently use in CI https://github.com/metno/pyaerocom/blob/main-dev/tests/fixtures/aeroval/cfg_test_exp1.py.  Creating a fixture that prevents proc.run() for doing anything could be a good way to test it without adding too much compute time to CI.

heikoklein:
The idea of having a proper command-line interface matches very well the with Usability Guidelines for AeroTools. We haven't decided on a proper configuration-format yet, and json is rather a data-transfer format that a configuration language, in particular since readability is not enforced, and comments are not part of the json standard.
Since we have already json-files which can be used for configuration, please call the the cli something like `aeroval_json` rather than calling it **the** aeroval config.

lewisblake (approved):
:unicorn:

PR #999, opened 2024-02-16
Logging improvements

### LINKED ISSUE #730: Reclassify "all" logging categories
To make the logs readable...

heikoklein:
Reducing log-levels to WARN (stdout) and INFO (file)
Enabling and documenting how to  modify log-level in user-space using a logging.ini file.

closes #730


tests/test__logging.py
lewisblake, line 1: Can you add tests to the new logging functionality?

docs/api.rst
```diff
@@ -3,6 +3,65 @@ Core API
...
+
+  1. ``info``-messages or worse are logged to ``pyaerocom.log.$PID`` or
+   (dynamic feature) the file given in the environment variable ``PYAEROCOM_LOG_FILE``
+  2. ``warning``-messages or worse are also printed on stdout.
+   (dynamic feature) Output to stdout is disabled if the script is called non-interactive.
+
+Putting a file with the name ``logging.ini`` in the scripts current working directory will use that
+configuration instead of above described default. An example ``logging.ini`` doing about the same as
+described above, except for the dynamic features, and enable ``debug`` logging on one package,
+here ``pyaerocom.io.ungridded``:
```
lewisblake: maybe add "is provided"

lewisblake (commented):
Small things. Looks good and thanks for documentation. Would be good to include tests if possible.

tests/test__logging.py
heikoklein: Sorry, no tests here. I don't test python logging works, so that we can change it if needed.

The only new functionality is that `logging.ini` from the working directory is automatically read. I could test that, but while this is a good and compatible way forward to the existing logging, it is still a design flaw of not having an application: logging.ini should be initialized by an application, not by a library. I don't write tests for design flaws.

docs/api.rst
heikoklein: done

lewisblake (approved):
:unicorn:

lewisblake:
Having played with this a bit more, I wonder if 

- [ ] we should create a `logs` directory which is the default location for the logs to go. Currently they fill up the directory.
- [ ] add logging.ini to the .gitignore (and `logs/*` if we choose to put the default there)
- [ ] create a `pya` command line which clears the logs

heikoklein:
I don't think logging.ini and logs/* should be added to .gitignore of pyaerocom since these files should be in the run-time directories, and not in the development tree.

I like the idea with a default logs/* directory, we move from default pyaerocom.log to pyaerocom.log.$PID anyway so going to logs/pyaerocom.log.$PID is fine.

pya command for clearing the logs raises a few questions: which log-files? The user-defined log-files from logging.ini or the default ones? What about: If the log-defaults are used (i.e. no user logging.ini), log-files are automatically deleted if older than 7 days? (this was the original intention of the old-logging, but it never worked.) We might delete some information which might be needed after a while, but it is only logging, so that is ok I think.

lewisblake:
> I don't think logging.ini and logs/* should be added to .gitignore of pyaerocom since these files should be in the run-time directories, and not in the development tree.

In theory, sure. In practice, for developers anyway, these are often the same place. Adding it to the .gitignore would only benefit developers, not be an issue for package users not working in the development tree. This has already been an issue for me so I included it in one of my PRs.

> I like the idea with a default logs/* directory, we move from default pyaerocom.log to pyaerocom.log.$PID anyway so going to logs/pyaerocom.log.$PID is fine.

Cool. I'll make an issue and reference this comment,

> pya command for clearing the logs raises a few questions: which log-files? The user-defined log-files from logging.ini or the default ones? What about: If the log-defaults are used (i.e. no user logging.ini), log-files are automatically deleted if older than 7 days? (this was the original intention of the old-logging, but it never worked.) We might delete some information which might be needed after a while, but it is only logging, so that is ok I think.

If the logs are in a `log` directory, then I was thinking of a command which just removed everything in this directory (reduces the chance that folks mess up a rm *-ish command). That said, I like your idea too.

PR #1009, opened 2024-02-23
Introduce FLOAT_DECIMALS (consequently refactor JSON utilities)

### LINKED ISSUE #1008: Make the precision with which the JSON output files are written configurable
https://github.com/metno/pyaerocom/blob/7aa06f50251e6016b8285ed693d7378b881b897b/pyaerocom/_lowlevel_helpers.py#L18

This function has fixed `precision=5`, which writes the JSON files with 5 decimals. This is too imprecise for some variables (e.g., `vcdno2` in units of mol m-2). Changing the default to something higher (e.g., 10) would have adverse effects on the contour plots, rendering them too heavy to quickly display on the web, which would negatively effect some project experiments. Hence, a pyaeroval user should be able to configure this precision argument based on project needs.

Implement an argument, for example `round_floats_precision`, which can be provided in a pyaeroval configuration file.

 - Would you be able to work on this solution yourself?
 Yes 

**Additional context**
Without this precision level, plots on Aeroval can look "blocky", like a step function, which is undesirable. Initial testing increasing the precision argument default to 10 make the plots look much better.

lewisblake:
Closes #1008


lewisblake:
This PR started with the desire to "Make the precision with which the JSON output files are written configurable". In order to do this, it turned out to be "cleaner" to refactor the code quite a bit. I think this is a small step towards separating out `pyaerocom/aeroval` into it's own package.

#### What's changed?
- Functions which are JSON-related have been moved into a module `pyaerocom/aeroval/json_utils.py`. Previously they were being imported into modules in `pyaerocom/aeroval` from `pyaerocom/_lowlevel_helpers.py`, which blurred the lines between `pyaerocom` and `pyaerocom/aeroval` (pyaeroval). Since all the functions which write json files are in pyaeroval, it makes sense that JSON-related functions were moved into `pyaerocom/aeroval`.
- An attribute `round_floats_precision` has been added to the `StatisticsSetup` class https://github.com/metno/pyaerocom/blob/7aa06f50251e6016b8285ed693d7378b881b897b/pyaerocom/aeroval/setupclasses.py#L100 which enables a Modeler to set a `round_floats_precision` argument in a pyaeroval experiment configuration file. Previously this value was harded coded in `round_floats()` https://github.com/metno/pyaerocom/blob/7aa06f50251e6016b8285ed693d7378b881b897b/pyaerocom/_lowlevel_helpers.py#L18 to 5, and this default value is maintained by the new attribute. 
-  `round_floats` can not be set up with a default any longer without breaking the intended functionality (I think this is due to how it is recursively called.), so instead the function `write_json` now accepts a precision argument, also with a default of 5, and `round_floats()` asserts that this argument is provided. Note that all config sets ups need access to `write_json` but not all instances of `write_json` have access to a configuration file. Hence why both have a default, but happy if we can come up with a cleaner solution.
- Methods related to experiments have been moved out of https://github.com/metno/pyaerocom/blob/7aa06f50251e6016b8285ed693d7378b881b897b/pyaerocom/aeroval/experiment_output.py#L34 and instead into https://github.com/metno/pyaerocom/blob/7aa06f50251e6016b8285ed693d7378b881b897b/pyaerocom/aeroval/experiment_output.py#L114
- Unused code has been removed (`JSONFile` class ??), tests modified and/or removed, logging.ini added to the .gitignore.

.gitignore
```diff
@@ -1,5 +1,5 @@
 pyaerocom.log*
-
+logging.ini
```
avaldebe, line 2: why do you want to ignore the logging configuration?

pyaerocom/aeroval/coldatatojson_helpers.py
```diff
@@ -53,7 +53,7 @@ def get_json_mapname(obs_name, var_name_web, model_name, model_var, vert_code, p
     return f"{obs_name}-{var_name_web}_{vert_code}_{model_name}-{model_var}_{period}.json"
 
 
-def _write_stationdata_json(ts_data, out_dir):
+def _write_stationdata_json(ts_data, out_dir, precision):
```
avaldebe: could you add type to `precision`?

pyaerocom/aeroval/coldatatojson_helpers.py
```diff
@@ -79,17 +79,17 @@ def _write_stationdata_json(ts_data, out_dir):
     else:
         current = {}
     current[ts_data["model_name"]] = ts_data
-    write_json(current, fp, ignore_nan=True)
+    write_json(current, fp, precision, ignore_nan=True)
 
 
-def _write_site_data(ts_objs, dirloc):
+def _write_site_data(ts_objs, dirloc, precision):
```
avaldebe: I see the number of decimals as "global config".

Would it be possible to define the precision a constant? e.g. at the top of this module, or in a config module?

avaldebe (commented):
I'm 100% behind the code reconstruction part of this PR (move all JSON functions to own module under aeroval)

I'm not so sure about the implementation of the float precision on serialization part of this PR.

If writing JSONs with a different precision is a "once in a blue moon" it could be implemented as a module "CONSTANT".
if different JSON files need different precision, maybe a contect manager would be simpler to use.

.gitignore
lewisblake: If provided, it's user specific and not part of the code history.

lewisblake:
> I'm 100% behind the code reconstruction part of this PR (move all JSON functions to own module under aeroval)

Thanks. I agree this PR contains two fairly distinct pieces, the precision part motivating the refactoring.

> I'm not so sure about the implementation of the float precision on serialization part of this PR.
>
> If writing JSONs with a different precision is a "once in a blue moon" it could be implemented as a module "CONSTANT". if different JSON files need different precision, maybe a contect manager would be simpler to use.

To be honest, it probably is a bit of a "once in a blue moon" thing, but perhaps only because the functionality hasn't been available yet. I think it's preferable that as someone running a pyaeroval experiment, I have the ability to control this in someway. It seems natural to me that this would be part of "how I want the statistics", and therefore `StatisticsSetup`, but I agree that there is a lot of code changed in order to implement this functionality.

We could have inside `json_utils.py` a constant `JSON_PRECISION=5` so that "Modelers" who want to change this could do something like 

```python
from aeroval.json_utils import JSON_PRECICION
JSON_PRECISION = 7
```  
in their config files, but I also see some downsides to this appraoch. Such an implementation would be fairly unique to this parameter, not be picked up by RTD, and how would this work with the future configuration formats we want to move to (e.g., the JSON one we can support as of today or perhaps a TOML version)?

If you prefer it, I can change it to this way, or come by my office and we can chat about a better solution.

.gitignore
avaldebe: `pyaerocom/logging.ini` is part of the repository, this change ignores all changes to that file

avaldebe:
> ```python
> from aeroval.json_utils import JSON_PRECICION
> JSON_PRECISION = 7
> ```
> 
> in their config files, but I also see some downsides to this appraoch. Such an implementation would be fairly unique to this parameter, not be picked up by RTD, and how would this work with the future configuration formats we want to move to (e.g., the JSON one we can support as of today or perhaps a TOML version)?
> 
> If you prefer it, I can change it to this way, or come by my office and we can chat about a better solution.

I'm not sure what is the best way to provide this functionality. Let's talk about the details once we have ironed out the code restructuration.

avaldebe (approved):
🚀 🎆

.gitignore
avaldebe: I was wrong about the ignore behavior

lewisblake: Det går bra. Takk for hjelpen!

PR #1014, opened 2024-02-27
Cleanup using autoflake

lewisblake:
Clean up code with [autoflake](https://pypi.org/project/autoflake/) to remove usused import


avaldebe (approved):
nice clean up.
maybe we could add a cleanup stage to pre-commit?

lewisblake:
Thanks. Maybe, but it would need some configuring.

PR #1017, opened 2024-02-29
Pydantic for Aeroval setupclasses.py

lewisblake:
Playing with pydantic to see what structure in pyaerocom/aeroval/setupclasses.py can be simplified / flattened.


lewisblake:
Designed to close #1031

lewisblake:
Note: https://github.com/metno/pyaerocom/blob/5ff89df968047ea55a5b1712ff4ef014227e1587/tests/aeroval/test_experiment_output.py#L200 doesn't pass when I run it but does pass in the debugger

Edit: was probably a race condition. Adding `assert` statements verify the paths prior solves this issue.

lewisblake:
Marking this as ready for review but I would really appreciate it if I could get some feedback on some of the implementation details. Starting lines 308 and 336 in `setupclasses.py` there are some comments which explain in more detail what I would like feedback on. Also happy to go through this in person @avaldebe after you have chance to look at it if you have any suggestions.

pyaerocom/aeroval/setupclasses.py
```diff
@@ -88,15 +96,12 @@ def get_json_output_dirs(self, assert_exists=True):
         return out
 
 
-class ModelMapsSetup(ConstrainedContainer):
-    maps_freq = EitherOf(["monthly", "yearly"])
+class ModelMapsSetup(BaseModel):
+    maps_freq: Literal["monthly", "yearly"] = "monthly"
+    maps_res_deg: PositiveInt = 5
```
avaldebe, line 104: The original class supported `kwargs`, was that feature redundant?

pyaerocom/aeroval/setupclasses.py
```diff
@@ -154,40 +160,31 @@ class StatisticsSetup(ConstrainedContainer):
...
-        self.main_freq = self.DEFAULT_FREQS[0]
-        self.freqs = self.DEFAULT_FREQS
-        self.add_seasons = True
-        self.periods = []
+    MIN_NUM: PositiveInt = 1
+    weighted_stats: bool = True
+    annual_stats_constrained: bool = False
+    add_trends: bool = False
+    trends_min_yrs: PositiveInt = 7
+    stats_tseries_base_freq: str | None = None
```
avaldebe, line 171: Why are some type annotations `X | None` (PEP 604, Python 3.10+) and others `Optional[X]`?
Does `pyndantic` supports PEP 604 style annotations?
If so, please use consistently.

pyaerocom/aeroval/setupclasses.py
```diff
@@ -154,40 +160,31 @@ class StatisticsSetup(ConstrainedContainer):
...
+    use_fairmode: bool = False
+    use_diurnal: bool = False
+    obs_only_stats: bool = False
+    model_only_stats: bool = False
+    drop_stats: tuple[str, ...] = ()
+    stats_decimals: int | None = None
+    round_floats_precision: Optional[int] = None
```
avaldebe: why `Optional[int]` instead of `int | None`?

pyaerocom/aeroval/setupclasses.py
```diff
@@ -154,40 +160,31 @@ class StatisticsSetup(ConstrainedContainer):
...
+    if round_floats_precision:
+        set_float_serialization_precision(round_floats_precision)
+
+
+class TimeSetup(BaseModel):
+    DEFAULT_FREQS: Literal["monthly", "yearly"] = "monthly"
+    SEASONS: list[str] = ["all", "DJF", "MAM", "JJA", "SON"]
+    main_freq: str = "monthly"
+    freqs: list[str] = ["monthly", "yearly"]
+    periods: list[str] = []
```
avaldebe: list are mutable, maybe this default should be writen as `Field(default_factory=list)`?

pyaerocom/aeroval/setupclasses.py
```diff
@@ -223,112 +220,215 @@ def _get_all_period_strings(self):
...
+    # It's hard to come up with a way to do this that doesn't introduce breaking changes
+    # Introducing breaking changes only seems appropriate after a config format is settled.
+    # Note: all these computed fields could be more easily defined if the config were
+    # rigid enough to have they explicitly defined (e.g., in a TOML file), rather than dumping everything
+    # into one large config dict and then dishing out get the relevant parts to each class.
+    @computed_field
+    @property
+    def time_cfg(self) -> TimeSetup:
+        if hasattr(self, "model_extra") & bool(
+            time_cfg_keys := set(self.model_extra).intersection(set(TimeSetup.model_fields))
```
avaldebe: Is `time_cfg_keys` mean to be the intersection of `set(self.model_extra)` and `set(TimeSetup.model_fields)`?

If so, maybe this cam be refactor:
- without the `intersection` method as `time_cfg_keys := set(self.model_extra) & set(TimeSetup.model_fields)`; or
- without the 2nd set as `time_cfg_keys := set(self.model_extra).intersection(TimeSetup.model_fields)`

avaldebe: also, the `&` between the booleans is better written as `and`, like this
`if hasattr(self, "model_extra") and bool(...)`

pyaerocom/aeroval/setupclasses.py
```diff
@@ -223,112 +220,215 @@ def _get_all_period_strings(self):
...
+        ):
+            subset_dict = {k: self.model_extra[k] for k in time_cfg_keys}
```
avaldebe: Is `time_cfg_keys` really needed?
How is this different than:
```Python
    def time_cfg(self) -> TimeSetup:
        if not hasattr(self, "model_extra"):
            return TimeSetup()
        
        model_args = {key: val for key, val in self.model_extra.items() if key in TimeSetup.model_fields}
        TimeSetup(**model_args)
```

pyaerocom/aeroval/setupclasses.py
```diff
@@ -223,112 +220,215 @@ def _get_all_period_strings(self):
...
+            return TimeSetup()
+
+    @computed_field
+    @property
+    def modelmaps_opts(self) -> ModelMapsSetup:
+        if hasattr(self, "model_extra") & bool(
+            cfg_extra_keys := set(self.model_extra).intersection(set(ModelMapsSetup.model_fields))
+        ):
+            subset_dict = {k: self.model_extra[k] for k in cfg_extra_keys}
+            return ModelMapsSetup(**subset_dict)
```
avaldebe: I think this is simpler

``` python
    def modelmaps_opts(self) -> ModelMapsSetup:
        if not hasattr(self, "model_extra"):
            return ModelMapsSetup()

        model_args = {key: val for key, val in self.model_extra.items() if key in ModelMapsSetup.model_fields}
        return ModelMapsSetup(**model_args)
```

pyaerocom/aeroval/setupclasses.py
```diff
@@ -223,112 +220,215 @@ def _get_all_period_strings(self):
...
+    # def _update(self, obj: Generic[T]) -> Generic[T]:
+    #     if hasattr(self, "model_extra"):
+    #         extra_obj_cfg_keys = set(self.model_extra).intersection(set(obj.__fields__))
+    #         if extra_obj_cfg_keys:
+    #             subset_dict = {k: self.model_extra[k] for k in extra_obj_cfg_keys}
+    #             return obj(**subset_dict)
+    #         else:
+    #             return obj()
+    #     else:
+    #         return obj()
```
avaldebe: please remove dead code

pyaerocom/aeroval/setupclasses.py
```diff
@@ -223,112 +220,215 @@ def _get_all_period_strings(self):
...
+            subset_dict = {k: self.model_extra[k] for k in cfg_extra_keys}
+            # need to pass some default values to the ColocationSetup if not provided in config
+            default_dict = {"save_coldata": True, "keep_data": False, "resample_how": "mean"}
+            for key in default_dict:
+                if key not in subset_dict:
+                    subset_dict[key] = default_dict[key]
+
+            return ColocationSetup(**subset_dict)
+        else:
+            return ColocationSetup(save_coldata=True, keep_data=False, resample_how="mean")
```
avaldebe: this could be written in a simpler way, like on my previous comments
``` python
    def colocation_opts(self) -> Collocation:
        if not hasattr(self, "model_extra"):
            return ColocationSetup(save_coldata=True, keep_data=False, resample_how="mean")
        
        model_args = ...
```

pyaerocom/aeroval/setupclasses.py
```diff
@@ -223,112 +220,215 @@ def _get_all_period_strings(self):
...
+    @computed_field
+    @property
+    def webdisp_opts(self) -> WebDisplaySetup:
+        if hasattr(self, "model_extra") & bool(
+            cfg_extra_keys := set(self.model_extra).intersection(set(WebDisplaySetup.model_fields))
+        ):
+            subset_dict = {k: self.model_extra[k] for k in cfg_extra_keys}
+            return WebDisplaySetup(**subset_dict)
+        else:
+            return WebDisplaySetup()
```
avaldebe: same as before

```python
    def webdisp_opts(self) -> WebDisplaySetup:
        if not hasattr(self, "model_extra"):
            return WebDisplaySetup()

         model_args = ...
```

pyaerocom/aeroval/setupclasses.py
```diff
@@ -223,112 +220,215 @@ def _get_all_period_strings(self):
...
+    @computed_field
+    @property
+    def processing_opts(self) -> EvalRunOptions:
+        if hasattr(self, "model_extra") & bool(
+            cfg_extra_keys := set(self.model_extra).intersection(set(EvalRunOptions.model_fields))
+        ):
+            subset_dict = {k: self.model_extra[k] for k in cfg_extra_keys}
+            return EvalRunOptions(**subset_dict)
+        else:
+            return EvalRunOptions()
```
avaldebe: idem

pyaerocom/aeroval/setupclasses.py
```diff
@@ -223,112 +220,215 @@ def _get_all_period_strings(self):
...
+    @computed_field
+    @property
+    def statistics_opts(self) -> StatisticsSetup:
+        if hasattr(self, "model_extra") & bool(
+            cfg_extra_keys := set(self.model_extra).intersection(set(StatisticsSetup.model_fields))
+        ):
+            subset_dict = {k: self.model_extra[k] for k in cfg_extra_keys}
+            return StatisticsSetup(**subset_dict)
+        else:
+            return StatisticsSetup(weighted_stats=True, annual_stats_constrained=False)
```
avaldebe: idem

pyaerocom/aeroval/setupclasses.py
```diff
@@ -223,112 +220,215 @@ def _get_all_period_strings(self):
...
+
+    # ObsCollection and ModelCollection require special attention b/c they're not based on Pydantic.
+    # TODO Use Pydantic for ObsCollection and ModelCollection
+    obs_cfg: ObsCollection | dict = ObsCollection()
+
+    @field_validator("obs_cfg")
+    def validate_obs_cfg(cls, v):
+        if not isinstance(v, ObsCollection):
+            return ObsCollection(v)
+        return v
```
avaldebe: think this is better the other way arround
``` python
    def validate_obs_cfg(cls, v) -> ObsCollection:
        if isinstance(v, ObsCollection):
            return v
        return ObsCollection(v)
        
```

pyaerocom/aeroval/setupclasses.py
```diff
@@ -223,112 +220,215 @@ def _get_all_period_strings(self):
...
+
+    @field_serializer("obs_cfg")
+    def serialize_obs_cfg(self, obs_cfg: ObsCollection):
+        return obs_cfg.json_repr()
+
+    model_cfg: ModelCollection | dict = ModelCollection()
```
avaldebe, line 433: this variable will get lost among all the class methods,
is it possible to move it up with the rest of the variable annotations?

pyaerocom/aeroval/setupclasses.py
```diff
@@ -223,112 +220,215 @@ def _get_all_period_strings(self):
...
+
+    @field_validator("model_cfg")
+    def validate_model_cfg(cls, v):
+        if not isinstance(v, ModelCollection):
+            return ModelCollection(v)
+        return v
```
avaldebe: think it looks better the other way arroung
``` python
    def validate_model_cfg(cls, v) -> ModelCollection:
        if isinstance(v, ModelCollection):
            return v
        return ModelCollection(v)
```

pyaerocom/aeroval/setupclasses.py
```diff
@@ -223,112 +220,215 @@ def _get_all_period_strings(self):
...
+
+    @field_serializer("model_cfg")
+    def serialize_model_cfg(self, model_cfg: ModelCollection):
+        return model_cfg.json_repr()
 
     @property
```
avaldebe: no `computed_field` decorator?

pyaerocom/aeroval/setupclasses.py
```diff
@@ -396,16 +496,19 @@ def to_json(self, outdir: str, ignore_nan: bool = True, indent: int = 3) -> None
         return filepath
 
     @staticmethod
-    def from_json(filepath: str) -> "EvalSetup":
+    def from_json(filepath: str) -> EvalSetup:
```
avaldebe: maybe annotate with return `Self` from [PEP 673]?

[PEP 673]: https://peps.python.org/pep-0673/

tests/aeroval/test_aeroval_HIGHLEV.py
```diff
@@ -102,21 +103,6 @@ def test_reanalyse_existing(eval_config: dict):
...
-    assert output.is_dir()
-    assert list(output.glob("**/*.nc"))
-
-    proc.exp_output.delete_experiment_data(also_coldata=True)
-    assert not output.exists()
-
-    proc.run()
-    assert output.is_dir()
-    assert list(output.glob("**/*.nc"))
-
```
avaldebe, line 119: are this `asserts` no longer relevant?

avaldebe (changes requested):
mostly minor changes. The classes still mix data and behavior, but changing that is out of the scope of this PR.

pyaerocom/aeroval/setupclasses.py
lewisblake: Yes. Will do.

pyaerocom/aeroval/setupclasses.py
lewisblake: `model_cfg` and `obs_cfg` are lumped together at the bottom because they are based on `ObsCollection` and `ModelCollection` which are not (yet) based on pydantic. I have added comments to more clearly outline which parts of the code are pydantic-based attributes, non-pydantic-base attributes, and methods.

pyaerocom/aeroval/setupclasses.py
lewisblake: Agreed.

pyaerocom/aeroval/setupclasses.py
lewisblake: No, I don't think so. I think this is just a method which tells pydantic how to serialize this particular attribute, because it doesn't know on it's own.

pyaerocom/aeroval/setupclasses.py
lewisblake: Sure, good suggestion. `EvalSetup` doesn't currently have any child classes but good idea to get ahead of any issue.

tests/aeroval/test_aeroval_HIGHLEV.py
lewisblake: No, these tests were overly verbose and the functionality covered by reparameterizing the test to accept a boolean `reanalyse_existing`. 

This test highlights one of the "bad practices", that only really happens in the tests as far as I can tell. With the current setup of this Pydantic-based implementation, once the config is defined, it is immutable. So one can not change attributes in `cfg` like this anymore by setting them. They will not change. You must change the actual config dictionary prior to giving it to `EvalSetup`, and hence these tests were broken and needed to be rewritten.

pyaerocom/aeroval/setupclasses.py
lewisblake: Agreed this looks better.

pyaerocom/aeroval/setupclasses.py
lewisblake: As far as I can tell, yes. Moreover, we don't want these classes taking keyword arguments which aren't defined. Part of this rewrite is to rigorously define the configuration interface, so passing arguments which are not present should raise an error.

pyaerocom/aeroval/setupclasses.py
lewisblake: Agreed. Will change.

pyaerocom/aeroval/setupclasses.py
lewisblake: This is cleaner. Changed.

pyaerocom/aeroval/setupclasses.py
```diff
@@ -1,10 +1,15 @@
...
 from getpass import getuser
 from pathlib import Path
-from typing import Literal, Self
+
+if sys.version_info < (3, 10):
+    from typing_extensions import Self
+else:
+    from typing import Self
+
+from typing import Literal
```
avaldebe: `Self` was added to the `typing` module on Python 3.11, see [PEP 673](https://peps.python.org/pep-0673/), so you want the if written as follows

``` python
from typing import Literal

if sys.version_info >= (3, 11):
    from typing import Self
else:
    from typing_extensions import Self
```

also by moving the `Literal` import up, the conditional import sits between imports from the standard library and from 3rd party modules.

avaldebe (changes requested):
please make sure that `typing_extensions` is an explicit dependency.
Accordig to the [typing_extensions changelog](https://github.com/python/typing_extensions/blob/main/CHANGELOG.md#release-401-november-30-2021) any version after 4.0.1 should work.


On `pyproject.toml` you'll need to add the following line

``` diff
dependencies = [
[...]
    # python < 3.11
    'tomli>=2.0.1; python_version < "3.11"',
    'importlib-resources>=5.10; python_version < "3.11"',    
+   'typing-extensions>=4.0.1; python_version < "3.11"',
```

The equivalent line on `pyaerocom_env.yml` would be
``` diff
dependencies:
[...]
## python < 3.11
  - tomli >=2.0.1
  - importlib-resources >=5.10
+ - typing-extensions >=4.0.1
```

pyaerocom/aeroval/setupclasses.py
lewisblake: Weird, this is an outdated version. In the newer version of the code check to see if `sys.version_info < (3, 11)`, because of the reason you mention. I agree with you styling of the import though, so I will change it to a >= 3.11 statement so it is organized this way.

avaldebe (approved):
✅

PR #1018, opened 2024-03-01
Smaller logging improvements

### LINKED ISSUE #1016: Remove print-statments from recent PR and use logging instead
**Describe the bug**
- The recent PR https://github.com/metno/pyaerocom/pull/976 introduced many `print` statements where logging should be used

**To Reproduce**
1. run a pyaerocom script
2. output appears in STDOUT but not in the logs/pyaerocom.logs.$PID

**Expected behavior**
STDOUT should have by default all logger.warnings and higher, while logger.info and higher should be in the logs.
All output should be configurable by logging.ini

heikoklein:
This PR converts print-statements to logger-statements and addresses #1016 .
It also makes logging of metadatastandards-deviations rather a info than a warnings. The aerocom had 3 metadatastandards while pyaerocom tests for a 4th and newer metadatastandard which isn't really needed yet.

closes #1016

### DIFF
--- a/pyaerocom/io/readungridded.py
+++ b/pyaerocom/io/readungridded.py
@@ -147,7 +147,6 @@ class ReadUngridded:
     def INCLUDED_DATASETS(self):
         lst = []
         for reader in self.INCLUDED_READERS:
-            logger.debug(f"{reader}: {reader.SUPPORTED_DATASETS}")
             lst.extend(reader.SUPPORTED_DATASETS)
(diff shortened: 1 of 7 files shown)


pyaerocom/io/readungridded.py
dulte, line 1: Both print/debug statements can be removed

dulte (changes requested):

dulte (approved):

PR #1020, opened 2024-03-01
Removes debug print

dulte:


dulte:
Redundant. Same as PR #1018

PR #1028, opened 2024-03-12
fixes wrongly assigned var_name when using use keyword closes #1027

### LINKED ISSUE #1027: The "use" keyword in variables.ini changes the variable before it uses the alias definition in aliases.ini
**Describe the bug**
The issue occurs when reading gridded model data using the pyaercom api. 
Several of the optical depth variables have a default "use" variable defined in variables.ini e.g:
```
[abs550dust]
description = Absorption aerosol optical depth (AAOD) at 870 nm due to dust
standard_name = atmosphere_absorption_optical_thickness_due_to_dust_ambient_aerosol_particles
use = abs550aer
```
This tell pyaerocom the variable to use if it does not find abs550dust and will use abs550aer instead. However abs550dust have an alias defined in  aliases.ini `abs550dust = abs550du` so it does exist. However what pyaercom does because of the use keyword being set to "abs550aer". Is to take abs550dust to be abs550aer and then attempts to read abs550aer, which does not exist for this model.

 - Pyaerocom version: 0.16.0
 - Computing platform: ubuntu (Met laptop)
 - Configuration file (if applicable): Na
 - Error message (if applicable): Na

**To Reproduce**
This example will fail because it does not find abs550dust
```python
import pyaerocom as pya
data = pya.io.ReadGridded(data_dir='/lustre/storeB/project/aerocom/aerocom-users-database/DURF/histSST/GEOS-i32_DURF-histSST/')

abs_data = data.read_var('abs550dust',start=1995,stop=1996)

```
if in  [variable.ini](https://github.com/metno/pyaerocom/blob/1b893324b9537dbdd43156e150b71dff990350d6/pyaerocom/data/variables.ini#L223) the `use = abs550aer` is removed it will work fine.

**Expected behavior**
Look for alias variables before possibly looking for the variable defined by the use keyword in variables.ini. 

**Additional context**
I would remove all use keyword from variables.ini and rather have the user define whether it should use another variable if the preferred one is not available. Otherwise we risk comparing apples and oranges, without knowing. In addition the error you get if the variable defined by the use keyword is missing does not tell you what the problem is. Removing the use keyword in the variable.ini file make pyaerocom behave as expected.

Ovewh:
The use keyword lead to variable getting assigned wrong var_name and thus leading to pyaerocom looking for aliases of the wrong variable

### DIFF
--- a/pyaerocom/data/variables.ini
+++ b/pyaerocom/data/variables.ini
@@ -21,170 +21,202 @@ comments_and_purpose = For comparison with AERONET and satellite measurements
 [od550lt1aer]
 description = AOD due to fine aerosol at 550 nm (particle smaller than D=1 um)
 use = od550aer
+var_name = od550lt1aer
 
 [od550gt1aer]
 description = AOD due to fine aerosol at 550 nm (particle greater than D=1 um)
 use=od550aer
+var_name = od550gt1aer
 
 [od550lt1ang]
 description = AOD filtred by Angstrom exponent less than 1
 use = od550aer
-
+var_name = od550lt1ang
 
 [od440aer]
 description = Aerosol optical depth (AOD) at 440 nm
 wavelength_nm = 440
 use = od550aer
+var_name = od440aer
 
 [od865aer]
 description = Aerosol optical depth (AOD) at 865 nm
 use = od550aer
+var_name = od865aer
 
 [od870aer]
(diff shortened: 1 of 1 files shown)


lewisblake:
Thanks! Wait for #1029 to be merged into main-dev, integrate those changes here, and then I will review the PR once tests are passing.

lewisblake (approved):
:unicorn: :stars:

PR #1029, opened 2024-03-12
Fix pytest 8 warning

avaldebe:
with pytest 8.1+, CI fails with this error
```
@lustre_unavail
@pytest.fixture(scope="module")constructs are deprecated: https://docs.pytest.org/en/stable/deprecations.html#applying-a-mark-to-a-fixture-function.
```

this PR fixes it


heikoklein (approved):
Thanks, nice solution.

PR #1049, opened 2024-03-14
Use continue instead of pass & implement tests for `vert_coords.py`

### LINKED ISSUE #1047: is this a bug? or it is masking a bug elsewhere?
https://github.com/metno/pyaerocom/blob/f4f5c3b0823b91a03eff9d2975d8fa8bb95b4d07/pyaerocom/vert_coords.py#L446-L448

`if` clause does nothing, I suspect that this was supposed to be something like

``` python
 for coord in coord_list: 
     if coord in self: 
         continue
```

but I do not know enough about this module to say for certain

lewisblake:
## Change Summary

Fixes `pass` instead of `continue` error in `vert_cords.py` and increases test coverage of `vert_coords.py` by implementing tests and removing unused code.

## Related issue number

NA

## Checklist

* [ ] Start with a draft-PR
* [x] Tests pass locally
* [x] Tests pass on CI
* [x] Make PR ready to review


avaldebe (commented):
is this change covered by the test suite?

lewisblake:
No, it has been specifically excluded from the testing coverage. This is probably not desired anymore. https://github.com/metno/pyaerocom/blob/7de51e0885a61e3a7eaa137f3a52f0ecfe981ca8/pyproject.toml#L136

lewisblake:
Will need to write tests for `vert_coords.py` because we using this code elsewhere now.

First step is add to test coverage.

lewisblake:
The code coverage indicates we are not testing this module. There is no `test_vert_coords` either.

thorbjoernl:
There seem to be a number of functions in `vert_coords.py` that aren't being used anywhere in the code base. Can these be removed? This seems to apply to:

- `is_supported()`
- `pressure2altitude()`
- ~~`geopotentialheight2altitude()` (Which is also unimplemented)~~

tests/test_vert_coords.py
```diff
@@ -0,0 +1,125 @@
...
+def test_atmosphere_hybrid_sigma_pressure_coordinate_to_pressure(
+    a: np.ndarray, b: np.ndarray, ps: float, p0: float | None, exception
+):
+    if exception is not None:
+        with pytest.raises(exception):
+            atmosphere_hybrid_sigma_pressure_coordinate_to_pressure(a, b, ps, p0)
+    else:
+        result = atmosphere_hybrid_sigma_pressure_coordinate_to_pressure(a, b, ps, p0)
+
+        assert len(a) == len(b) == len(result)
```
avaldebe: please consider to split into 2 test functions:
- "happy path" test function: no exceptions are raised; and
- "something went wrong" test function: exceptions are expected

tests/test_vert_coords.py
```diff
@@ -0,0 +1,125 @@
...
+
+    def test_search_aux_coords(self, alt: AltitudeAccess):
+        with pytest.raises(CoordinateNameError):
+            alt.search_aux_coords(["gkjfdshglk"])
+
+        assert alt.search_aux_coords(["lat"])
+        assert alt.search_aux_coords("lon")
+        assert not alt.search_aux_coords("z")
+        assert alt.search_aux_coords(["lat", "lon", "time"])
+        assert not alt.search_aux_coords(["lat", "lon", "time", "z"])
```
avaldebe: I see no need for a test class, please consider using test functions instead

pyaerocom/vert_coords.py
```diff
@@ -54,14 +57,17 @@ def atmosphere_sigma_coordinate_to_pressure(sigma, ps, ptop):
...
     if not isinstance(ptop, float):
-        try:
-            ptop = float(ptop)
-        except Exception as e:
-            raise ValueError(f"Invalid input for ptop. Need floating point\nError: {repr(e)}")
+        if not isinstance(ptop, np.ndarray):
+            try:
+                ptop = float(ptop)
+            except ValueError as e:
+                raise ValueError(f"Invalid input for ptop. Need floating point\nError: {repr(e)}")
```
avaldebe: you can simplify the error section with 
``` python
    if not isinstance(ptop, float | np.ndarray):
        try:
            ptop = float(ptop)
        except ValueError as e:
            raise ValueError("Invalid input for ptop. Need floating point") from e
```

Using `|` on `isinstance(ptop, float | np.ndarray)` is a [Python 3.10 feature](https://docs.python.org/3/library/stdtypes.html#types-union).

The mechanism behind `raise ValueError(...) from e` is called [exception chaining](https://docs.python.org/3/tutorial/errors.html#exception-chaining), and will show the original exception as part of the stack trace.

pyaerocom/vert_coords.py
```diff
@@ -112,14 +118,14 @@ def geopotentialheight2altitude(geopotential_height):
 
     Parameters
     ----------
-    geopotential_height
+    geopotential_height : float
```
avaldebe: annotating the return of this function should add this info to the docs,
and it will provide type hints to your IDE
``` python
def geopotentialheight2altitude(geopotential_height: float) -> float:
```

pyaerocom/vert_coords.py
```diff
@@ -112,14 +118,14 @@ def geopotentialheight2altitude(geopotential_height):
...
         input geopotential height values in m
 
     Returns
     -------
     Computed altitude levels
     """
-
+    raise NotImplementedError("This function is a dummy function")
```
avaldebe: can this function be removed?

tests/test_vert_coords.py
```diff
@@ -0,0 +1,125 @@
...
+)
+def test_VerticalCoordinate_lev_increases_with_alt(
+    name: str, expected: bool | None, exception: Exception | None
+):
+    vert = VerticalCoordinate(name)
+    if exception:
+        with pytest.raises(exception):
+            vert.lev_increases_with_alt
+    else:
+        assert vert.lev_increases_with_alt == expected
```
avaldebe: please put the exceptions on a separate test

tests/test_vert_coords.py
```diff
@@ -0,0 +1,125 @@
...
+        pytest.param(1.0, 0, 0.0, id="ptop-is-float"),
+        pytest.param(1.0, 0, np.ones(10), id="ptop-is-ndarray"),
+    ),
+)
+def test_atmosphere_sigma_coordinate_to_pressure(sigma, ps, ptop):
+    result = atmosphere_sigma_coordinate_to_pressure(sigma, ps, ptop)
+
+    assert type(result) == type(ptop)
+    if isinstance(result, np.ndarray):
+        assert len(result) == len(ptop)
```
avaldebe, line 57: Is it possible to check the result on the `else`?
Also, what should be the value?
You can numerical values with `assert result == pytest.approx(expected_reesult)`

tests/test_vert_coords.py
```diff
@@ -0,0 +1,125 @@
...
+
+def test_VerticalCoordinate_exceptions():
+    # Unsupported variable name during initialization.
+    with pytest.raises(ValueError):
+        VerticalCoordinate("jtlkjhsklrg")
+
+    # Attempt calculating pressure for unsupported variable.
+    vert = VerticalCoordinate("asc")
+    with pytest.raises(CoordinateNameError):
+        vert.calc_pressure(np.ones(5))
```
avaldebe: please split this test or use `pytest.mark.parameterize`.
As written, a failure on the 1st assert will shadow a failure on the second.

tests/test_vert_coords.py
```diff
@@ -0,0 +1,125 @@
...
+# TODO: This test is currently failing because of a ValueError caused by
+# a variable name containing an underscore. This seems like an unrelated
+# issue to me, so I have marked it xfail for now.
+def test_AltitudeAccess_check_access(alt: AltitudeAccess):
+    assert alt.check_altitude_access()
+
+
+def test_AltitudeAccess_search_aux_coords(alt: AltitudeAccess):
+    with pytest.raises(CoordinateNameError):
+        alt.search_aux_coords(["gkjfdshglk"])
```
avaldebe, line 147: please put the exception part on a separate test

avaldebe (changes requested):
good effort, please work a bit more on the tests

pyaerocom/vert_coords.py
thorbjoernl: It is being referenced here: https://github.com/metno/pyaerocom/blob/8e09efc1e48105b50bcf385a4f31a0ce53157ce8/pyaerocom/vert_coords.py#L206 but isn't otherwise used. If removing this doesn't break any tests I'll remove it.

thorbjoernl: No tests broke from removing it, so I removed it.

tests/test_vert_coords.py
thorbjoernl: Done

tests/test_vert_coords.py
thorbjoernl: Done.

tests/test_vert_coords.py
thorbjoernl: Done

tests/test_vert_coords.py
thorbjoernl: I've added expected values, and an additional testcase using a sine-wave to cover a broader range of the function domain.

pyaerocom/vert_coords.py
thorbjoernl: Fixed.

test.py
avaldebe, line 1: what is the purpose of this file?
are the methods here the same as the ones on `vert_coords`?

thorbjoernl:
While fixing the CR issues I discovered that there were a number of references to code I had previously removed remaining in the code. These were not called in the code, and mostly used [here](https://github.com/metno/pyaerocom/blob/b664a5046853d7b467c1fb971eabbbbef373c626/pyaerocom/griddeddata.py#L1349) which is an elif clause which always raises an exception, and is thus unlikely to be part of the program flow. I've cut out most of that clause, as well as any logic related to the `vert_cord.py` check_altitude_access() without breaking any tests.

avaldebe:
> While fixing the CR issues I discovered that there were a number of references to code I had previously removed remaining in the code. These were not called in the code, and mostly used [here](https://github.com/metno/pyaerocom/blob/b664a5046853d7b467c1fb971eabbbbef373c626/pyaerocom/griddeddata.py#L1349) which is an elif clause which always raises an exception, and is thus unlikely to be part of the program flow. I've cut out most of that clause, as well as any logic related to the `vert_cord.py` check_altitude_access() without breaking any tests.

I do not understand what are you trying to say hare.
Where the removed, non-implemented methods, used somewhere?
Then, what did you do with the calls?
Did you removed them?

thorbjoernl:
It is basically cleaning up left overs from my original cleanup. I had removed some functions, but missed functions which called on the removed functions. This didn't surface because that path is only in the mentioned elif clause. This just removes the remaining dangling references. So in short, yes, I removed them.

thorbjoernl:
> > While fixing the CR issues I discovered that there were a number of references to code I had previously removed remaining in the code. These were not called in the code, and mostly used [here](https://github.com/metno/pyaerocom/blob/b664a5046853d7b467c1fb971eabbbbef373c626/pyaerocom/griddeddata.py#L1349) which is an elif clause which always raises an exception, and is thus unlikely to be part of the program flow. I've cut out most of that clause, as well as any logic related to the `vert_cord.py` check_altitude_access() without breaking any tests.
> 
> I do not understand what are you trying to say hare. Where the removed, non-implemented methods, used somewhere? Then, what did you do with the calls? Did you removed them?

avaldebe (approved):
🚀 🎆

PR #1077, opened 2024-03-23
3.12 tests moved to ubuntu 24.04 and disabled; PR-checklist updated

### LINKED ISSUE #1076: venv 3.12 tests on Ubuntu 22.04 are too slow
The venv 3.12 tests on Ubuntu 22.04 take twice as long (~16min) than the tests in all other environments (7-8min), see Actions.

3.12 and Ubuntu 22.04 don't match well, 3.12 should be disabled until a better match (24.04) exists.

heikoklein:
## Change Summary

3.12 tests on ubuntu 22.04 are twice as slow as others. Moving therefore 3.12 tests to ubuntu 24.04 and disabling 24.04 since those tests are queued too long.

## Related issue number

closes #1076 

## Checklist

* [x] Start with a draft-PR
* [x] The pull request title is a good summary of the changes
* [x] Tests pass first locally and then on CI
* [x] My PR is ready to review and I have selected a reviewer

### DIFF
--- a/.github/PULL_REQUEST_TEMPLATE.md
+++ b/.github/PULL_REQUEST_TEMPLATE.md
@@ -9,9 +9,12 @@
 
 ## Checklist
 
+* [ ] Tests pass locally
+* [ ] Tests for the changes exist where applicable
 * [ ] Start with a draft-PR
 * [ ] The pull request title is a good summary of the changes
+* [ ] PR is set to AeroTools and milestone
 * [ ] Documentation reflects the changes where applicable
-* [ ] Tests for the changes exist where applicable
-* [ ] Tests pass first locally and then on CI
-* [ ] My PR is ready to review and I have selected a reviewer
+* [ ] Tests pass on CI
+* [ ] At least 1 reviewer is selected
+* [ ] My PR is ready to review
--- a/.github/workflows/CI.yml
+++ b/.github/workflows/CI.yml
@@ -42,9 +42,13 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        python-version: ['3.10', '3.11', '3.12']
+        python-version: ['3.10', '3.11']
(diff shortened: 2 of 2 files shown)


heikoklein:
Currently, CI tests with ubuntu-24.04 are queued (too long/forever?). Disabling therefore tests of 3.12/ubuntu-24-04.

avaldebe:
Are the tests are also slow on `Ubuntu 2024.04`?
Do you know why?
On my local tests the creation of the Python 3.12 environment takes most of the time,
because some packages with c-extensions do not provide wheels for Python 3.12 and need to be compiled from scratch.
If this is the case, the cache GH action should alleviate this somehow.
Maybe the cache is not configured correctly?

Or are the tests themselves slower? If this is the case, maybe we should investigate why are twice as slow.

lewisblake (commented):
I think @avaldebe makes a convincing arguement that the issue is from the caching of pip https://github.com/metno/pyaerocom/issues/1076#issuecomment-2017656840 . Indeed checking the caching of pip and tox in the action, shows that this works on 3.11 and 3.10 but fails on 3.12. So probably removing 3.12 is not the best solution and we should rather fix this issue which should speed up the CI for 3.12 to about the same runtime as the other Python versions.

The PR-checklist should be covered in a separate PR. We decided last week on the format. If you want to change/modify it, please submit a PR and we can discuss it.

lewisblake:
@heikoklein please check the changes in https://github.com/metno/pyaerocom/pull/1079 and determine whether this PR is needed. https://github.com/metno/pyaerocom/pull/1079#issuecomment-2017997510 and checking CI on the merge shows the issue is no longer present.

heikoklein:
Good to see that all tests now take below 10min :+1: Closing this since it is no longer required.

Issue #1114, opened 2024-04-10
Allow configuration and implementation of custom statistics in calc_statistics()


thorbjoernl:
# Summary
Follow up on the discussion [here](https://github.com/metno/pyaerocom/pull/993) on how to allow for configuration and implementation of custom statistics to `mathutils.calc_statistics()`. Relates to #1112

PR #1117, opened 2024-04-12
Refactoring of calc_statistics

### LINKED ISSUE #1112: Wrong logic in mathutils.calc_statistics()`
The logic in `mathutils.calc_statistics`seems subtly wrong (or at least undocumented).

https://github.com/metno/pyaerocom/blob/913db174b59c255a60ce8262c111287b963ce8f2/pyaerocom/mathutils.py#L251

Identified issues:
- Calculations of multiple statistics do not take into account constraints given by `lowlim` or `highlim` (For example `totnum`, `num_valid`, `data_mean`/std, ...)
- This breaks the check for `min_num_valid` since it is based on the unconstrained data count.

thorbjoernl:
## Change Summary

Refactoring of `calc_statistics` and starting to implement functionality for user defined statistics calculations.

- Move statistic calculations from mathutils to its own submodule.
  - `stats/stats.py` implements the main interface (mainly `calculate_statistics()`)
  - `stats/implementations.py`implements the individual stats as functions.
  - `stats/types.py` - Type definitions.
  - `stats/stat_filters.py` - Filtering of stats dictionary (eg. `DropStats`).
  - `stats/data_filters.py` - Filtering of data (eg. lowlim / highlim, NaN removal).
- Refactor `calc_statistics()` into `calculate_statistics`
  - Individual statistics are now implemented in dedicated functions.
  - Clearer division of concerns between processing steps (Filter data, calculation, filter result dict).
- Custom implemented stats can be calculated by passing a dict mapping a name to a function implementing the stat calculation. The interface for exposing this to the user is not yet implemented.
- Remove some superfluous helper functions which check for NaN, when numpy already does that.
- Documentation is a work in progress.
- Tested with old unittests which pass, and output of `cfg_benchmark.py` looks sensible.

## Related issue number

fix #1112 

Relates to #1114 

## Checklist

* [x] Start with a draft-PR
* [x] Tests pass locally
* [x] Tests pass on CI
* [x] Make PR ready to review


pyaerocom/stats/stats.py
```diff
@@ -0,0 +1,268 @@
...
+from pyaerocom.stats.data_filters import FilterByLimit, FilterNaN
+from pyaerocom.stats.implementations import *
+from pyaerocom.stats.stat_filters import FilterDropStats
+from pyaerocom.stats.types import DataFilter, StatisticsCalculator, StatisticsFilter, StatsDict
+
+
+def filter_data(
+    data: np.array,
+    ref_data: np.array,
+    weights: Optional[np.array],
```
lewisblake: Pretty much everywhere you use `Optional` you can instead do `| None`. This is the preferred way of typing hinting Optional since 3.10 and since we dropped 3.9 support, we should be good.

pyaerocom/stats/stats.py
```diff
@@ -0,0 +1,268 @@
...
+    if (stats_filters is None) and (drop_stats is not None):
+        stats_filters = [FilterDropStats(drop_stats)]
+
+    if (lowlim is not None) or (highlim is not None):
+        data_filters.append(FilterByLimit(lowlim=lowlim, highlim=highlim))
+    #### ----- END BACKWARDS COMPATIBILITY
+
+    result = dict()
+
+    result["totnum"] = float(len(data))
```
lewisblake: Does this need to be a float?

pyaerocom/stats/types.py
```diff
@@ -0,0 +1,17 @@
+from typing import Callable, Optional
```
lewisblake: See previous comment about `Optional`

pyaerocom/aeroval/coldatatojson_helpers.py
```diff
@@ -21,9 +21,9 @@
 from pyaerocom.config import ALL_REGION_NAME
 from pyaerocom.exceptions import DataCoverageError, TemporalResolutionError
 from pyaerocom.helpers import start_stop
-from pyaerocom.mathutils import _init_stats_dummy, calc_statistics
 from pyaerocom.region import Region, find_closest_region_coord, get_all_default_region_ids
 from pyaerocom.region_defs import HTAP_REGIONS_DEFAULT, OLD_AEROCOM_REGIONS
+from pyaerocom.stats.stats import _init_stats_dummy, calc_statistics_helper
```
lewisblake: I'll make note of this here although it applies within `stats.py` as well. It is my strong preference that the "helper" function should not be what is imported into other module. I'd prefer that whatever you import into other modules is the "main" function and the "helper" (if is is really needed) is what is used internally in that function.

pyaerocom/stats/stats.py
```diff
@@ -0,0 +1,268 @@
...
+        tuple consisting of the modified data, ref_data and weights array.
+    """
+    if filters is not None:
+        for f in filters:
+            data, ref_data, weights = f(data, ref_data, weights)
+
+    return (data, ref_data, weights)
+
+
+def calculate_statistics(
```
lewisblake: I think this should be the "helper" function. But I would also suggest you try to come up with a more "helpful" name than "helper". What is the core functionality of this function? As you can see, there are "helper" functions all over the codebase and when debugging one finds that naming them as such isn't actually very helpful. I therefore suggest we do not add any more "helper" functions, and rather really try and name them something more useful that will help illuminate their functionality months down the road when we have forgotten exactly what the implementation details are.

pyaerocom/stats/types.py
```diff
@@ -0,0 +1,17 @@
...
+]
+
+# Type definition for a callable which calculates a statistic.
+StatisticsCalculator = Callable[[np.array, np.array, Optional[np.array]], np.float64]
+
+# Type definition for a callable which filters out statistics from the resulting stats dictionary.
+StatisticsFilter = Callable[[dict[str, np.float64]], dict[str, np.float64]]
+
+# Type definition for a stats dictionary.
+StatsDict = dict[str, np.float64]
```
lewisblake, line 12: Could this be a NamedTuple? Similair to https://github.com/metno/pyaerocom/blob/47f0ae27478fd6f9ece8977e33fdee930b22b532/pyaerocom/colocation_3d.py#L39

pyaerocom/stats/stats.py
```diff
@@ -0,0 +1,268 @@
...
+    data : ndarray
+        array containing data, that is supposed to be compared with reference
+        data
+    ref_data : ndarray
+        array containing data, that is used to compare `data` array with
+    data_filters : Optional[list[DataFilter]]
+        list of filter functions applied to the data arrays before calculating stats.
+    statistics : dict[str, StatisticsCalculator]
+        mapping between statistics name and functions to calculate that statistics.
+    stats_filter : Optional[list[StatisticsFilter]]
```
lewisblake: Change documentation about `Optional`

tests/test_pyaerocomstats.py
lewisblake, line 1: If you make a directory under `pyaerocom`, please make a corresponding directory under `tests` and place all tests in there with the name `test_<filename>.py`. There is no `pyaerocomstats.py` module, so this naming could be a little confusing-

pyaerocom/stats/stats.py
```diff
@@ -0,0 +1,268 @@
...
+        raise ValueError("Invalid input. Data arrays must be one dimensional")
+
+    if len(data) != len(ref_data):
+        raise ValueError("Length mismatch between data and ref_data.")
+
+    if weights is not None:
+        if len(weights) != len(data):
+            raise ValueError("Invalid input. Length of weights must match length of data.")
+
+    #### ----- BACKWARDS COMPATIBILITY ------
```
lewisblake: The backwards compatibility thing is not totally clear to me. Which of your changes would introduce breaking changes? Could you please document (in this issue) what would else would needed to be done to the source code in other places to move to this newer style of computing the statistics?

lewisblake (changes requested):
Quick review for now. Will take another, deeper look once you incorporate these first changes. Looks really good! I think this is a really positive step in the right direction. I do have some questions, though. F. eks, how could this implementation facilitate user defined statistics? Can we theoretically already add a module where they would go? It's okay to not have all the answers to some of those longer-term proposals yet, but documenting how this is different to what is was in light of where we want to go to in terms of development would be really useful.

pyaerocom/stats/stats.py
lewisblake: For example `drop_stats` being deprecated? What is the preferred way of setting this in a config file now?

pyaerocom/stats/types.py
```diff
@@ -0,0 +1,17 @@
+from typing import Callable, Optional
+
+import numpy as np
+
+# Type definition for a callable which filters (ie. excludes) data before calculating stats.
+DataFilter = Callable[
```
magnusuMET: Would it make sense for a filter to instead return a mask of the (in)valid data? Saves some boilerplate for the implementor of the function

pyaerocom/stats/stats.py
thorbjoernl: Fixed

pyaerocom/stats/stats.py
thorbjoernl: Fixed

pyaerocom/stats/types.py
thorbjoernl: NamedTuple doesn't support adding additional fields at runtime so I don't think it would work here.

pyaerocom/stats/stats.py
thorbjoernl: Fixed

tests/test_pyaerocomstats.py
thorbjoernl: Fixed

pyaerocom/stats/stats.py
thorbjoernl: Yeah, this is confusing.

I didn't want to change both the tests and the implementation at the same time, to increase confidence in the new implementation, so I designed `calculate_statistics()` to be able to be called exactly the same as the old function. I plan to remove this logic which will simplify it a lot.

The logic of dropping stats would be implemented by passing a `FilterDropStats` instance to `stat_filters` (eg. `calculate_statistics(..., stat_filters=[FilterDropStats(["R", ...])`, though longer term it probably makes more sense to not calculate them in the first place by not including them in the `statistics` dict.

pyaerocom/stats/data_filters.py
```diff
@@ -0,0 +1,45 @@
...
+
+    def __init__(self, lowlim: float | None, highlim: float | None):
+        self.lowlim = lowlim
+        self.highlim = highlim
+
+    def __call__(
+        self, data: np.ndarray, ref_data: np.ndarray, weights: np.ndarray | None
+    ) -> tuple[np.ndarray, np.ndarray, np.ndarray | None]:
+        if self.lowlim is not None:
+            valid = np.logical_and(data > self.lowlim, ref_data > self.lowlim)
```
magnusuMET:
```suggestion
            valid = np.logical_and(data > self.lowlim, ref_data < self.highlim)
```

lewisblake: This should be covered by tests

magnusuMET: Nevermind, I see there are two blocks where this one is for the lower limit.

pyaerocom/stats/data_filters.py
```diff
@@ -0,0 +1,34 @@
...
+
+    def __init__(self, lowlim: float | None, highlim: float | None):
+        self.lowlim = lowlim
+        self.highlim = highlim
+
+    def __call__(
+        self, data: np.ndarray, ref_data: np.ndarray, weights: np.ndarray | None
+    ) -> tuple[np.ndarray, np.ndarray, np.ndarray | None]:
+        if self.lowlim is not None:
+            valid1 = np.logical_and(data > self.lowlim, ref_data > self.lowlim)
```
magnusuMET, line 33: `valid1` is not defined if `self.lowlim is None` (same problem with `valid2`)

pyaerocom/stats/data_filters.py
```diff
@@ -0,0 +1,34 @@
...
+    Filters out data which is outside of a lower and / or upper limit.
+    """
+
+    def __init__(self, lowlim: float | None, highlim: float | None):
+        self.lowlim = lowlim
+        self.highlim = highlim
+
+    def __call__(
+        self, data: np.ndarray, ref_data: np.ndarray, weights: np.ndarray | None
+    ) -> tuple[np.ndarray, np.ndarray, np.ndarray | None]:
```
magnusuMET:
```suggestion
    ) -> np.ndarray:
```

pyaerocom/stats/data_filters.py
```diff
@@ -0,0 +1,34 @@
...
+
+
+class FilterNaN:
+    """
+    Excludes data for which either data or ref_data is NaN.
+    """
+
+    def __call__(
+        self, data: np.ndarray, ref_data: np.ndarray, weights: np.ndarray | None
+    ) -> tuple[np.ndarray, np.ndarray, np.ndarray | None]:
```
magnusuMET:
```suggestion
    ) -> np.ndarray:
```

pyaerocom/stats/data_filters.py
thorbjoernl: @lewisblake Added tests

pyaerocom/stats/data_filters.py
thorbjoernl: Should be fixed

dulte (approved):
Looks great :)

lewisblake (approved):

Issue #1085, opened 2024-04-02
Implement ColocationSetup, ObsCollection, and ModelCollection classes with Pydantic


lewisblake:
**Is your feature request related to a problem? Please describe.**

`EvalSetup` is being implemented with pydantic in #1017. However, these three attributes of `EvalSetup` are not based on Pydantic's `BaseModel` and therefore are not subject to its parsing and validation. 

**Describe the solution you would like to see**

I would like to see the classes `ObsCollection`, `ModelCollection`, and `ColocationSetup` be based on pydantic's `BaseModel` so that everything in the config interface `EvalSetup` is based on pydantic and gets its validation and parsing.

 - Would you be able to work on this solution yourself?
 Yes, perhaps. Or someone else could pick this up.
 - How can the Pyaerocom development team assist you?
 Dicussions on the structure of these classes. Whether they should be split into classes which hold data and classes which manipulate data. 

**Additional context**
Extends the work in #1017. Will likely need to rewrite some of the workarounds in that PR, which should simplify `EvalSetup`. 

Target classes are:

https://github.com/metno/pyaerocom/blob/364120b5dd6487e909738c068fc3ea3458e015f3/pyaerocom/colocation_auto.py#L41

https://github.com/metno/pyaerocom/blob/364120b5dd6487e909738c068fc3ea3458e015f3/pyaerocom/aeroval/collections.py#L80

https://github.com/metno/pyaerocom/blob/364120b5dd6487e909738c068fc3ea3458e015f3/pyaerocom/aeroval/collections.py#L173

NB:  Arguably `ObsCollection` and `ModelCollection` could be dealt with in one PR where as `ColocationSetup` could be dealt which in another.

Issue #857, opened 2023-04-27
Extend Pyaerocom to read colocated files


lewisblake:
We need to be able to read collocated files.

1. Define interface for accepted format
2. Extend pyaerocom-preproc to test this format
3. Modify pyaerocom workflow to process these data (similar to reading straight from cache)

lewisblake:
- @jgriesfeller needs this for the pyaerocom-plotting project
- says there is likely a collocated reader file already, matter of getting this to be read earlier in the pipeline
- starts as a specification problem. will need to read in collocated data objects created outside of pyaerocom, so we need to define exactly the specifications of these netCDF files. can start with a ncdump -h -ct on a collocated file to share with partners

jgriesfeller:
Since I just started reading the files for pyaerocom_plotting:
Reading the files is easy:
```Python
from pyaerocom.colocateddata import ColocatedData
file = "/home/jang/data/aeroval-local-web/coldata/c3s//LTS_MODIS/LTS.SU/od550aer_od550aer_MOD-LTS.SU_REF-AeronetSunV3L2_20030101_20181231_daily_ALL-wMOUNTAINS.nc"
bla = ColocatedData(file)
```

```
>>> type(bla)
<class 'pyaerocom.colocateddata.ColocatedData'>
```

Getting the workflow right is the bigger problem.
Note that the file above was written by aeroval

lewisblake:
```console
(pya) [lewisb@ppi-r8login-b1 IFS-Reanalysis-GHG]$ ncdump -h vmrco2_vmrco2_MOD-IFS-Reanalysis-GHG_REF-ICOS_20180101_20211231_daily_ALL-wMOUNTAINS.nc
netcdf vmrco2_vmrco2_MOD-IFS-Reanalysis-GHG_REF-ICOS_20180101_20211231_daily_ALL-wMOUNTAINS {
dimensions:
        data_source = 2 ;
        time = 1461 ;
        station_name = 35 ;
variables:
        string data_source(data_source) ;
        int64 time(time) ;
                time:units = "days since 2018-01-01 12:00:00" ;
                time:calendar = "proleptic_gregorian" ;
        string station_name(station_name) ;
        double latitude(station_name) ;
                latitude:_FillValue = NaN ;
                latitude:standard_name = "latitude" ;
                latitude:units = "degrees" ;
        double longitude(station_name) ;
                longitude:_FillValue = NaN ;
                longitude:standard_name = "longitude" ;
[...]

PR #1120, opened 2024-04-14
Pydantic ColocationSetup

lewisblake:
## Change Summary

This PR does quite a bit. Changes include:

- Implement `ColocationSetup` in terms of Pydantic. Move this class into it's own module: `colocation_setup.py`
- `Colocator` is no longer a child class of `ColocationSetup`. To get an instance of this class you must provide and instance of `ColocationSetup` or a dictionary. The option to pass a dictionary comes with a deprecation warning that it will be removed in future versions. Many tests have been changed to reflect this preferred method of instantiating a `Colocator`, where first one declares an instance of `ColocationSetup` and then passes this to `Colocator`. Dependency injection was preferred in this case to class inheritance in favor or simplifying the code between classes which contain data and classes which manipulate data.
- Many data types in `ColocationSetup` have been defined more rigorously based on the expected behavior in the code. For example, `obs_vars` is type-hinted to be a tuple (of strings) to indicate to programmers that it is immutable. `start` and `stop` if provided as strings will be coerced into pandas Timestamps. Many of these changes have been changed in config files.
- `EvalSetup` now has it's `colocation_opts` created using same method that other pydantic-based attributes ge filled. This is an area for future improvement once the config file structure is decided.
- `get_colocator` has been hacked in a way to get this work, and probably could be made better. That said. this function goes against how I think a simpler version of pyaeroval would work. So some design should take place so that it isn't needed.
- Tests which tested that the old style of using the `Colocator` and `ColocationSetup` have been removed, because this behavior is no longer supported nor encouraged.

It is worth noting that `ColocationSetup` ideally would be immutable. I have left comments in the Pydantic `ConfigDict` of this class to keep it in mind that this should be a future goal. Once you define a collocation setup at the beginning of your experiment, that should be it. And additional data that get discovered along the way should be placed somewhere else. Also I would like `validate_assignment=True` to be enabled, but it is currently commented out as it breaks the CAMS2_83 CLI, and so more work would be needed to make this change and probably some changes on the website.

## Related issue number
Part of #1085. Works towards defining the interface for a `ColocatedData` object which we will need data validation on (#857).

## Checklist

* [x] Start with a draft-PR
* [x] Tests pass locally
* [x] Tests pass on CI
* [x] Make PR ready to review


pyaerocom/aeroval/setupclasses.py
```diff
@@ -33,7 +33,9 @@
     check_if_year,
 )
 from pyaerocom.aeroval.json_utils import read_json, set_float_serialization_precision, write_json
-from pyaerocom.colocation_auto import ColocationSetup
+
+# from pyaerocom.colocation_auto import ColocationSetup # Old
+from pyaerocom.colocation_setup import ColocationSetup  # New
```
heikoklein: confusing comments  here

pyaerocom/aeroval/setupclasses.py
```diff
@@ -448,9 +443,35 @@ def colocation_opts(self) -> ColocationSetup:
...
-    @field_serializer("colocation_opts")
-    def serialize_colocation_opts(self, colocation_opts: ColocationSetup):
-        return colocation_opts.json_repr()
+    ##################################
+    ## Non-BaseModel-based attributes
+    ##################################
+
+    # These attributes require special attention b/c they're not based on Pydantic's BaseModel class.
+
+    # TODO: Use Pydantic for ColocationSetup
```
heikoklein: Isn't this done by this PR?

tests/fixtures/mscw_ctm.py
```diff
@@ -56,16 +56,36 @@ def create_fake_MSCWCtm_data(year="2019", numval=1, tst=None):
...
+#     units = "ug m-3"
+#     ds = xr.Dataset()
+
+#     ds[var_name] = data
+#     ds[var_name].attrs.update(units=units, var_name=var_name)
+
+#     ds.to_netcdf(path=path / "Base_month.nc")
+
+#     return str(path)
+
```
heikoklein: remove?

heikoklein (changes requested):
Looks good. Minimal rather internal api-changes, i.e. `col.model_id` -> `col.colocation_setup.model_id` and `eval_setup.colocation_opts["start"]` to `eval_setup.colocation_opts.start`. Please document these.

Some no longer needed comments are still in the code and should be removed.

pyaerocom/aeroval/setupclasses.py
lewisblake: Yes will remove dead code

pyaerocom/aeroval/setupclasses.py
lewisblake: From development, will remove

lewisblake:
> Looks good. Minimal rather internal api-changes, i.e. `col.model_id` -> `col.colocation_setup.model_id` and `eval_setup.colocation_opts["start"]` to `eval_setup.colocation_opts.start`. Please document these.
> 
> Some no longer needed comments are still in the code and should be removed.

Thanks. I just checked and (1) still works, (2) only occurs in private methods. Should we still document this given this class is based on pydantic and pydantic doesn't allow this functionality?

heikoklein:
If (2) only occurs in private methods, we don't need to document or implement the no-longer existing dict-like behaviour.

heikoklein (approved):

PR #1121, opened 2024-04-15
start work for map engine improvements

charlienegri:
## Change Summary

improvements for the map engine

## Checklist

* [x] Start with a draft-PR
* [x] Tests pass locally
* [x] Tests pass on CI
* [ ] Make PR ready to review

### DIFF
--- a/pyaerocom/aeroval/modelmaps_engine.py
+++ b/pyaerocom/aeroval/modelmaps_engine.py
@@ -12,6 +12,7 @@ from pyaerocom.exceptions import (
     DataCoverageError,
     DataDimensionError,
     DataQueryError,
+    ModelVarNotAvailable,
     TemporalResolutionError,
     VariableDefinitionError,
     VarNotAvailableError,
@@ -88,9 +89,8 @@ class ModelMapsEngine(ProcessingEngine, DataImporter):
                 _files = self._process_map_var(model_name, var, self.reanalyse_existing)
                 files.extend(_files)
 
-            except TypeError as ex:
+            except ModelVarNotAvailable as ex:
                 logger.warning(f"{ex}")
-                return files
             except (
                 TemporalResolutionError,
                 DataCoverageError,
@@ -131,13 +131,16 @@ class ModelMapsEngine(ProcessingEngine, DataImporter):
         AttributeError
             If the data has the incorrect number of dimensions or misses either
             of time, latitude or longitude dimension.
+        ModelVarNotAvailable
+            If model/var data cannot be read
         """
 
         try:
             data = self.read_model_data(model_name, var)
         except Exception as e:
-            logger.warning(f"{e}")
-            return
+            raise ModelVarNotAvailable(
+                f"Cannot read data for model {model_name} (variable {var}): {e}"
+            )
 
         check_var_ranges_avail(data, var)
 
--- a/pyaerocom/aeroval/modelmaps_engine.py
+++ b/pyaerocom/aeroval/modelmaps_engine.py
@@ -5,7 +5,7 @@ from pyaerocom import GriddedData, TsType
 from pyaerocom.aeroval._processing_base import DataImporter, ProcessingEngine
 from pyaerocom.aeroval.glob_defaults import var_ranges_defaults
 from pyaerocom.aeroval.helpers import check_var_ranges_avail
-from pyaerocom.aeroval.json_utils import write_json
+from pyaerocom.aeroval.json_utils import round_floats, write_json
 from pyaerocom.aeroval.modelmaps_helpers import calc_contour_json, griddeddata_to_jsondict
 from pyaerocom.aeroval.varinfo_web import VarinfoWeb
 from pyaerocom.exceptions import (
@@ -126,7 +126,13 @@ class ModelMapsEngine(ProcessingEngine, DataImporter):
             If the data has the incorrect number of dimensions or misses either
             of time, latitude or longitude dimension.
         """
-        data = self.read_model_data(model_name, var)
+        try:
+            data = self.read_model_data(model_name, var)
+        except Exception as e:
+            logger.warning(f"{e}")
+
+        data = round_floats(data)
+
         check_var_ranges_avail(data, var)
 
         if var in var_ranges_defaults:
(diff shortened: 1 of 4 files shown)


charlienegri:
the reason for the map run failing when one of the models for cams2_83 is missing is that currently there is no try/except on reading the model data as `data = self.read_model_data(model_name, var)`, so a first minimal improvement is to add that..  
also round_floats=False but am not sure where this is best called here  
ideally it would be cool if the model data could be cached in the standard run because we are reading it twice otherwise I fear

heikoklein:
Do we need round_floats for maps at all? Can you ask @AugustinMortier ?

If round_floats is needed, please check if it is also needed for the contourjson which is calculated after data and I'm not sure how contour-values are written.
grid_lons and lats are also floats and need to be rounded.

round_floats currently does not accept numpy arrays:
```python
>>> x.dtype
dtype('float64')
>>> type(x)
<class 'numpy.ndarray'>
>>> isinstance(x, np.float64)
False
>>> isinstance(x[0], np.float64)
True
```
so you will need to adapt the signature and add a new instance if you want to modify a complete nd-array, too. round_floats was up to now only used for json-data.

heikoklein:
I double-checked what I wrote and some things were not true.

- geojson-contour has already a build-in rounding of 5 digits, no further rounding needed here.
- lat_res_deg=lat_res is already the resolution after regridding, so that doesn't need a float_rounds
  -  the data itself is regridded, i.e. not what you have rounded previously.
  - you might need to apply rounding to https://github.com/metno/pyaerocom/blob/4781b831027e92166d06c7b5e1ca9fc44f635017/pyaerocom/aeroval/modelmaps_helpers.py#L71
  - since we apply the 5 digits for maps anyway in geojson, why not hardcode it here, too, instead of using round_floats?

heikoklein (approved):

charlienegri:
I had to do some more work to get the desired behavior for cams2_83, I tested the code now by merging it to the cams2-83-experiments-2 branch and it works for me, when a model has missing data it's skipped, @heikoklein can you take another look?

pyaerocom/aeroval/modelmaps_engine.py
```diff
@@ -85,6 +88,9 @@ def _run_model(self, model_name: str, var_list):
                 _files = self._process_map_var(model_name, var, self.reanalyse_existing)
                 files.extend(_files)
 
+            except TypeError as ex:
+                logger.warning(f"{ex}")
+                return files
```
heikoklein: The warning could be a bit more explicit, i.e. include {model_name} and {var}, e.g. `f"Exception during processing of  model for {model_name} ({var}): {ex}"` (You can't be sure that the logger.info in line 85 is written, depending on logging configuration.)

I think `return files` is wrong here. Not sure why you get the TypeError, but I guess you want to extend the other `var` in `var_list`. The real return is in line 103.

pyaerocom/aeroval/modelmaps_engine.py
```diff
@@ -126,10 +132,16 @@ def _process_map_var(self, model_name, var, reanalyse_existing):
             If the data has the incorrect number of dimensions or misses either
             of time, latitude or longitude dimension.
         """
-        data = self.read_model_data(model_name, var)
+
+        try:
+            data = self.read_model_data(model_name, var)
+        except Exception as e:
+            logger.warning(f"{e}")
```
heikoklein: Longer message: `f"Skipping due to exception during reading of model {model_name} ({var}): {e}"`

And very strange: why do you catch the Exception here, and then you catch most exceptions in line 94. I would solve that with a special exception, i.e. `raise ModelUnavailableException` which could be caught in line 94. Exceptions are nothing else than special return statements, much better than a dummy, unspecified `return None` in line 140. (Is None what causes the TypeError in line 91?)

pyaerocom/aeroval/modelmaps_helpers.py
```diff
@@ -5,6 +5,7 @@
 from cartopy.mpl.geoaxes import GeoAxes
 from matplotlib.axes import Axes
 from matplotlib.colors import ListedColormap, to_hex
+from numpy import round
```
heikoklein: Slightly confusing import since `np.round` overwrites here the built-in `round`.

heikoklein (commented):
Looks good. Just added some suggestions.

pyaerocom/aeroval/modelmaps_engine.py
heikoklein: https://github.com/metno/pyaerocom/blob/main-dev/pyaerocom/exceptions.py has already a VariableNotFoundError, which might be applicable here? Or you define a new exception.

pyaerocom/aeroval/modelmaps_helpers.py
heikoklein: cleaner: 
```python3
import numpy as np
...
np.round(...)
```

pyaerocom/aeroval/modelmaps_engine.py
charlienegri: true, I want a continue, in reality it does not matter tho....  
I'll chat you in person

tests/aeroval/test_modelmaps_engine.py
```diff
@@ -1,18 +1,19 @@
...
-    engine._process_map_var("LOTOS", "concco", False)
-    print(caplog.text)
-    assert "no such entry LOTOS" in caplog.text
+    try:
+        engine._process_map_var("LOTOS", "concco", False)
+    except ModelVarNotAvailable as e:
+        # print(e)
+        assert "no such entry LOTOS" in str(e)
+    else:
+        assert False
```
heikoklein: following the pytest documentation https://docs.pytest.org/en/7.1.x/how-to/assert.html#assertions-about-expected-exceptions, the try/except part could be written like
```python
import pytest
...
with pytest.raises(ModelVarNotAvailable) as excinfo:
    engine._process_map_var("LOTOS", "concco", False)
assert "no such entry LOTOS" in str(excinfo.value)
```

heikoklein (approved):

PR #1128, opened 2024-04-18
Merge CAMS2-83 code to main-dev

### LINKED ISSUE #1113: Merging cams2-83-experiments-2 to main-dev
CAMS2_83 code has lots of additional code, all gathered in branch https://github.com/metno/pyaerocom/tree/cams2-83-experiments-2

For easier maintenance, the cams2--83-experiments should as far as possible be merged to main-dev, so that tests in main-dev catch up errors early during development.

heikoklein:
## Change Summary

Merging cams-2-83-experiment-2 to main-dav

## Related issue number

#1113 

## Checklist

* [x] Start with a draft-PR
* [x] Tests pass locally
* [x] Tests pass on CI
* [x] Make PR ready to review


heikoklein:
Considering the size of this PR, I would like to have 2 reviewers here.

heikoklein:
Ignoring codecov since this is special production code tested outside pyaerocom.

heikoklein:
All tests pass on PPI, too.

heikoklein:
@charlienegri Can we remove the QUIET option? It does 2 things: suppressing tqdm and modify logging. 

- tqdm can be autosuppressed with disable=None when run non-interactive
- logging should be configured by logging.ini

pyaerocom/aeroval/coldatatojson_helpers.py
```diff
@@ -866,6 +867,7 @@ def _process_map_and_scat(
                         #  Code for the calculation of trends
                         if add_trends and freq != "daily":
                             (start, stop) = _get_min_max_year_periods([per])
+                            # (start, stop) = _start_stop_from_periods(per)
```
lewisblake: dead code

pyaerocom/aeroval/coldatatojson_helpers.py
```diff
@@ -1148,6 +1150,7 @@ def _process_heatmap_data(
                             if add_trends and freq != "daily":
                                 # Calculates the start and stop years. min_yrs have a test value of 7 years. Should be set in cfg
                                 (start, stop) = _get_min_max_year_periods([per])
+                                # (start, stop) = _start_stop_from_periods(per)
```
lewisblake: please remove dead code

pyaerocom/aeroval/coldatatojson_helpers.py
```diff
@@ -1371,9 +1374,18 @@ def _init_data_default_frequencies(coldata, to_ts_types):
     return data_arrs
 
 
-def _start_stop_from_periods(periods):
-    start, stop = _get_min_max_year_periods(periods)
-    return start_stop(start, stop + 1)
+# def _start_stop_from_periods(period):
```
lewisblake: please remove dead code

pyaerocom/aeroval/coldatatojson_helpers.py
```diff
@@ -46,7 +46,8 @@ def get_stationfile_name(station_name, obs_name, var_name_web, vert_code):
 
 def get_json_mapname(obs_name, var_name_web, model_name, model_var, vert_code, period):
     """Get name base name of json file"""
-    return f"{obs_name}-{var_name_web}_{vert_code}_{model_name}-{model_var}_{period}.json"
+    periodmod = period.replace("/", "")
```
lewisblake, line 49: please add a comment about what periodmod means and why this is being applied. Something about edge cases from cams2-83

pyaerocom/aeroval/helpers.py
```diff
@@ -75,13 +84,22 @@ def _check_statistics_periods(periods: list) -> list:
...
+            if len(spl[0]) != len(spl[1]):
+                raise ValueError(f"{spl[0]} not on the same format as {spl[1]}")
+
         if len(spl) > 2:
             raise ValueError(
                 f"Invalid value for period ({per}), can be either single "
                 f"years or period of years (e.g. 2000-2010)."
             )
-        _per = "-".join([str(int(val)) for val in spl])
+        years = True if len(spl[0]) == 4 else False
```
lewisblake, line 96: please add some comments about why this additional processing is here and what it is doing. Maybe there should be a uniform comment througtout the code `# cams2-83 requriement` so we don't forget later what some of this more cryptic code is doing and why we have it

pyaerocom/aeroval/setupclasses.py
```diff
@@ -59,6 +65,7 @@ class OutputPaths(BaseModel):
         "hm",
         "hm/ts",
         "contour",
+        "forecast",
```
lewisblake: we should make @AugustinMortier aware that this will be produced by other experiments than cams2-83 in case there is some special processing that occurs on the website

pyaerocom/aeroval/setupclasses.py
```diff
@@ -362,6 +375,16 @@ def modelmaps_opts(self) -> ModelMapsSetup:
         }
         return ModelMapsSetup(**model_args)
 
+    @computed_field
+    @property
```
lewisblake: I suspect that this may need to be a `cached_property` to prevent the debugging issue we had previously that cams2-83 highlighted

pyaerocom/aeroval/setupclasses.py
```diff
@@ -362,6 +375,16 @@ def modelmaps_opts(self) -> ModelMapsSetup:
...
+    def cams2_83_cfg(self) -> CAMS2_83Setup:
+        if not hasattr(self, "model_extra"):
+            return CAMS2_83Setup()
+        model_args = {
```
lewisblake, line 386: Could potentially save some minor compute time by applying the same logic used in `ProjectInfo` as the `CAMS2_83Setup` class is a special case with only one attribute. Could keep this here if it is foreseeable that this class will we built upon in the future.

pyaerocom/io/readungridded.py
```diff
@@ -486,8 +488,12 @@ def read_dataset(
         UngriddedData
             data object
         """
+        force_caching = False
```
lewisblake, line 491: I assume this will be set in a config. Has this been tested with the new `EvalSetup`? Should we add an attribute to the `CAMS2_83Setup` for this?

pyaerocom/scripts/cams2_83/cli.py
```diff
@@ -0,0 +1,341 @@
...
+    - Add option for running only som observations/models/species
+    - Add options with defaults for the different folders (data/coldata/cache)
+"""
+
+
+app = typer.Typer(add_completion=False, no_args_is_help=True)
+logger = logging.getLogger(__name__)
+
+
+def clear_cache():
```
lewisblake, line 41: Why can't we use the regular CLI's `clear_cache`? Seems like it would be cleaner then importing `list_cache_files` and then implementing the same functionality

pyaerocom/scripts/cams2_83/config.py
```diff
@@ -0,0 +1,209 @@
...
+    # cams2_83_dateshift=0,
+    # Regional filter for analysis
+    filter_name="ALL-wMOUNTAINS",
+    # colocation frequency (no statistics in higher resolution can be computed)
+    ts_type="hourly",
+    # The size of map used to display the results
+    map_zoom="Europe",
+    # Options for time
+    freqs=["hourly", "daily"],  # Possible frequencies
+    periods=[
```
lewisblake, line 42: is this still true?

pyaerocom/scripts/cams2_83/config.py
```diff
@@ -0,0 +1,209 @@
...
+        daily=dict(hourly=18),
+    ),
+)
+
+
+##################################################
+#        The model configs
+##################################################
+# folder_CAMS2_83 = f'/home/danielh/lustre/storeB/project/fou/kl/CAMS2_83/test_data'
+folder_CAMS2_83 = "/lustre/storeB/project/fou/kl/CAMS2_83/model"
```
lewisblake: Should this be a `Path`

pyaerocom/scripts/cams2_83/config.py
```diff
@@ -0,0 +1,209 @@
...
+
+
+MODELS_CONFIG = {
+    # "CAMS2-83": dict(
```
lewisblake: remove dead code?

pyaerocom/scripts/cams2_83/engine.py
```diff
@@ -0,0 +1,215 @@
...
+            obs_var = model_var = "UNDEFINED"
+
+        model = ModelName[coldata[0].model_name.split("-")[2]]
+        vert_code = coldata[0].get_meta_item("vert_code")
+        obs_name = coldata[0].obs_name
+        mcfg = self.cfg.model_cfg.get_entry(model.name)
+        var_name_web = mcfg.get_varname_web(model_var, obs_var)
+        seasons = self.cfg.time_cfg.get_seasons()
+
+        # for i in range(1, forecast_days):
```
lewisblake: remove dead code

pyaerocom/scripts/cams2_83/engine.py
```diff
@@ -0,0 +1,215 @@
...
+
+        regions_how = "country"
+        use_country = True
+        for i in range(forecast_days):
+            coldata[i].data["season"] = coldata[i].data.time.dt.season
+            (regborders, regs, regnames) = init_regions_web(coldata[i], regions_how)
+
+        results = {}
+
+        # seasons = [s for s in seasons if s in coldata[0].data["season"].data] + ["all"]
```
lewisblake: dead code

pyaerocom/scripts/cams2_83/engine.py
lewisblake, line 1: Generally methods in this file could benefit from more documentation in the form of docstrings

pyaerocom/scripts/cams2_83/evaluation.py
```diff
@@ -0,0 +1,147 @@
...
+    else:
+        if start_period == dates[-1]:
+            periods.append(f"{start_period:%Y%m%d}")
+        else:
+            periods.append(f"{start_period:%Y%m%d}-{dates[-1]:%Y%m%d}")
+
+    return periods
+
+
+def years_starting_in_november(start_date: date, end_date: date) -> list[str]:
```
lewisblake: this could benefit from some documentation

tests/fixtures/cams2_83/config.py
```diff
@@ -0,0 +1,50 @@
+from __future__ import annotations
+
+from pathlib import Path
+
+import pytest
+
+from . import cfg_test
+
+# @pytest.fixture()
```
lewisblake: dead code

tests/fixtures/cams2_83/config.py
```diff
@@ -0,0 +1,50 @@
...
+    monkeypatch.setattr(
+        "pyaerocom.io.cachehandler_ungridded.CacheHandlerUngridded.cache_dir", tmp_path
+    )
+    cache_file = tmp_path / "tmp.pkl"
+    cache_file.write_bytes(b"")
+    assert cache_file.exists()
+    return tmp_path
+
+
+# @pytest.fixture()
```
lewisblake: dead code

lewisblake (changes requested):
Impressive work getting this PR so far. I know it's taken a lot. Feel free to take my comments in stride. This is a very large PR and introduces a lot of changes that are specific to CAMS2_83 and so generally I think an effort to delineate parts of the code that are specific to cams2-83 would benefit future programmers reading the code. The tests I understand we are not going to worry about for now, but one day I think we should revisit them.

pyaerocom/io/readungridded.py
charlienegri: in the config we have 
```
OBS_CONFIG["EEA"] = dict(
    obs_id="CAMS2_83.NRT",
    # obs_id="EEAAQeRep.NRT",
    obs_vars=species_list,
    web_interface_name="EEA-NRT",
    obs_vert_type="Surface",
    read_opts_ungridded=dict(files=[], force_caching=True),
    obs_filters=obs_filters,
)
``` 
and since the latest changes in `EvalSetup` I think we have a bug :  
the .pkl files with the obs are correctly written when the standard run starts, but then when I run the medianscores the cache appears to be empty even though it is explicitly not cleared, so now that you mention this, we might indeed be missing something but I do not understand exactly what's going on yet

pyaerocom/scripts/cams2_83/cli.py
charlienegri: in  `pyaerocom/scripts/cli.py` `clearcache` is a command 
```
@main.command()
def clearcache():
    """Delete cached data objects"""

    delete = typer.confirm("Are you sure you want to delete all cached data objects?")
    if delete:
        print("Okay then.... Here we go!")
        for path in list_cache_files():
            path.unlink()
    else:
        print("Wise decision, pyaerocom will handle it for you automatically anyways 😜")
```
how would one use that in `pyaerocom/scripts/cams2_83/cli.py`?

pyaerocom/scripts/cams2_83/config.py
charlienegri: which field are you referring to? the `freqs` and `periods` ones are modified by the cli

pyaerocom/aeroval/setupclasses.py
charlienegri: testing did not highlight a visible issue here so far I think  but maybe we are indeed doing a lot of extra work that a `cached_property` would spare, I do not see why not change it

pyaerocom/aeroval/setupclasses.py
charlienegri: t̶h̶i̶s̶ ̶c̶o̶u̶l̶d̶ ̶p̶e̶r̶h̶a̶p̶s̶ ̶b̶e̶ ̶a̶ ̶c̶a̶s̶e̶-̶b̶a̶s̶e̶d̶ ̶d̶e̶f̶i̶n̶i̶t̶i̶o̶n̶,̶ ̶d̶i̶f̶f̶e̶r̶e̶n̶t̶ ̶i̶f̶ ̶`̶u̶s̶e̶_̶c̶a̶m̶s̶2̶_̶8̶3̶`̶ ̶  I think it's not a problem to have an extra empty folder in other experiments, perhaps the name should be medianscores tho or something less generic

... or then we keep `JSON_SUBDIRS` as it is and we add something after the loop on JSON_SUBDIRS has ended,  like: 
```
        for subdir in self.JSON_SUBDIRS:
            loc = self._check_init_dir(os.path.join(base, subdir), assert_exists)
            out[subdir] = loc
        if self.proj_id == "cams2-83":
            loc = self._check_init_dir(os.path.join(base, "forecast"), assert_exists)
            out["forecast"] = loc
```
means `proj_id` for cams2-83 is hard coded here and cannot be changed, which is not great, but at the moment I do not see other easy solutions... `use_cams2_83` from the config would be way better but I do not know how to access it from here

pyaerocom/aeroval/coldatatojson_helpers.py
charlienegri: I am actually not sure why this was introduced, we could also have had filenames written with slashes for cams2-83.. anyway they are expected without now on aeroval so I'll write a comment

pyaerocom/scripts/cams2_83/config.py
charlienegri: this is dead code too: the model path is defined as cl option
```
model_path: Path = typer.Option(
        DEFAULT_MODEL_PATH, exists=True, readable=True, help="path to model data"
    ),
```
and `DEFAULT_MODEL_PATH` is from `./pyaerocom/io/cams2_83/reader.py` where it's defined as `DATA_FOLDER_PATH = Path("/lustre/storeB/project/fou/kl/CAMS2_83/model")`

pyaerocom/io/readungridded.py
charlienegri: I checked and the observations are cached as they should, I have looked into the way the new code from main-dev handles cache and I conclude that the new RM_CACHE_OUTDATED env variable, True by default, is what is clearing the cache before the median scores run despite the cams2_83 cli being set to skip the clearing in that case

so forget what I wrote above, it's not a bug, I just need to set that variable to false for cams2_83

pyaerocom/scripts/cams2_83/evaluation.py
charlienegri: so... I am not sure I get why this is necessary at all, I have been looking into what the periods are in different kind of evaluations and:
- if the type of evaluation is "long" like we have for the multi-seasonal runs done every 3 months, the periods are just a year range + single years, the function that determines them is `make_period_ys` e.g. with `start_date = datetime(2021,6,1)` and `end_date = datetime(2024,2,29)` the periods will be `['2021-2024', '2021', '2022', '2023', '2024']`, so this case is pretty easy
- for the rest of the evaluations the periods list has basically just a single item which is the date-range in most cases, for example with `start_date = datetime(2023,12,28)` and `end_date = datetime(2024,1,12)`, `make_period(start_date,end_date) = ['20231228-20240112']`    
 unless it's an across-season range and then the function `years_starting_in_november` comes to play,  e.g. with `start_date = datetime(2024,5,28)` and `end_date = datetime(2024,6,12)` we have `make_period(start_date,end_date) = ['20240528-20240612', '20240528-20241130']`   
  because in the function make_period we have the case 
  ```
    periods = [f"{start_date:%Y%m%d}-{end_date:%Y%m%d}"]
    if periods != season_periods:
        periods += nov_periods  # season_periods
  ``` 
  and here  
  `seasons_in_period(start_date, end_date) = ['20240528-20240531', '20240601-20240612']`  
  `years_starting_in_november(start_date, end_date) = ['20240528-20241130']`  
  
  .....right now I do not get why one would want `['20240528-20240612', '20240528-20241130']` as periods, seems just wrong, but maybe I am not aware of something on the web server side

charlienegri: my guess is that it might have to do with the season selection in the drop down menu

charlienegri: anyway, I am not sure where and how it would be best to document all this

pyaerocom/aeroval/coldatatojson_helpers.py
charlienegri: I am not sure it's even relevant anymore now that we have switched to `_get_min_max_year_periods`, I will test and get rid of it if so

charlienegri: actually I am not sure about what's going on here because if I remove the replace, the code crashes with FileNotFoundError, like 
```
FileNotFoundError: [Errno 2] No such file or directory: 
'/lustre/storeB/users/charlien/cams2-83-tests/data/cams2-83/test-forecast-day/ma
p/EEA-NRT-concno2_Surface_ENSEMBLE-concno2_2024/03/01.json'
```
but I was assuming that `get_json_mapname` is just where the output filename is established since in coldatatojson_engine.py we have
```
            map_name = get_json_mapname(
                obs_name, var_name_web, model_name, model_var, vert_code, period
            )
            outfile_map = os.path.join(out_dirs["map"], map_name)
            write_json(map_data, outfile_map, ignore_nan=True)
``` 
but apparently not? I am not sure I follow why it crashes without the replace..
also I do not know when exactly do the periods as strings acquire slashes, they are not written with slashes in the config, must be some to-string conversion in ColocatedData object....

pyaerocom/scripts/cams2_83/evaluation.py
charlienegri: update: I ran a week across seasons and this is how the second period, the one added by `periods += nov_periods`,  looks like 
Screenshot from 2024-04-22 14-18-15

I have no idea what's the point of this

pyaerocom/aeroval/coldatatojson_helpers.py
charlienegri: the slash comes from aeroval/helpers.py
```
        if years:
            _per = "-".join([str(int(val)) for val in spl])
        else:
            _per = "-".join([to_pandas_timestamp(val).strftime("%Y/%m/%d") for val in spl])
```
I suspect that if I get rid of it here I can skip the periodmod thing, I will test

charlienegri: does not seem to be the case, so periodmod is here to stay

charlienegri:
my latest commit breaks the webserver, see test-forecast-day here https://aeroval-test.met.no/charlien/pages/evaluation/?project=cams2-83&experiment=test-forecast-day&station=ALL&parameter=concpm10
basically, the slash in the period strings matters for many parts of the web visualization, so I will revert my last changes and leave it as it was, we can still remove the periodmod line and get filenames written with slashes, but then we will need to tell @AugustinMortier to adapt to this once/if we will deploy this code to production (filenames were written with slashes initially but not anymore now)

pyaerocom/aeroval/helpers.py
```diff
@@ -128,17 +148,42 @@ def _get_min_max_year_periods(statistics_periods):
...
         single year analysis).
     """
-    startyr, stopyr = 1e6, -1e6
+    # startyr, stopyr = 1e16, -1e16
+    startyr, stopyr = start_stop("2100", "1900")
     for per in statistics_periods:
         sl = _period_str_to_timeslice(per)
-        perstart, perstop = int(sl.start), int(sl.stop)
+        # perstart, perstop = sl.start, sl.stop
+        perstart, perstop = start_stop(sl.start, sl.stop)
```
charlienegri, line 156: I suspect here we can use the same code from main-dev, but it's hard to test without implementing the changes, so I might just do it

lewisblake (approved):
:star2: :unicorn: :sparkles: :rocket:

PR #1134, opened 2024-04-22
Pydantic ColocatedData

lewisblake:
## Change Summary

The `ColocatedData` class is now a derived class of pydantic's `BaseModel`. Doing so ensures data validation and defines an interface, part of which we will need to guarantee successful reading in colocated data objects created outside of pyaerocom, for example, in CAMS2_82.

## Related issue number

Part of #857. Related work in #1120 deriving `ColocationSetup` from `BaseModel`

## Checklist

* [x] Start with a draft-PR
* [x] Tests pass locally
* [x] Tests pass on CI
* [x] Make PR ready to review

### DIFF
--- a/pyaerocom/colocateddata.py
+++ b/pyaerocom/colocateddata.py
@@ -4,11 +4,13 @@ import logging
 import os
 import warnings
 from ast import literal_eval
+from functools import cached_property
 from pathlib import Path
 
 import numpy as np
 import pandas as pd
-import xarray
+import xarray as xr
+from pydantic import BaseModel, ConfigDict, computed_field, field_validator, model_validator
 
 from pyaerocom import const
 from pyaerocom.exceptions import (
@@ -33,7 +35,7 @@ from pyaerocom.time_resampler import TimeResampler
 logger = logging.getLogger(__name__)
 
 
-class ColocatedData:
+class ColocatedData(BaseModel):
     """Class representing colocated and unified data from two sources
 
     Sources may be instances of :class:`UngriddedData` or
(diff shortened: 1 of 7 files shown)


pyaerocom/colocateddata.py
```diff
@@ -81,55 +85,60 @@ class ColocatedData:
...
+    ##   Pydantic-based Attributes
+    #################################
+
+    data: Path | str | xr.DataArray | np.ndarray | None = None
+
+    @model_validator(mode="after")
+    def validate_data(self):
+        # if (
+        #     self.data is not None
+        # ):  # LB: Check that this is needed. may be covered by the model_validator
```
thorbjoernl: Is this still needed?

pyaerocom/colocateddata.py
```diff
@@ -81,55 +85,60 @@ class ColocatedData:
...
-                self.open(data)
-            elif isinstance(data, xarray.DataArray):
-                self.data = data
-            elif isinstance(data, np.ndarray):
-                if not data.ndim in (3, 4):
-                    raise DataDimensionError("invalid input, need 3D or 4D numpy array")
-                elif not data.shape[0] == 2:
-                    raise DataDimensionError(
-                        "first dimension (data_source) must be of length 2(obs, model)"
-                    )
```
thorbjoernl, line 102: Shouldn't these apply to DataArrays also?

pyaerocom/colocateddata.py
```diff
@@ -81,55 +85,60 @@ class ColocatedData:
...
-        return self._data
+                extra_args_from_class_initialization = {}
+            data = xr.DataArray(self.data, **extra_args_from_class_initialization)
+            self.data = data
 
-    @data.setter
-    def data(self, val):
-        if not isinstance(val, xarray.DataArray):
-            raise ValueError("Invalid input for data attribute, need instance of xarray.DataArray")
-        self._data = val
```
thorbjoernl, line 132: I'd put the validation from the constructor in the setter, unless there is a good reason why validation shouldn't apply when using the setter directly.

pyaerocom/colocateddata.py
```diff
@@ -1878,20 +1885,20 @@ def plot_coordinates(self, marker="x", markersize=12, fontsize_base=10, **kwargs
...
-    def __repr__(self):
-        tp = type(self).__name__
-        dstr = "<empty>" if self._data is None else repr(self._data)
-        return f"pyaerocom.{tp}: data: {dstr}"
+    # def __repr__(self):
+    #     tp = type(self).__name__
+    #     dstr = "<empty>" if self.data is None else repr(self.data)
+    #     return f"pyaerocom.{tp}: data: {dstr}"
 
-    def __str__(self):
```
thorbjoernl, line 1889: Is this commented code still needed?

tests/test_colocateddata.py
```diff
@@ -23,9 +24,9 @@ def test_ColocatedData__init__(data: Path | str | ArrayLike):
 @pytest.mark.parametrize(
     "data,exception",
     [
-        (None, AttributeError),
-        ("Blaaaaa", IOError),
-        (np.ones(3), DataDimensionError),
+        # (None, AttributeError),
```
thorbjoernl: Maybe test that exceptions are raised consistently by data-types (ie. xarray and numpy array).

thorbjoernl (changes requested):

pyaerocom/colocateddata.py
lewisblake: This is the old logic, which probably does need to be applied to xarray data objects since they can be make outside of pyaerocom now.

pyaerocom/colocateddata.py
lewisblake: The validation is now handled by pydantic through type hinting, so this manual validation isn't needed anymore. Line 125 has the type hinting for what `data` can be, and if someone passes something else, it will crash.

pyaerocom/colocateddata.py
lewisblake: `BaseModel` has a `__str__` method, which this class would inheret, but I agree this overloading is probably what we want.

tests/test_colocateddata.py
lewisblake: Added a test

thorbjoernl (dismissed):
~~LGTM :rocket:~~

thorbjoernl (approved):
LGTM :robot: :rocket:

PR #1166, opened 2024-05-11
code to make the cams2_83 mos evaluation work

charlienegri:
## Change Summary

add code to make it possible to run the MOS/ENS evaluation starting from pre-made colocated data

this addresses https://github.com/metno/AeroToolsIssues/issues/1

## Checklist

* [x] Start with a draft-PR
* [x] Tests pass locally
* [x] Tests pass on CI
* [x] Make PR ready to review

### DIFF
--- a/pyaerocom/aeroval/experiment_processor.py
+++ b/pyaerocom/aeroval/experiment_processor.py
@@ -60,7 +60,7 @@ class ExperimentProcessor(ProcessingEngine, HasColocator):
                 )
             else:
                 preprocessed_coldata_dir = ocfg["coldata_dir"]
-                mask = f"{preprocessed_coldata_dir}/*/*.nc"
+                mask = f"{preprocessed_coldata_dir}/{model_name}/*.nc"
                 files_to_convert = glob.glob(mask)
                 engine = ColdataToJsonEngine(self.cfg)
                 engine.run(files_to_convert)
--- a/pyaerocom/aeroval/experiment_processor.py
+++ b/pyaerocom/aeroval/experiment_processor.py
@@ -60,7 +60,7 @@ class ExperimentProcessor(ProcessingEngine, HasColocator):
                 )
             else:
                 preprocessed_coldata_dir = ocfg["coldata_dir"]
-                mask = f"{preprocessed_coldata_dir}/*.nc"
+                mask = f"{preprocessed_coldata_dir}/*/*.nc"
                 files_to_convert = glob.glob(mask)
                 engine = ColdataToJsonEngine(self.cfg)
                 engine.run(files_to_convert)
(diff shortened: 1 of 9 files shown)


pyaerocom/scripts/cams2_83/cli.py
```diff
@@ -338,3 +408,72 @@ def main(
...
+        min=1,
+        max=cpu_count(),
+        help="CPUs for running the median scores",
+    ),
+):
+    if pool > mp.cpu_count():
+        logger.warning(
+            f"The given pool {pool} is larger than the maximum CPU count {mp.cpu_count()}. Using that instead"
+        )
+        pool = mp.cpu_count()
```
heikoklein: You can give a warning, but let the user shoot in his own foot.

pyaerocom/scripts/cams2_83/cli.py
```diff
@@ -286,7 +351,12 @@ def main(
         help="Will only make and print the config without running the evaluation",
     ),
     pool: int = typer.Option(
-        1, "--pool", "-p", min=1, max=cpu_count(), help="CPUs for reading OBS"
+        1,
+        "--pool",
+        "-p",
+        min=1,
+        max=cpu_count(),
```
heikoklein: Don't set cpu_count() as maximum if you don't know if this is right.

Some tasks are IO bound, so a maximum of 4 is often too much in this case. In theory, we could have tasks which connect to external resources, so that we might want to use much more than the cpus available. (A process based webserver setup used to be 2-4 times the number of cores.)

We don't need to add a maximum here.

pyaerocom/scripts/cams2_83/cli.py
```diff
@@ -338,3 +408,72 @@ def main(
...
+    ),
+    id: str = typer.Option(CFG["exp_id"], help="experiment ID"),
+    name: str = typer.Option(CFG["exp_name"], help="experiment name"),
+    description: str = typer.Option(CFG["exp_descr"], help="experiment description"),
+    pool: int = typer.Option(
+        1,
+        "--pool",
+        "-p",
+        min=1,
+        max=cpu_count(),
```
heikoklein: remove, see above

pyaerocom/aeroval/experiment_processor.py
```diff
@@ -122,7 +122,11 @@ def run(self, model_name=None, obs_name=None, var_list=None, update_interface=Tr
         if not self.cfg.model_cfg:
             logger.info("No model found, will make dummy model data")
             self.cfg.webdisp_opts.hide_charts = ["scatterplot"]
-            self.cfg.webdisp_opts.hide_pages = ["maps.php", "intercomp.php", "overall.php"]
+            self.cfg.webdisp_opts.hide_pages = [
+                "maps.php",
+                "intercomp.php",
+                "overall.php",
+            ]
```
heikoklein, line 129: Do we still write .php pages? I thought @AugustinMortier had removed those?

heikoklein (approved):
Looks good, just some minor comments.

pyaerocom/aeroval/experiment_processor.py
charlienegri: I have no idea but this is main-dev code, so I'd rather address that somewhere else than in this PR

pyaerocom/scripts/cams2_83/cli.py
charlienegri: how about setting it to 1 then?

charlienegri: nvm, I like your approach best

pyaerocom/aeroval/experiment_processor.py
heikoklein: ok

pyaerocom/scripts/cams2_83/cli.py
heikoklein: I can create 5000 processes without problems on my 6-core machine. Why not just let the user do what he asked for, even if it does not make sense to you?

charlienegri:
tested with a MOS/ENS evaluation run

PR #1212, opened 2024-06-18
First try at fixing col setup without start stop. Model validator for…

### LINKED ISSUE #1211: Empty colocated object when start and stop are not given
**Describe the bug**
The colocator returns an empty object when no start and stop are given in the colocation setup

**To Reproduce**
1. Make a colocation setup without start and stop
2. Run a colocator with said setup
3. The result is an empty colocated object

dulte:
… pydating is also removed

## Change Summary

Changes how the colocation setup and colocator handles start and stop times

## Related issue number
Fix #1211

## Checklist

* [x] Start with a draft-PR
* [ ] Tests pass locally
* [ ] Tests pass on CI
* [ ] Make PR ready to review

### DIFF
--- a/pyaerocom/colocation/colocation_setup.py
+++ b/pyaerocom/colocation/colocation_setup.py
@@ -463,6 +463,7 @@ class ColocationSetup(BaseModel):
         for key in self.FORBIDDEN_KEYS:
             if key in self.model_fields:
                 raise ValidationError
+
         return self
 
     @model_validator(mode="after")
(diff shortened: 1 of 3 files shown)


pyaerocom/colocation/colocation_setup.py
```diff
@@ -463,6 +463,36 @@ def validate_no_forbidden_keys(self):
...
+
+    @model_validator(mode="after")
+    # @classmethod
+    def validate_start_stop_xand(self):
+        if not (self.start and self.stop):
+            if self.start or self.stop:
+                raise ValueError("Both start and stop need to be provided or both not provided.")
+        return self
+
+    # @model_validator(mode="after")
```
lewisblake: Can we not just leave this in?

dulte: We can. But then we have to remove the `return self` above. I'm not sure if this might lead to other strange behavior

dulte:
Some test do not work due to 
```        if self.colocation_setup.model_use_climatology:
            if self.colocation_setup.stop is not None or not isinstance(
                self.colocation_setup.start, int
            ):
                raise ColocationSetupError(
                    "Conflict: only single year analyses are support for model "
                    'climatology fields, please specify "start" as integer '
                    'denoting the year, and set "stop"=None'
                )
```
part of the colocator code

lewisblake:
@dulte pushing this back until September to match the milestone for the associated PR. This is the type of issue I think we should clarify at the AeroTools backend design retreat, so please keep in mind as an example.

pyaerocom/colocation/colocation_setup.py
```diff
@@ -464,12 +464,16 @@ def validate_no_forbidden_keys(self):
             if key in self.model_fields:
                 raise ValidationError
 
-    @cached_property
-    def basedir_logfiles(self):
-        p = Path(self.basedir_coldata) / "logfiles"
-        if not p.exists():
-            p.mkdir(parents=True, exist_ok=True)
-        return str(p)
+        return self
```
avaldebe, line 467: why return `self`?

pyaerocom/colocation/colocation_setup.py
```diff
@@ -464,12 +464,16 @@ def validate_no_forbidden_keys(self):
...
+
+    @model_validator(mode="after")
+    # @classmethod
+    def validate_start_stop_xand(self):
+        if not (self.start and self.stop):
+            if self.start or self.stop:
+                raise ValueError("Both start and stop need to be provided or both not provided.")
+
+    #     return self
```
avaldebe, line 476: the previous validator returns `self, why this one does not?

pyaerocom/colocation/colocation_setup.py
```diff
@@ -464,12 +464,16 @@ def validate_no_forbidden_keys(self):
...
-        if not p.exists():
-            p.mkdir(parents=True, exist_ok=True)
-        return str(p)
+        return self
+
+    @model_validator(mode="after")
+    # @classmethod
+    def validate_start_stop_xand(self):
+        if not (self.start and self.stop):
+            if self.start or self.stop:
```
avaldebe, line 473: are the nested `if` needed?
maybe this could be re-written as
``` python
        if not (self.start and self.stop) and self.start or self.stop:
```
or 
``` python
        if bool(self.start) != bool(self.stop):
```

pyaerocom/colocation/colocation_setup.py
```diff
@@ -492,6 +496,13 @@ def validate_obs_config(cls, v: PyaroConfig):
                 cls.obs_id = v.name
         return v
 
+    @cached_property
+    def basedir_logfiles(self):
+        p = Path(self.basedir_coldata) / "logfiles"
+        if not p.exists():
+            p.mkdir(parents=True, exist_ok=True)
```
avaldebe, line 503: I see no need for the `if`, the `exist_ok=True` parameter does the checking.
You can simple write:
``` python
        p.mkdir(parents=True, exist_ok=True)
```

avaldebe (changes requested):
I see a few details that can be improved.
Also, not that `start` and `stop` need to be present I would appreciate tests when one of them is missing in order to ensure that an exception is raised.

dulte:
Will be closed for now. Correct way of handling start/stop is still not decided and will be postpone until a problem arise

PR #1220, opened 2024-06-21
Fixes for map-plotting of emep-results

heikoklein:
## Change Summary

Several fixes to get map-plotting useful for the emep-reports like

- Fixing units
- Fixing scales
- Using a default-scale 0-10 for variables not previously defined in glob_defaults.py
- removing misleading warnings

## Checklist

* [x] Start with a draft-PR
* [ ] Tests pass locally
* [x] Tests pass on CI
* [x] Make PR ready to review


heikoklein:
@AugustinMortier and @charlienegri I would like to have both your reviews, in particular with the change to using the default-range for the contours from (min-max) to (0-10).

charlienegri:
as far as I can tell your cmap edit applies only to the case when `var` is NOT in `var_ranges_defaults.keys()`, so it will not affect cams2_83 (which was the scope I had edited `modelmaps_engine.py` within), as all the relevant variables should be in `var_ranges_defaults` ... so no problems for me. but Augustin is the one best equipped to evaluate this.

charlienegri (approved):

lewisblake:
Can this be merged? @charlienegri

charlienegri:
> Can this be merged? @charlienegri

ok for me, Augustin is the only one who could object but I think the changes do not impact existing projects (which should have the variables listed in `var_ranges_defaults`)

AugustinMortier:
Because those are new variables, and it's nor replacing current scales, there's no objection from my side :)

PR #1165, opened 2024-05-08
O3 mda8

thorbjoernl:
## Related issue number

fix #863 

## Checklist

* [x] Start with a draft-PR
* [ ] The PR title is a good summary of the changes
* [ ] PR is set to AeroTools and a tentative milestone
* [ ] Documentation reflects the changes where applicable
* [ ] Tests for the changes exist where applicable
* [ ] Tests pass locally
* [ ] Tests pass on CI
* [ ] At least 1 reviewer is selected
* [ ] Make PR ready to review

PR #1224, opened 2024-06-24
MDA8

### LINKED ISSUE #863: MDA8
Hi everyone,

The daily maximum of the 8-hourly running mean ozone concentration (MDA8) is a metric often used by the WHO, and it would be really great if the observed values could be compared to EMEP (the model already outputs it). Is there any chance it would be possible to evaluate MDA8 using pyaerocom?

Willem

thorbjoernl:
## Change Summary

Implements the possibility to calculate mda8 values for specific variables. This is done by creating an additional ColocatedData object after colocation.

## Related issue number

closes #863 

## Checklist

* [x] Start with a draft-PR
* [x] Tests pass locally
* [x] Tests pass on CI
* [x] Make PR ready to review


thorbjoernl:
Notes from old PR: https://github.com/metno/pyaerocom/pull/1165

>Idea: O3 and CO both need MDA8. Therefore maybe making an operator which takes a ColocatedData object and applies for example a daily maximum of the 8 hourly mean (MDA8) could be useful. Would need to be specified in a config, and then dealt with here (similar to how we use `use_fairmode`): https://github.com/metno/pyaerocom/blob/63ca935cdea19c73ff45d960d9a91c46a384a720/pyaerocom/aeroval/coldatatojson_engine.py#L120

>Want a way to keep both the original timeseries and the MDA8. Expand ColocatedDataList?

pyaerocom/mda8/mda8.py
lewisblake, line 1: Are you sure this should be it's own directory? I would have thought under `/stats` might have been a natural place

pyaerocom/colocation/colocator.py
```diff
@@ -378,9 +380,21 @@ def run(self, var_list: list = None):
...
+
+                if obs_var in ["conco3", "vmro3"]:
+                    mda8 = None
+                    try:
+                        mda8 = mda8_colocated_data(
+                            coldata, obs_var=f"{obs_var}mda8", mod_var=f"{mod_var}mda8"
+                        )
+                    except ValueError:
+                        logger.warning(
+                            "Tried calculating mda8 for [{obs_var}, {mod_var}], but failed."
```
lewisblake: Should {obs_var}, {mod_var} be f-strings? Or did we want to avoid f-strings in warnings? Could be more useful to see exactly where this is happening.

pyaerocom/mda8/mda8.py
```diff
@@ -0,0 +1,86 @@
...
+    length = np.sum(~np.isnan(x))
+    if length < min_periods:
+        return np.nan
+
+    return np.nanmax(x)
+
+
+def mda8_colocated_data(
+    coldat: ColocatedData | ColocatedDataLists, /, obs_var: str, mod_var: str
+) -> ColocatedData | ColocatedDataLists:
```
lewisblake: Only the `ColocatedData` case needs to be accounted for. `ColocatedDataLists` is for vertical profiles, which I don't imagine us needing MDA8 for. So can remove all logic for `ColocatedDataLists`

pyaerocom/mda8/mda8.py
```diff
@@ -0,0 +1,86 @@
+import numpy as np
+import xarray as xr
+
+from pyaerocom.colocation.colocated_data import ColocatedData
+from pyaerocom.colocation.colocation_3d import ColocatedDataLists
+
+
+def min_periods_max(x: np.ndarray, /, min_periods=1) -> float:
```
lewisblake: type hint can return nan

tests/mda8/test_mda8.py
```diff
@@ -0,0 +1,116 @@
...
+        [[[x] for x in values]],
+        dims=["data_source", "time", "station_name"],
+        coords={"time": time},
+    )
+
+    mda8 = _calc_mda8(arr)
+
+    assert mda8.shape[1] == len(exp_mda8)
+
+    np.testing.assert_array_equal(mda8[0, :, 0], exp_mda8)
```
lewisblake: Check to see difference between this and `pytest.approx`.

pyaerocom/mda8/mda8.py
```diff
@@ -0,0 +1,86 @@
...
+        )
+    for cd in coldat.colocateddata_for_profile_viz:
+        colocateddata_for_profile_viz.append(
+            mda8_colocated_data(cd, obs_var=obs_var, mod_var=mod_var)
+        )
+
+    return ColocatedDataLists(colocateddata_for_statistics, colocateddata_for_profile_viz)
+
+
+def _calc_mda8(data: xr.DataArray) -> xr.DataArray:
```
lewisblake: Please provide the definition and link to the EU directive document annex XI in the docstring. Text given in issue.

tests/mda8/test_mda8.py
lewisblake, line 1: Add a few tests using the "Gauss" trick we talked about. Then add some nans and check those edge cases as well.

tests/mda8/test_mda8.py
```diff
@@ -0,0 +1,116 @@
...
+            1.28879322,
+            1.28807846,
+            1.28700801,
+        ],
+        decimal=5,
+    )
+
+
+@pytest.mark.parametrize("coldataset", ("fake_3d_hr",))
+def test_coldata_to_mda8_ColocatedDataLists(coldata):
```
lewisblake: Can remove

lewisblake (changes requested):

pyaerocom/colocation/colocator.py
```diff
@@ -378,9 +380,21 @@ def run(self, var_list: list = None):
                 coldata = self._run_helper(
                     mod_var, obs_var
                 )  # note this can be ColocatedData or ColocatedDataLists
-                if mod_var not in data_out:
-                    data_out[mod_var] = {}
                 data_out[mod_var][obs_var] = coldata
+
+                if obs_var in ["conco3", "vmro3"]:
```
lewisblake: We may also need this for CO. One way to do this would be to define a global variable `MDA8_VARS` which contains these variables and others we foresee needing, and check to see if obs_var is contained in that.

lewisblake (changes requested):
Small changes. Otherwise looks good

lewisblake (approved):
:smiley_cat:

PR #1230, opened 2024-06-26
put all temporary and testing data below the path  ~/MyPyaerocom

### LINKED ISSUE #1229: Testdata minimal should be user-specific
The test-data minimal is currently shared between users. These should be downloaded into the users' home directory instead to avoid conflicts between versions of the dataset and permission errors which was the cause of tests failing with sqlite OperationalError (See recent discussion in chat / meetings).

jgriesfeller:
The title says it all

## Change Summary

just return always a link to ~/MyPyaerocom in the Config class.
Makes all temporary pyaerocom data personal

## Related issue number

fix #1229 

## Checklist

* [x] Start with a draft-PR
* [x] Tests pass locally
* [x] Tests pass on CI
* [x] Make PR ready to review

### DIFF
--- a/pyaerocom/config.py
+++ b/pyaerocom/config.py
@@ -370,8 +370,7 @@ class Config:
     @property
     def OUTPUTDIR(self):
         """Default output directory"""
-        if not check_write_access(self._outputdir):
-            self._outputdir = chk_make_subdir(self.HOMEDIR, self._outhomename)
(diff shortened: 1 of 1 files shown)


lewisblake (approved):
We should maybe mention in the release that this change has occurred, so that people are aware if their `/home` is too full.

PR #1235, opened 2024-07-01
Add terms and conditions to the general issue template

lewisblake:
## Change Summary
Adds an assertion that must be checked under the General Issues template stating the issue is not one of the previously defined ones. This prevents, for example, people creating "bug" issues but not actually filling out the bug report so we do not get all the information needed to reproduce the bug.

## Related issue number

NA

## Checklist

* [x] Start with a draft-PR
* [x] Tests pass locally
* [x] Tests pass on CI
* [x] Make PR ready to review

### DIFF
--- a/.github/ISSUE_TEMPLATE/general_issue.md
+++ b/.github/ISSUE_TEMPLATE/general_issue.md
@@ -5,3 +5,5 @@ title: ''
 labels: ''
 
 ---
+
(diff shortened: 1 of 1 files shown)


heikoklein (changes requested):
Sorry, but I don't see the problem this issue tries to solve. We have no legal binding to solve any issue, so the user must always accept the possibility that the issue will not be addressed. I don't see that we are overwhelmed with general issues either. We don't reject the issue if the reporter does not tick the box.

lewisblake:
Happy to explain in person when you get into the office today.

PR #1245, opened 2024-07-06
Simplify mscw-reader path requirements, read year_available from files

### LINKED ISSUE #896: EMEP reader: Years in path
Is there something that can be done for the requirement for a year to be in the path? It is also unclear to me at which part of the path the year should be. 

For example, for an evaluation against 2021 data, the following path to model data was not accepted:

Path: /lustre/storeB/users/willemvc/modelruns/Gauss_test_2021/u7_ref
Error: Failed to load model data: u7ref (concpm10). Reason Could not find any year in u7_ref

But these paths were okay:

Path: /lustre/storeB/users/willemvc/modelruns/Gauss_test_2021/rep2023
Path: /lustre/storeB/users/willemvc/modelruns/Gauss_test_2021/vra2021

heikoklein:
## Change Summary

- The mscw-reader no longer reads unknown model-files. Known model-files are now clearly stated instead of guessed by globs.
- There is no longer a requirement of 4-digits in the data-dir which were interpreted as year.
- years-available are now read from the model-files like Base_fullrun.nc
- Trends/multi-year analysis can be given as a `data-dir` to a directory containing subdirectories with four-digits. The subdirectories must contain Base_*.nc files (with * being hour, day, month, fullrun)
- the `filepath` attribute was never used and therefore removed

## Related issue number

fixes #896 

## Checklist

* [x] Start with a draft-PR
* [x] Tests pass locally
* [x] Tests pass on CI
* [x] Make PR ready to review

### DIFF
--- a/pyaerocom/io/mscw_ctm/reader.py
+++ b/pyaerocom/io/mscw_ctm/reader.py
@@ -305,7 +305,7 @@ class ReadMscwCtm:
             file = os.path.split(path)[1]
 
             if self._get_tst_from_file(file) != ts_type:
-                logger.info(f"ignoring file {path}: not of type {ts_type}")
+                logger.debug(f"ignoring file {path}: not of type {ts_type}")
                 continue
 
             try:
--- a/pyaerocom/io/mscw_ctm/reader.py
+++ b/pyaerocom/io/mscw_ctm/reader.py
@@ -1,3 +1,4 @@
+import functools
 import glob
 import logging
 import os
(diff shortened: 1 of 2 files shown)


pyaerocom/io/mscw_ctm/reader.py
```diff
@@ -234,74 +235,91 @@ def _get_namelist_from_folder(self):
...
+        else:
+            folders = [d for _, d in sorted(zip(yrs, folders))]
+        return list(set(folders))
 
-            namelist.append(d)
+    @staticmethod
+    @functools.cache
+    def _get_year_from_nc(filename: str) -> int:
+        with xr.open_dataset(filename) as nc:
+            return np.mean(nc["time"][:]).data.astype("datetime64[Y]").astype(int) + 1970
```
lewisblake, line 272: Isn't EMEP output yearly files? If so, wouldn't it suffice to just take the first element? Or do you think this is safer?

pyaerocom/io/mscw_ctm/reader.py
```diff
@@ -171,16 +171,16 @@ class ReadMscwCtm:
 
     DEFAULT_FILE_NAME = "Base_day.nc"
 
+    YEAR_PATTERN = r".*((?:19|20)\d\d).*"
```
lewisblake, line 175: Could you please add some explanation in the code about what this regex is doing?

lewisblake (changes requested):
Minor changes. Looks good otherwise

pyaerocom/io/mscw_ctm/reader.py
heikoklein: The model starts on the 31.12. for spinup. Hourly and daily data can therefore start a day too early. It's safer to take the middle, and doesn't require any extra time to speak of, since xarray anyway reads all coordinate-variables including the time-variable.

pyaerocom/io/mscw_ctm/reader.py
heikoklein: Added docs: https://github.com/metno/pyaerocom/blob/ea8d117ec29362b44760f1436c210bf17da96ccc/pyaerocom/io/mscw_ctm/reader.py#L174

lewisblake (approved):

PR #1268, opened 2024-07-16
Use map_freqs in ModelMapsEngine

### LINKED ISSUE #1259: Incorrect map frequencies: ModelMapsSetup's maps_freq is not being picked up in ModelMapsEngine::_process_map_var
**Describe the bug**
The model_maps attribute which can be specified in a config file is not being picked up in `ModelMapsEngine::_process_map_var`, rather the frequencies checked there are based on these two lines of code:

```python
        freq = min(TsType(fq) for fq in self.cfg.time_cfg.freqs)
        freq = min(freq, self.cfg.time_cfg.main_freq)
```

Note that the min corresponds to the coarsest time resolution, in this case yearly. In these lines of code the freq being checked are coming from `time_cfg` insead.

 - Pyaerocom version: 0.21.dev0
 - Computing platform: Linux
 - Configuration file (if applicable): https://aeroval-test.met.no/emep_evaluation/pages/maps/?project=emep&experiment=2024-XXX_2022_ebas2&station=Ispra&observation=EBAS-m&parameter=concNtno3&tab=timeseries&statistic=nmb
 - Error message (if applicable): NA, runs w/o error

**To Reproduce**
1. Can either rerun Heiko's EMEP experiment or pick up any other config

**Expected behavior**
The frequency shown on the maps should be that given by `maps_freq`

**Screenshots**
See link

**Additional context**
NA

lewisblake:
## Change Summary

Pass the correct maps_freq in ModelMapsEngine. 

## Related issue number

- Short-term version of #1267
- Fix #1259 

## Checklist

* [x] Start with a draft-PR
* [x] Tests pass locally
* [x] Tests pass on CI
* [x] Make PR ready to review

### DIFF
--- a/pyaerocom/aeroval/modelmaps_engine.py
+++ b/pyaerocom/aeroval/modelmaps_engine.py
@@ -161,8 +161,7 @@ class ModelMapsEngine(ProcessingEngine, DataImporter):
                 logger.info(f"Skipping processing of {outname}: data already exists.")
                 return []
 
-        freq = min(TsType(fq) for fq in self.cfg.time_cfg.freqs)
-        freq = min(freq, self.cfg.time_cfg.main_freq)
+        freq = TsType(self.cfg.modelmaps_opts.maps_freq)
         tst = TsType(data.ts_type)
 
         if tst < freq:
--- a/pyaerocom/aeroval/modelmaps_engine.py
+++ b/pyaerocom/aeroval/modelmaps_engine.py
@@ -161,7 +161,12 @@ class ModelMapsEngine(ProcessingEngine, DataImporter):
                 logger.info(f"Skipping processing of {outname}: data already exists.")
                 return []
 
-        freq = TsType(self.cfg.modelmaps_opts.maps_freq)
+        maps_freq = TsType(self.cfg.modelmaps_opts.maps_freq)
+        if maps_freq == "coarsest":  # TODO: Implement this in terms of a TsType object. #1267
+            freq = min(TsType(fq) for fq in self.cfg.time_cfg.freqs)
+            freq = min(freq, self.cfg.time_cfg.main_freq)
+        else:
+            freq = maps_freq
         tst = TsType(data.ts_type)
 
         if tst < freq:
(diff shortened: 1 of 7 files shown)


lewisblake:
Looks like it is being passed down to the outputted config correctly as well:

```json
  "modelmaps_opts": {
        "maps_freq": "monthly",
        "maps_res_deg": 5
    },
```

heikoklein (commented):
The fix looks good, but I am worried that this will break many configs where maps_freq wasn't given explicitly, e.g. https://aeroval.met.no/pages/infos/?project=cams2-83&experiment=forecast-current-season has daily maps.

I suggest that maps_freq has a default of the currently used min(time_cfg.freqs, main_freq) rather than "monthly".

lewisblake:
The CAMS2_83 case brings up a good point, the `Literal["monthly", "yearly"]` type-hinting won't work if we need daily maps. The `"monthly"` default comes from the original default in this class, so it has supposedly always been the default, even if not implemented correctly. IIRC, this default comes from input from Augustin based on a concern that Aeroval in general becomes too slow showing contour maps for longer time series. I worry that if we set a default which is too fine, and people forget about this input when setting up their config, the compute time will significantly increase, and this will only become noticeable towards the end of the processing. I'll talk with Charlie a today about what his intention was when he last modified this piece of code.

heikoklein:
Can't we get back to the de-facto default - rather than the never implemented literal default, e.g. use the coarsest resolution, e.g.
`Literal["hourly", "daily", "monthly", "yearly", "coarsest"]`, default `"coarsest"`, and if coarsest is selected, we keep the old code, i.e. `min(ts_type)`?

heikoklein (commented):
Do we have some public documentation of `modelmap_opts->maps_freq`, e.g. in https://pyaerocom.readthedocs.io/en/latest/aeroval-examples.html (i only see it in the API/developer  documentation, but not in the user-documentation for config-files).

heikoklein (approved):
Looks good. I would love some docs for the config. (Maybe an issue of it's to create a fully documented config page?)

lewisblake:
No I don't see any documentation outside of the API documentation you mention. I agree, would love a full documented config. Config files are already a topic for the AeroTools workshop, and once more clearly defined, should be better documented.

PR #1273, opened 2024-07-19
Use locking for aerovaldb writes

### LINKED ISSUE #1270: pyaerocom/aeroval output should use aerovaldb-locking for all output
To make parallelization of pyaeroval-runs possible, the output of the different runs should not conflict with each other. AerovalDb has now locking implemented and pyaeroval should therefore use this locking feature in a fine-grained manner, e.g. log for as short as possible (but not so often that writes are much delayed).

- pyaeroval can be run for different models/variables in parallel, but to the same data directory
- the different model-runs don't interfere with each others.

thorbjoernl:
## Change Summary

- Implements io using aerovaldb in most places which weren't previously using it.
- Use locking when performing writes.
- Allows specifying avdb_resource in configuration but falls back to using old json_basedir if not provided.

## Related issue number

closes #1270

## Checklist

* [x] Start with a draft-PR
* [x] Tests pass locally
* [x] Tests pass on CI
* [x] Pin aerovaldb to tag.
* [x] Performance benchmark.
* [x] Make PR ready to review

## Additional Notes
- Test regressions
  - A test in `test_aeroval_HIGHLEV` outputs additional files. These seem to match the config for the test so I think the old behavior was wrong. I have adjusted the test case for the new behavior.


thorbjoernl:
### Performance Benchmark

(On PPI)

|                  | `main-dev`             | `aerovaldb-locking` (locking disabled) | `aerovaldb-locking` (locking enabled) |
|---------------|--------------------------|---------------------------------------------------|---------------------------------------------------|
| `pytest .`   |  25m12                  | 28m52                                                  | 26m27                                                             |

pyaerocom/aeroval/experiment_output.py
```diff
@@ -220,7 +222,8 @@ def _sync_heatmaps_with_menu_and_regions(self) -> None:
                             hm_data = self._check_hm_all_regions_avail(all_regions, hm_data)
                             hm[vardisp][obs][vert_code][mod][modvar] = hm_data
 
-            self.avdb.put_by_uuid(hm, fp)
+            with self.avdb.lock():
+                self.avdb.put_by_uri(hm, fp)
```
heikoklein, line 246: `with self.avdb.lock():` must start before line 205 `data = self.avdb.get_by_uri(fp)` otherwise data might be updated by another process.

pyaerocom/aeroval/experiment_output.py
```diff
@@ -403,7 +406,8 @@ def _check_clean_ts_file(self, fp) -> bool:
                 modified = True
                 logger.info(f"Removing data for model {mod_name} from ts file: {fp}")
 
-        self.avdb.put_by_uuid(data_new, fp)
+        with self.avdb.lock():
+            self.avdb.put_by_uri(data_new, fp)
```
heikoklein, line 430: Lock should be before line 389 (reading of fp)

pyaerocom/aeroval/setupclasses.py
```diff
@@ -538,10 +539,9 @@ def to_json(self, outdir: str, ignore_nan: bool = True, indent: int = 3) -> None
             json indentation
 
         """
-        filepath = os.path.join(outdir, self.json_filename)
-        data = self.json_repr()
-        write_json(data, filepath, ignore_nan=ignore_nan, indent=indent)
-        return filepath
+        with aerovaldb.open(f"json_files:{outdir}") as db:
+            with db.lock():
```
heikoklein, line 553: `json_files:{outdir}` must be a configuration option. The user needs to be able to switch to another implementation of aerovaldb from the configuration.

heikoklein (changes requested):
Some potential concurrency bugs and missing configuration-switch for aerovaldb implementations (currently hard-coded to `json_file` and `file-locking`). Looks good otherwise.

pyaerocom/aeroval/_processing_base.py
```diff
@@ -69,6 +73,222 @@ def run(self, *args, **kwargs) -> list:
         """
         pass
 
+    def _add_heatmap_timeseries_entry(
```
heikoklein: Why are the implementations part of the base class? ModelMapEngine which inherits from this class should not have a _add_heatmap_timeseries_entry ...

They could be kept in the helpers as static functions, e.g. as 
```python
def _add_heatmap_timeseries_entry(engine: ModelMapEngine, ...
```
since they only access public information of the engine.

pyaerocom/aeroval/coldatatojson_helpers.py
```diff
@@ -8,6 +8,7 @@
 from datetime import datetime, timezone
 from typing import Callable, Literal
 
+import aerovaldb
```
heikoklein: is that used?

pyaerocom/aeroval/experiment_output.py
```diff
@@ -5,6 +5,7 @@
 from collections import namedtuple
 
 import aerovaldb
+import aerovaldb.aerovaldb
```
heikoklein: This looks wrong? Maybe the export/__init__.py of `aerovaldb` needs to be improved if such code here is needed?

pyaerocom/aeroval/experiment_output.py
```diff
@@ -78,7 +85,7 @@ class ExperimentOutput(ProjectOutput):
 
     def __init__(self, cfg: EvalSetup):
         self.cfg = cfg
-        super().__init__(cfg.proj_id, cfg.path_manager.json_basedir)
+        super().__init__(cfg.proj_id, f"{cfg.avdb_cfg.resource}:{cfg.path_manager.json_basedir}")
```
heikoklein: pyaerocom should not need to know about a json_basedir, which only exists if the json is used as `avdb_cfg.resource=json`.
Database-resources like hire for aerovaldb usually have a one-line resource identifier, typically as uri. There should be no 2 config places for aerovaldb which you have to concatenate, only one.

pyaerocom/utils.py
```diff
@@ -142,3 +144,34 @@ def print_file(path: Path | str):
...
+def recursive_defaultdict(d: Mapping | None = None) -> defaultdict:
+    """
+    Creates a nested default dict which creates empty subdicts
+    automatically on access. Useful to avoid lots of boiler plate
+    code for creating empty subdicts.
+
+    Example:
+
+    ```
+    dd = nested_defaultdict()
```
heikoklein: `recursive_defaultdict()` ?

pyaerocom/utils.py
```diff
@@ -142,3 +144,34 @@ def print_file(path: Path | str):
     for line in path.read_text().splitlines():
         if line.strip():
             print(line)
+
+
+def recursive_defaultdict(d: Mapping | None = None) -> defaultdict:
+    """
+    Creates a nested default dict which creates empty subdicts
```
heikoklein: nested -> recursive

pyaerocom/utils.py
```diff
@@ -142,3 +144,34 @@ def print_file(path: Path | str):
     for line in path.read_text().splitlines():
         if line.strip():
             print(line)
+
+
+def recursive_defaultdict(d: Mapping | None = None) -> defaultdict:
```
heikoklein: Change return-type to `dict`, since that is how it should be used. No need to expose defaultdict.

Recursion can be very dangerous. Would it be possible to limit the recursion-levels on construction, e.g.
```
def recursive_defaultdict(d: Mapping | None = None, max_depth: int) -> defaultdict:
```

heikoklein (changes requested):

pyaerocom/utils.py
thorbjoernl: defaultdict is a subclass of dict, so I think defaultdict is the correct return type. Any check isinstance check for dict will still return true.

recursivedefaultdict only calls the factory function on access of a key in the dictionary, not when the dictionary is initialized. I'm not sure the recursion is a problem in this instance.

pyaerocom/aeroval/_processing_base.py
thorbjoernl: The main reason was to avoid having to either pass in an aerovaldb instance to the function, or open a new connection in each function call, neither of which I particularly like. How would you feel about moving these to `ExperimentOutput`? As far as I can see each Engine class has an instance of ExperimentOutput, so self.experiment_output._add_heatmap... would work.

heikoklein (approved):
Looks good, thanks.

Issue #1175, opened 2024-05-29
Evaluation of uemep output in pyaeroval


heikoklein:
**Is your feature request related to a problem? Please describe.**
The uemep-model produces air-quality output for different areas in very high resolution (up to 50m). The areas are usually in a projection, e.g. UTM-33 for Norway or LAEA for Europe and the output-files are often calculated in smaller subdomains which sum up to the total (tiles). Variables have also different names than those used for e.g. EMEP.
Example data are (thanks @eivindgw ):
* Europe in LAEA: /lustre/storeB/project/fou/kl/emep/ModelRuns/uEMEP/CAMEO/uEMEP/Jan2019_v1/
* Norway in UTM-33:
  * Tiles (daily and yearly per municipality): /lustre/storeB/project/fou/kl/emep/ModelRuns/uEMEP/Calculations/Mapping_2022_v4e/
  * Stations (hourly-file per dag, all known NRT stations): /lustre/storeB/project/fou/kl/emep/ModelRuns/uEMEP/Calculations/Stations_2022_v4e/

**Describe the solution you would like to see**
A pyaerocom config should exist to run these data. It should be easy to make the results available on the aeroval-test web-page.

heikoklein:
Files are usually in one directory.
Files have two place-holders: datetime and tiles, e.g. 
`uEMEP_emep_report_2024_eecca_00166_20220101_00.nc`
`uEMEP_emep_report_2024_eecca_{tile:05d}_{date:%Y%m%d}_{utc:02d}.nc`

data can be hourly to yearly, `tile` or `date` or `utc` can be omitted.

`date` and `utc` tiles can be concatenated with `xarray.open_mfdataset`, `tile` should be separately.

heikoklein:
@eamousing @brucerd @eivindgw FYI

heikoklein:
Files might contain hourly, or yearly output (maybe also daily and monthly output). Since it is not possible to see from the file what time-resolution it has (in particular if it only has one time-stamp), resolution should be given as additional parameter (`ts_type`)

heikoklein:
EMEP4NO and similar emep runs in are input for uemep and should work, too. Requires collocation with projection information.

Example-file: /lustre/storeB/project/fou/kl/CAMS50/output_EMEP/202406/CWF_18AQNO-20240624_hour.nc

@eamousing @brucerd @eivindgw I would like to have an older, complete year EMEP4NO file, e.g. 2022 or 2020.

[...]

PR #1274, opened 2024-07-20
Collocation of emep-mscw output with gridded data in a projection

heikoklein:
## Change Summary

- Add `ProjectionInformation` `proj_info` to `GriddedData`, the projection projection-information consist of a projection functions to convert between lat/lon and x/y in the projection, as well as of the dimension-names of the x/y-axes.
- Read `ProjectionInformation` when reading a variable in mscw-ctm reader and add that to GriddedData
- Collocation, when using the xarray-collocation is now able to convert the observations lat/lon to x/y and perform collocation on x/y values in the projection-plane when reading the gridded-data.
- Make `pyproj` an explicit dependency (currently only implicit dependency of `iris` and `cartopy`)

## Related issue number

solves the EMEP4NO part of #1175 

## Checklist

* [x] Start with a draft-PR
* [x] Tests pass locally
* [x] Tests pass on CI
* [x] Make PR ready to review

### DIFF
--- a/pyaerocom/colocation/colocation_utils.py
+++ b/pyaerocom/colocation/colocation_utils.py
@@ -759,9 +759,17 @@ def colocate_gridded_ungridded(
     else:
         # gridded data with projection,
         # add x/y information to ungridded
-        xvals, yvals = data.proj_info.to_proj(latitude, longitude)
-        xrange = (np.min(xvals), np.max(xvals))
-        yrange = (np.min(yvals), np.max(yvals))
+        for coord in data.cube.dim_coords:
+            if coord.var_name == data.proj_info.x_axis:
+                vals = coord.points
+                xrange = (np.min(vals), np.max(vals))
+            if coord.var_name == data.proj_info.y_axis:
+                vals = coord.points
+                yrange = (np.min(vals), np.max(vals))
+        if xrange is None or yrange is None:
+            raise VariableDefinitionError(
+                f"x/y axis not found in cube: {data.proj_info.x_axis}, {data.proj_info.y_axis}"
+            )
         data_ref = data_ref.filter_by_projection(data.proj_info.to_proj, xrange, yrange)
 
     # get timeseries from all stations in provided time resolution
--- a/pyaerocom/colocation/colocation_utils.py
+++ b/pyaerocom/colocation/colocation_utils.py
@@ -681,10 +681,12 @@ def colocate_gridded_ungridded(
     """
     if filter_name is None:
         filter_name = const.DEFAULT_REG_FILTER
-    try:
-        data.check_dimcoords_tseries()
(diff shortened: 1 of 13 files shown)


pyaerocom/colocation/colocation_utils.py
```diff
@@ -681,10 +681,12 @@ def colocate_gridded_ungridded(
     """
     if filter_name is None:
         filter_name = const.DEFAULT_REG_FILTER
-    try:
-        data.check_dimcoords_tseries()
-    except DimensionOrderError:
-        data.reorder_dimensions_tseries()
+    if data.proj_info is None:
+        # reordering only implemented if no projection
```
lewisblake, line 685: This try/except block appears several times but I forget what the reordering of the dimensions does. Could you please extend your comment to make a note of what it's doing and why this only happens for non-projected data.

pyaerocom/scripts/testdata-minimal/README.md
lewisblake, line 1: Aren't these changes already in main-dev?

pyaerocom/ungriddeddata.py
```diff
@@ -1926,6 +1926,44 @@ def apply_filters(self, var_outlier_ranges=None, **filter_attributes):
...
+        :param yrange: y range (min/max included) in the projection plane
+        """
+        meta_matches = []
+        totnum = 0
+        for meta_idx, meta in self.metadata.items():
+            lon = meta["longitude"]
+            lat = meta["latitude"]
+            x, y = projection(lat, lon)
+
+            matchX = in_range(x, xrange[0], xrange[1])
```
lewisblake: change to snake-case https://peps.python.org/pep-0008/

lewisblake (changes requested):
small changes, looks good

pyaerocom/scripts/testdata-minimal/README.md
heikoklein: Not sure what you mean? git tells us here they aren't?

pyaerocom/colocation/colocation_utils.py
heikoklein: Sorry, I just implemented the code for projected data and left the code for non-projected data untouched. I can only guess from function names that the old code tries to reorder to "time, level, lon, lat" if this is not the natural order of data - not sure if this is a iris-requirement or where this old code came from.

I could add some synthetic test-data which might raise some problems when not ordered, but all real data we have doesn't have this problem. So I would rather only implement re-ordering of dimensions if this becomes an issue.

lewisblake (approved):

pyaerocom/colocation/colocation_utils.py
lewisblake: Sounds good, don't worry about it then.

PR #1312, opened 2024-08-16
One way to change freq for map processing

### LINKED ISSUE #1260: Model maps shown for EMEP should use the frequency of the Base_*.nc file rather than the colocated data and then resampling
An inefficiency in the creation of the model maps is that for some models, such as EMEP, we already have yearly files available, so reading in daily data and then temporally averaging is unnecessarily expensive.

In the case of making maps for the EMEP model, the `maps_freq` argument set in a config file should be used to read the `Base_{maps_freq}.nc` file already available, regardless of the colocation frequency `main_freq`.  Some information in the logger should be present to inform that this option is being used.

**Additional context**
Blocked by #1259

dulte:
## Change Summary

Made it so when ModelMapsEngine is run, main_freq is changed to maps_freq. If maps_freq=coarsest, then the coarsest freq of the given freqs are chosen

## Related issue number

Proposed fix to #1260

## Checklist

* [x] Start with a draft-PR
* [x] Tests pass locally
* [ ] Tests pass on CI
* [ ] Make PR ready to review


lewisblake (changes requested):
Based on our discussion I think we approach the other way. :smile_cat:

pyaerocom/aeroval/modelmaps_engine.py
```diff
@@ -131,9 +139,9 @@ def _process_map_var(self, model_name, var, reanalyse_existing):
         ModelVarNotAvailable
             If model/var data cannot be read
         """
-
         try:
-            data = self.read_model_data(model_name, var)
+            # data = self.read_model_data(model_name, var)
```
lewisblake: Please remove dead code. Also why rename `read_model_data` to `read_modeldata`?

lewisblake (changes requested):
Looks good. Only minor changes. I would opt for `model_data` not `modeldata` because the former is more readable and modeldata isn't an English word. I like the `_get_maps_freq` function. We should use that in the overlay maps as well. If you haven't please merge with a recent version of `main-dev` as well.

pyaerocom/aeroval/modelmaps_engine.py
dulte: Agree. I changed the name to read_model_data and removed the dead  code

dulte:
> Looks good. Only minor changes. I would opt for `model_data` not `modeldata` because the former is more readable and modeldata isn't an English word. I like the `_get_maps_freq` function. We should use that in the overlay maps as well. If you haven't please merge with a recent version of `main-dev` as well.

Agreed. I've changed the name. 

I have merged it in now. The overlay map now uses both `_get_maps_freq` and the new model reading method. Could you have a look over the new changes? The merging was quite messy, but I think I manged to put everything in the right order

lewisblake (approved):
Nice! Thank you :1st_place_medal:

PR #1322, opened 2024-08-22
Fix mda8 misalignment

### LINKED ISSUE #1323: MDA daily binning misaligned due to mislabeled data in 8hour mean calculation
**Describe the bug**
 - Pyaerocom version: 0.22.dev0
 - Computing platform: NA

**To Reproduce**
1. Add the following parametrization to `test_mda8.py:test_calc_mda8`
```
        pytest.param(
            xr.date_range(start="2024-01-01 06:00:00", periods=30, freq="1h"),
            np.arange(30),
            [np.nan, 25.5]
        ),
```
2. The test fails because the last 8 hour mean included in the first day is for the period 0000-0800 the next day, instead of 1600-2400. Therefore not enough data is included in day two and it results in nan.

**Expected behavior**
MDA8 for 2024-01-02 should be a non-nan numeric value (25.5).

**Screenshots**

**Additional context**
Caused by wrong alignment. xarray's DataArray.rolling labels data to the left, but labeling it to the right is expected by the current implementation. DataArray.rolling does not support a label-argument so another fix needs to be found. 

Related: https://github.com/metno/pyaerocom/pull/1322

thorbjoernl:
## Change Summary

Per discusion with @avaldebe, there is an issue where the mda8 calculation relies on undocumented xarray behaviour. This PR aims to make arguments explicit where possible as well as testing for the undocumented behaviour in xarray so that possible future regressions can be detected. This also fixes an issue where 8 hour average was left labeled instead of rightlabeled, resulting in the mda8 binning to be shifted by 8 hours from the defined window.

## Related issue number

closes #1323 

## Checklist

* [x] Start with a draft-PR
* [x] Tests pass locally
* [x] Tests pass on CI
* [x] Make PR ready to review

### DIFF
--- a/pyaerocom/stats/mda8/mda8.py
+++ b/pyaerocom/stats/mda8/mda8.py
@@ -100,10 +100,11 @@ def _calc_mda8(data: xr.DataArray) -> xr.DataArray:
 
 
 def _rolling_average_8hr(arr: xr.DataArray) -> xr.DataArray:
+    # Labeling should be right which probably is the default in xarray.
     return arr.rolling(time=8, min_periods=6).mean()
 
 
 def _daily_max(arr: xr.DataArray) -> xr.DataArray:
-    return arr.resample(time="24H", offset="1h").reduce(
+    return arr.resample(time="24H", origin="start_day", label="left", offset="1h").reduce(
         lambda x, axis: np.apply_along_axis(min_periods_max, 1, x, min_periods=18)
     )
--- a/pyaerocom/stats/mda8/mda8.py
+++ b/pyaerocom/stats/mda8/mda8.py
@@ -100,8 +100,13 @@ def _calc_mda8(data: xr.DataArray) -> xr.DataArray:
 
 
 def _rolling_average_8hr(arr: xr.DataArray) -> xr.DataArray:
-    # Labeling should be right which probably is the default in xarray.
-    return arr.rolling(time=8, min_periods=6).mean()
+    # Xarray labels the data left, while we want it to be labeled right, so we add 8h to
+    # the time index.
+    new_arr = arr.rolling(time=8, min_periods=6).mean()
+    # Xarray labels the data left, while we want it to be labeled right, so we add 8h to
+    # the time index.
+    new_arr["time"] = new_arr.get_index("time") + pd.Timedelta("8h")
+    return new_arr
 
 
 def _daily_max(arr: xr.DataArray) -> xr.DataArray:
--- a/pyaerocom/stats/mda8/mda8.py
+++ b/pyaerocom/stats/mda8/mda8.py
@@ -100,13 +100,8 @@ def _calc_mda8(data: xr.DataArray) -> xr.DataArray:
 
 
 def _rolling_average_8hr(arr: xr.DataArray) -> xr.DataArray:
-    # Xarray labels the data left, while we want it to be labeled right, so we add 8h to
-    # the time index.
-    new_arr = arr.rolling(time=8, min_periods=6).mean()
-    # Xarray labels the data left, while we want it to be labeled right, so we add 8h to
-    # the time index.
-    new_arr["time"] = new_arr.get_index("time") + pd.Timedelta("8h")
-    return new_arr
+    # Xarray labels the data right based on the last data point in the period for rolling.
+    return arr.rolling(time=8, center=False, min_periods=6).mean()
 
 
 def _daily_max(arr: xr.DataArray) -> xr.DataArray:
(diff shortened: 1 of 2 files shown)


pyaerocom/stats/mda8/mda8.py
```diff
@@ -100,10 +100,16 @@ def _calc_mda8(data: xr.DataArray) -> xr.DataArray:
 
 
 def _rolling_average_8hr(arr: xr.DataArray) -> xr.DataArray:
-    return arr.rolling(time=8, min_periods=6).mean()
+    # Xarray labels the data left, while we want it to be labeled right, so we add 8h to
+    # the time index.
+    new_arr = arr.rolling(time=8, min_periods=6).mean()
+    # Xarray labels the data left, while we want it to be labeled right, so we add 8h to
+    # the time index.
+    new_arr["time"] = new_arr.get_index("time") + pd.Timedelta("8h")
```
avaldebe: Why is the `new_arr.get_index("time")` needed?
Can it be written as 
``` python
    new_arr["time"] += pd.Timedelta("8h")
```

tests/stats/mda8/test_mda8.py
```diff
@@ -26,13 +27,13 @@ def test_data(time, values) -> xr.DataArray:
         pytest.param(
             xr.date_range(start="2024-01-01 01:00", periods=49, freq="1h"),
             [0.0] * 49,
-            [0, 0, np.nan],
+            [np.nan, 0, np.nan],
```
avaldebe: why is the MDA8 for the 1st day nan?
The 8-hour mean needs 6 or more hours to compute, so the 1st day has 19 8-hour means: 01-06, 01-07, 01-08, ..., 17-24

tests/stats/mda8/test_mda8.py
```diff
@@ -49,6 +50,13 @@ def test_data(time, values) -> xr.DataArray:
             [np.nan] * 3,
             id="with-nans",
         ),
+        # https://github.com/metno/pyaerocom/issues/1323
+        pytest.param(
+            xr.date_range(start="2024-01-01 06:00:00", periods=30, freq="1h"),
+            np.arange(30),
+            [np.nan, 25.5],
```
avaldebe: Not sure if the MDA8 for the 1st day is supposed to be NaN.
Similarly to my previous comment, there should be enough 8-hour means for the daily max.

tests/stats/mda8/test_mda8.py
```diff
@@ -154,6 +153,27 @@ def test_rollingaverage(test_data, exp_ravg):
...
+        dims=["data_source", "time", "station_name"],
+        coords={"time": xr.date_range(start="2024-01-01 00:00", periods=24, freq="1h")},
+    )
+
+    ravg = _rolling_average_8hr(data)
+
+    assert np.all(
+        ravg.get_index("time")
+        == xr.date_range(start="2024-01-01 00:00", periods=24, freq="1h") + pd.Timedelta("8h")
+    )
```
avaldebe: Can this be re-written without `np.all`?
Maybe you can reuse `data["time"]` like this
``` python
    assert ravg.get_index("time") == data["time"] + pd.Timedelta("8h").all()
```
but that looks very similar to the actual implementation,
perhaps something like this would be better
``` python
    assert set(data["time"]-ravg.get_index("time")) == {pd.Timedelta("-8h")}
```
or assert on the first and last time stamps
``` python
    assert date["time"].isel(time=0).dt.strftime("%F %R") == "2024-01-01 08:00"
    assert date["time"].isel(time=23).dt.strfmt("%F %R") == "2024-01-02 08:00"

avaldebe (dismissed):
I'm not sure about the number of valid 8-hour means are correct for the first MDA8

tests/stats/mda8/test_mda8.py
thorbjoernl: 1-9, 2-10, 3-11, 4-12, 5-13, (<- These are NaN because of 6/8 required for 8h average), 6-14, 7-15, 8-16, 9-17, 10-18, 11-19, 12-20, 13-21, 14-22, 15-23, 16-24, (Next day ->), 17-1...

I count 11 observations. Is this wrong?

pyaerocom/stats/mda8/mda8.py
thorbjoernl: += Doesn't work.

This is the documented way of doing time offset arithmetic: https://github.com/pydata/xarray/pull/8479/files

heikoklein (approved):

PR #1343, opened 2024-09-16
Adds unit conversion to read_pyaro

dulte:
## Change Summary

After some problems with the EEA reader and units, I've added unit conversion to read_pyaro

## Checklist

* [x] Start with a draft-PR
* [x] Tests pass locally
* [x] Tests pass on CI
* [ ] Make PR ready to review


dulte:
It is more or less done. But I get strange results with my calculated factor for the conversion between mg S/l and ugm-3. The resulting observations from EEA then gets 6 order of magnitude larger than the model. So there is something I don't understand with the mgS/l units...

dulte:
This might be the wrong way of handling the units. Will remove PR as of now

PR #1347, opened 2024-09-24
Extend aeroval to plot satellite pixels maps

### LINKED ISSUE #401: AeroVal maps feature for raster data
The geojson files can get very large for long time series (eg: >50MB for monthly data over 3 years in C3S project).

One year of data with monthly resolution (12 layers) has been shown to be fast enough both in loading the data and in plotting the maps. For multi-year projects and monthly resolutions, we could sinply split up the time period into individual years and provide a geojson file per year. 

We could also investigate web map tiles services (https://www.ogc.org/standards/wmts), but this is a major change.

lewisblake:
## Change Summary

- Remove old `json` file computations (in `contour` directory)
- `ranges.json` contains a unit for each variable 
- Implement requested attributes in `ModelMapsSetup`, and associated helper classes and methods, such as `BoundingBox` and `validate_plot_types`
- Implement `plot_overlay_pixel_maps`
- Rename `_process_map_var` to `_process_contour_map_var`
- Fix tests

I have added some small tests for `modelmaps_opts` in `setup_classes.py` but not for the actual processing of the pixel maps as analogous tests do not seem to exist for the contour plotting.

## Related issue number

closes #401, https://github.com/metno/AeroToolsIssues/issues/5

## Checklist

* [x] Start with a draft-PR
* [x] Tests pass locally
* [x] Tests pass on CI
* [ ] Make PR ready to review

### DIFF
--- a/pyaerocom/aeroval/modelmaps_engine.py
+++ b/pyaerocom/aeroval/modelmaps_engine.py
@@ -100,18 +100,21 @@ class ModelMapsEngine(ProcessingEngine, DataImporter):
             logger.info(f"Processing model maps for {model_name} ({var})")
 
             try:
-                if (
-                    self.cfg.modelmaps_opts.plot_types is CONTOUR
-                    or CONTOUR in self.cfg.modelmaps_opts.plot_types.get(model_name, False)
-                ):
+                make_contour, make_overlay = False, False
+                if isinstance(self.cfg.modelmaps_opts.plot_types, dict):
+                    make_contour = CONTOUR in self.cfg.modelmaps_opts.plot_types.get(
+                        model_name, False
+                    )
+                    make_overlay = OVERLAY in self.cfg.modelmaps_opts.plot_types.get(
+                        model_name, False
+                    )
+
+                if self.cfg.modelmaps_opts.plot_types is CONTOUR or make_contour:
                     _files = self._process_contour_map_var(
                         model_name, var, self.reanalyse_existing
                     )
                     files.extend(_files)
-                if (
-                    self.cfg.modelmaps_opts.plot_types is OVERLAY
-                    or OVERLAY in self.cfg.modelmaps_opts.plot_types.get(model_name, False)
-                ):
+                if self.cfg.modelmaps_opts.plot_types is OVERLAY or make_overlay:
                     # create overlay (pixel) plots
                     _files = self._process_overlay_map_var(
                         model_name, var, self.reanalyse_existing
--- a/pyaerocom/aeroval/modelmaps_engine.py
+++ b/pyaerocom/aeroval/modelmaps_engine.py
@@ -3,7 +3,9 @@ import os
 
 from pyaerocom import GriddedData, TsType
 from pyaerocom.aeroval._processing_base import DataImporter, ProcessingEngine
-from pyaerocom.aeroval.modelmaps_helpers import calc_contour_json, griddeddata_to_jsondict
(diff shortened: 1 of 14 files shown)


pyaerocom/aeroval/modelmaps_engine.py
```diff
@@ -176,24 +209,91 @@ def _process_map_var(self, model_name, var, reanalyse_existing):
...
+                data[i],
+                cmap=varinfo.cmap,
+                cmap_bins=varinfo.cmap_bins,
+                format=self.cfg.modelmaps_opts.overlay_save_format,
             )
 
-        return [fp_json, fp_geojson]
+            with self.avdb.lock():
+                self.avdb.put_map_overlay(
+                    overlay_plot,
```
heikoklein, line 293: Does this mean that put_map_overlay has no knowledge of the overlay_plot image-type (png/jpeg,webp)? And the get_map_overlay does not have that knowledge either?

@thorbjoernl 
I would feel better if the image-type were known..., e.g. 
`ROUTE_MAP_OVERLAY = "/v0/map-overlay/{project}/{experiment}/{source}/{variable}/{date}.{img_type}"` would even indicate that the files on dict in the jsonfiledb are images...

pyaerocom/aeroval/modelmaps_helpers.py
```diff
@@ -146,3 +103,41 @@ def calc_contour_json(data, cmap, cmap_bins):
...
+def plot_overlay_pixel_maps(
+    data: xarray.DataArray, cmap: str, cmap_bins: list[float], format: str
+):  # pragma: no cover
+    plt.close("all")
+    matplotlib.use("Agg")
+
+    fig, axis = plt.subplots(
+        1,
+        1,
+        subplot_kw=dict(projection=ccrs.Mercator()),
```
heikoklein, line 117: Why are there two projections, Mercator and PlateCarree in this function? We should only have one output projection.

pyaerocom/aeroval/setup_classes.py
```diff
@@ -121,6 +125,25 @@ def get_json_output_dirs(self, assert_exists=True):
...
+    map_observations_only_in_right_menu: bool = False
+    overlay_save_format: Literal["webp", "png"] = "webp"
+
+    @field_validator("plot_types")
+    def validate_plot_types(cls, v):
+        if isinstance(v, dict):
+            for m in v:
+                if not isinstance(v[m], set):
+                    v[m] = set([v[m]])
+                assert v[m] in PLOT_TYPE_OPTIONS
```
heikoklein: The python compiler may remove all `assert` statements for optimized code. You want to have a real exception here.

heikoklein (approved):
Looks good, with a few comments.

Use `assert` only for debugging assumptions, not as exception shortcut.

Discuss with Thorbjørn if image-types should be stored as attribute of the blob.

pyaerocom/aeroval/modelmaps_helpers.py
lewisblake: The `transform` argument tells cartopy which projection our data is in. The `projection` argument tells cartopy which projection to project into. https://scitools.org.uk/cartopy/docs/latest/tutorials/understanding_transform.html

pyaerocom/aeroval/modelmaps_engine.py
lewisblake: Yes I think this could make sense as well. I will not stop this from merging as you have approved the PR, but perhaps something to pick up later.

PR #1358, opened 2024-10-01
Fix model validator for pyaro config and ColocatedData

### LINKED ISSUE #1349: Clean up pydantic model validators
**Describe the bug**

Experiments are producing warnings with 

```console
/modules/rhel8/user-apps/fou-kl/aerotools/pya-v2024.09/lib/python3.11/site-packages/pydantic/main.py:211: UserWarning: A custom validator is returning a value other than `self`.
Returning anything other than `self` from a top level model validator isn't supported when validating via `__init__`.
See the `model_validator` docs (https://docs.pydantic.dev/latest/concepts/validators/#model-validators) for more details.
```

This may be a suspect 
https://github.com/metno/pyaerocom/blob/4ef307b88f204f62c8c8fa61d05dbb5faa6518be/pyaerocom/colocation/colocation_setup.py#L492

 - Pyaerocom version: 0.23.dev0
 - Computing platform: all

**To Reproduce**
1. run any config and check the output
2.
3.

**Expected behavior**
Cleaner standard out

**Screenshots**

**Additional context**
There are also warning related to iris.

lewisblake:
## Change Summary

Fix the pyaro config model validator, which had been raising warning and was never really implemented correctly before. Also fix the `ColocatedData` model_validator

## Related issue number

Close #1349

## Checklist

* [x] Start with a draft-PR
* [x] Tests pass locally
* [x] Tests pass on CI
* [x] Make PR ready to review

### DIFF
--- a/pyaerocom/colocation/colocated_data.py
+++ b/pyaerocom/colocation/colocated_data.py
@@ -34,19 +34,18 @@ from pyaerocom.time_resampler import TimeResampler
 logger = logging.getLogger(__name__)
 
 
-def ensure_correct_dimensions(data: np.ndarray | xr.DataArray):
+def ensure_correct_dimensions(data: xr.DataArray):
     """
-    Ensure the dimensions on either a numpy aray or xarray passed to ColocatedData.
+    Ensure the dimensions on an xarray.DataArray passed to ColocatedData.
     If a ColocatedData object is created outside of pyaerocom, this checking is needed.
     This function is used as part of the model validator.
     """
     shape = data.shape[0]
-    if isinstance(data, np.ndarray):
-        num_dims = data.ndim
-    elif isinstance(data, xr.DataArray):
-        num_dims = len(data.dims)
-    else:
+
+    if not isinstance(data, xr.DataArray):
         raise ValueError("Could not interpret data")
+    num_dims = len(data.dims)
+
     if num_dims not in (2, 3, 4):
         raise DataDimensionError("invalid input, need 2D, 3D or 4D numpy array")
     elif not shape == 2:
--- a/pyaerocom/colocation/colocated_data.py
+++ b/pyaerocom/colocation/colocated_data.py
@@ -124,6 +124,8 @@ class ColocatedData(BaseModel):
 
     @model_validator(mode="after")
     def validate_data(self):
+        if self.data is None:
+            return self
         if isinstance(self.data, Path):
             # make sure path is str instance
             self.data = str(self.data)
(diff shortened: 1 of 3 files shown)


pyaerocom/colocation/colocated_data.py
```diff
@@ -128,13 +128,11 @@ def validate_data(self):
             # make sure path is str instance
             self.data = str(self.data)
         if isinstance(self.data, str):
-            assert self.data.endswith("nc"), ValueError(
-                "Invalid data filepath str, must point to a .nc file"
-            )
+            if not self.data.endswith("nc"):
+                raise ValueError("Invalid data filepath str, must point to a .nc file")
```
thorbjoernl: I would print the file path as part of the exception.

pyaerocom/colocation/colocated_data.py
```diff
@@ -128,13 +128,11 @@ def validate_data(self):
...
             self.open(self.data)
         elif isinstance(self.data, xr.DataArray):
```
thorbjoernl: I think these can be flipped to reduce nesting: if isinstance(ndarray) -> make_DataArray; then ensure_correct_dimensions.

This would also simplify ensure_correct_dimensions to not have to handle ndarray

pyaerocom/colocation/colocation_setup.py
```diff
@@ -488,25 +489,26 @@ def basedir_logfiles(self):
...
-        if v is not None:
-            if isinstance(v, dict):
+            self.obs_config = None
+        if self.obs_config is not None:
+            if isinstance(self.obs_config, dict):
                 logger.info("Obs config was given as dict. Will try to convert to PyaroConfig")
-                v = PyaroConfig(**v)
-            if v.name != cls.obs_id:
+                self.obs_config = PyaroConfig(**self.obs_config)
+            if self.obs_id.name != self.obs_id:
```
thorbjoernl: According to the type definition obs_id is str or None which doesn't have a name property?

pyaerocom/colocation/colocated_data.py
```diff
@@ -128,13 +128,11 @@ def validate_data(self):
             # make sure path is str instance
             self.data = str(self.data)
         if isinstance(self.data, str):
-            assert self.data.endswith("nc"), ValueError(
-                "Invalid data filepath str, must point to a .nc file"
-            )
+            if not self.data.endswith("nc"):
+                raise ValueError("Invalid data filepath str, must point to a .nc file")
             self.open(self.data)
```
thorbjoernl, line 136: It is currently not clear that read files are dimension validated without going two function calls deep, which makes this look like a bug without being one. Maybe it can be made clearer?

thorbjoernl (changes requested):

pyaerocom/colocation/colocation_setup.py
lewisblake: Yes, good catch. was a typo, should be `self.obs_config.name`

thorbjoernl (approved):
LGTM

PR #1360, opened 2024-10-02
Pydantic ObsEntry

lewisblake:
## Change Summary

Deeper work motivated by desire to generalize model naming convention.
First step of making `ObsCollection` and `ModelCollection` child classes of `BaseModel` 
Partly related to improving the `ColocatedData` interface.
Rename `fileconventions` -> `file_conventions`
EvalSetup.get_obs_entry() now returns an instance of `ObsEntry`, not a dictionary. Clean up parts of the code which relied on this dictionary to instead, f. eks., access instance attributes.
Remove `HasColocator._get_diurnal_only()` and associated tests. Instead give `ObsEntry` an attribute `diurnal_only` with a default of `False`.

## Related issue number
Part of #1085
## Checklist

* [x] Start with a draft-PR
* [x] Tests pass locally
* [x] Tests pass on CI
* [x] Make PR ready to review


pyaerocom/aeroval/collections.py
```diff
@@ -22,8 +22,8 @@ def _check_entry_name(self, key):
 
     def __setitem__(self, key, value):
         self._check_entry_name(key)
-        if "web_interface_name" in value:
-            self._check_entry_name(value["web_interface_name"])
+        # if "web_interface_name" in value:
+        #     self._check_entry_name(value["web_interface_name"])
```
dulte: Remove dead code

tests/aeroval/test_setup_classes.py
```diff
@@ -32,34 +31,44 @@ def test_EvalSetup(cfg_exp1: dict):
...
+# def test_EvalSetup_INVALID_ENTRY_NAMES(cfg_exp1: dict, error: str):
+#     # with pytest.raises(EvalEntryNameError) as e:
+#     stp = EvalSetup(**cfg_exp1)
+#     assert stp
+#     # assert error in str(e.value)
+
+
+#     with pytest.raises(EvalEntryNameError) as e:
+#         EvalSetup.model_validate(cfg_exp1)
+#     assert error in str(e.value)
```
dulte: Remove dead code. 

Is there a need for a similar test for the new implementation?

dulte (dismissed):
Looks great. Some dead code. But other than that, I think it can be merged

pyaerocom/aeroval/obsentry.py
```diff
@@ -146,67 +204,3 @@ def get_vert_code(self, var):
...
-        """
-        if ovt in self.ALT_NAMES_VERT_CODES:
-            _ovt = self.ALT_NAMES_VERT_CODES[ovt]
-            logger.warning(f"Please use {_ovt} for obs_vert_code and not {ovt}")
-            return _ovt
-        valid = self.SUPPORTED_VERT_CODES + list(self.ALT_NAMES_VERT_CODES)
-        raise ValueError(
-            f"Invalid value for obs_vert_type: {self.obs_vert_type}. "
-            f"Supported codes are {valid}."
-        )
```
dulte, line 212: Have you made any validator that does the same as this function?

lewisblake: Implemented now :+1:

pyaerocom/aeroval/collections.py
```diff
@@ -148,10 +146,7 @@ def get_web_iface_name(self, key):
...
 
         """
-        entry = self[key]
-        if "web_interface_name" not in entry:
-            return key
-        return entry["web_interface_name"]
+        return self[key].web_interface_name if self[key].web_interface_name is not None else key
 
     @property
     def web_iface_names(self) -> list:
```
thorbjoernl: Should also be `get_web_interface_name` to be consistent

pyaerocom/aeroval/setup_classes.py
```diff
@@ -528,8 +528,9 @@ def serialize_model_cfg(self, model_cfg: ModelCollection):
     ##       Methods
     ###########################
 
-    def get_obs_entry(self, obs_name) -> dict:
-        return self.obs_cfg.get_entry(obs_name).to_dict()
+    def get_obs_entry(self, obs_name):
```
thorbjoernl: Add typehints?

pyaerocom/colocation/colocation_setup.py
```diff
@@ -320,7 +320,7 @@ class ColocationSetup(BaseModel):
     @classmethod
     def validate_obs_vars(cls, v):
         if isinstance(v, str):
```
thorbjoernl, line 322: Maybe I just don't understand what these pydantic validators are supposed to do but shouldn't this raise an exception for unexpected input? Right now it just returns the value?

The example in field_validator docs raises ValueError for unexpected input.

pyaerocom/aeroval/obsentry.py
```diff
@@ -133,80 +227,15 @@ def has_var(self, var_name):
 
     def get_vert_code(self, var):
```
thorbjoernl, line 228: typehint?

pyaerocom/aeroval/obsentry.py
```diff
@@ -133,80 +227,15 @@ def has_var(self, var_name):
...
         """Get vertical code name for obs / var combination"""
-        vc = self["obs_vert_type"]
+        vc = self.obs_vert_type
         if isinstance(vc, str):
             val = vc
         elif isinstance(vc, dict) and var in vc:
             val = vc[var]
         else:
             raise ValueError(f"invalid value for obs_vert_type: {vc}")
-        if val not in self.SUPPORTED_VERT_CODES:
```
thorbjoernl, line 143: Isn't this in principle impossible if pydantic validation is working correctly? We should trust the validation.

pyaerocom/aeroval/obsentry.py
```diff
@@ -101,22 +197,20 @@ def check_add_obs(self):
...
-                    const.add_ungridded_post_dataset(**self)
+                    const.add_ungridded_post_dataset(**self.model_dump())
                 except Exception:
                     raise InitialisationError(
                         f"Cannot initialise auxiliary reading setup for {self.obs_id}. "
                         f"Reason:\n{format_exc()}"
                     )
 
-    def get_all_vars(self) -> list:
+    def get_all_vars(self) -> tuple[str, ...]:
```
thorbjoernl, line 207: Typehint should be `str | tuple[str, ...]`

pyaerocom/aeroval/_processing_base.py
```diff
@@ -127,7 +104,7 @@ def get_colocator(self, model_name: str = None, obs_name: str = None) -> Colocat
             mod_cfg = self.cfg.get_model_entry(model_name)
             col_cfg["model_cfg"] = mod_cfg
 
-            # LB: Hack and at what lowlevel_helpers's import_from was doing
+            # Hack and at what lowlevel_helpers's import_from was doing
```
heikoklein, line 107: I don't understand this comment

pyaerocom/aeroval/_processing_base.py
```diff
@@ -136,12 +113,12 @@ def get_colocator(self, model_name: str = None, obs_name: str = None) -> Colocat
             pyaro_config = obs_cfg["obs_config"] if "obs_config" in obs_cfg else None
             col_cfg["obs_config"] = pyaro_config
 
-            # LB: Hack and at what lowlevel_helpers's import_from was doing
-            for key, val in obs_cfg.items():
+            # Hack and at what lowlevel_helpers's import_from was doing
+            for key, val in obs_cfg.model_dump().items():
```
heikoklein, line 117: unclear/unnecessary comment?

pyaerocom/aeroval/collections.py
```diff
@@ -11,7 +11,7 @@ class BaseCollection(BrowseDict, abc.ABC):
     #: maximum length of entry names
     MAXLEN_KEYS = 25
     #: Invalid chars in entry names
-    FORBIDDEN_CHARS_KEYS = ["_"]
+    FORBIDDEN_CHARS_KEYS = []  # "_"
```
heikoklein: remove comment, we have git-history

pyaerocom/aeroval/obsentry.py
```diff
@@ -101,22 +197,20 @@ def check_add_obs(self):
                 )
```
thorbjoernl, line 197: L97: self.obs_type does not exist

pyaerocom/aeroval/collections.py
```diff
@@ -11,7 +11,7 @@ class BaseCollection(BrowseDict, abc.ABC):
...
 
     def _check_entry_name(self, key):
         if any([x in key for x in self.FORBIDDEN_CHARS_KEYS]):
```
thorbjoernl, line 18: `FORBIDDEN_CHARS_KEYS` is an empty list and supposed to be a constant. Is there any point in keeping this around?

thorbjoernl (changes requested):
Mostly nitpicking a bit, otherwise lgtm

pyaerocom/aeroval/obsentry.py
```diff
@@ -1,29 +1,58 @@
...
+
+SUPPORTED_VERT_LOCS: tuple[str, str, str] = (
+    "ground",
+    "space",
+    "airborne",
+)
+
+
+class ObsEntry(BaseModel):
+    """Observation configuration for evaluation (BaseModel)
```
heikoklein, line 41: Generally, I don't like it if the interface changes (here access to all attributes from ["XXX"] to .XXX) This might destroy existing code. Ok since obsentry is mostly internal.

pyaerocom/colocation/colocation_setup.py
lewisblake: `obs_vars` can be either a string or tuple of strings. What this is doing is coercing a provided string into a tuple of strings (for consistency). If it is not a string or a tuple of strings, then pydantic raises an error immediately.

pyaerocom/aeroval/obsentry.py
```diff
@@ -101,22 +197,20 @@ def check_add_obs(self):
...
         Get list of all variables associated with this entry
 
         Returns
         -------
-        list
-            DESCRIPTION.
-
+        tuple[str, ...]
         """
         return self.obs_vars
```
heikoklein, line 215: unnecessary change of signature
```
return list(self.obs_vars)
```
would have worked as well, and didn't require the change to `(v,)` tuple code. A tuple[...] is in most cases better a list from a user-point perspective.

pyaerocom/aeroval/obsentry.py
lewisblake: Part of the validator of `obs_vars` is that if you provide `ObsEntry` with a string, it get coerced into a tuple with one element, so this method will only return a tuple containing (a) string(s)

heikoklein (approved):
OK-ish.

This PR breaks the pyaerocom interface at several places. It does not try to create a interoperability with the old interface. What it tries, i.e. introduction of immutable code, could most likely be achieved without all of these breaks, i.e. by introducing `__getitem__`.

Since the broken parts are mostly internal, I approve.

Some of the comments don't make any sense to me, so maybe just remove them. Easier to understand code than comments.

pyaerocom/aeroval/obsentry.py
lewisblake: Added. I agree could be useful. Originally I wanted to not have to import `ObsEntry` just for a type hint but I guess that's okay.

pyaerocom/aeroval/obsentry.py
lewisblake: Yes, this is leftover from the old "validators". The idea was that if it does not exist, it will crash at this step, which will tell a user that if they provided `obs_aux_requires` then they need to explcitly provide `obs_type`, and that if `obs_type` is not "ungridded", that currently it is not supported. We could add `obs_type` as an attribute with a default of "ungridded", but I'm concerned that people might forget to set it, and it could lead to unintended behavior. What do you think?

pyaerocom/aeroval/obsentry.py
lewisblake: This is part of an effort to make objects in the code immutable which should not change at run time, like has been changed elsewhere.

pyaerocom/aeroval/obsentry.py
thorbjoernl: ~~Should obs_vars be typehinted as just tuple then?~~

pyaerocom/aeroval/obsentry.py
lewisblake: Agree that where the user interacts with the code, we should try to minimize the changes to the interface. But also I think it's hard to provide structure to the code if everything is fundamentally a `BrowseDict` without everything becoming very nested. There's a balance there, and I think this is a good place to move away from the older dictionary-style accessing.

pyaerocom/colocation/colocation_setup.py
thorbjoernl: I see, so the typehint on obs_vars is what is accepted as input in a sense? Is there a way to make vscode use the narrower set of hints that the variable actually contains after coercion? Because right now a lot of the tooling struggles with assuming that the types can be any of the original types.

pyaerocom/aeroval/obsentry.py
lewisblake: No, because that is the type hint of what can be provided, not it's internal representation / what gets returned. If we type hint it as just a tuple, users have to always provide a tuple of strings, which would break the interface.

pyaerocom/aeroval/_processing_base.py
lewisblake: This has been in the code for many months now, just removing my initials. It's saying that the following lines of code were taken from a function which was broken apart.

pyaerocom/aeroval/obsentry.py
thorbjoernl: Can we have `obs_type` default to `None` and check that it is None instead? Then the user has to assign it and it doesn't appear like a bug?

lewisblake: Sure, added.

pyaerocom/aeroval/obsentry.py
lewisblake: This is a deleted line of code. However, the validator did need work, so I've added "ModelLevel" to the Literal pydantic will check against, and if provided, switch it to "Profile" with a logger warning. The way these model validators work though is that you have to deal with the input in any case, and return something.

pyaerocom/aeroval/obsentry.py
thorbjoernl: That's fair but this breaks a lot of the tooling around types such as autocomplete since vscode doesn't know what the real type of the attributes are.

Maybe we can look into doing something like this at a later point: https://github.com/pydantic/pydantic/discussions/4725#discussioncomment-4337576

pyaerocom/colocation/colocation_setup.py
lewisblake: > I see, so the typehint on obs_vars is what is accepted as input in a sense? 

Yes

>  Is there a way to make vscode use the narrower set of hints that the variable actually contains after coercion? Because right now a lot of the tooling struggles with assuming that the types can be any of the original types.

I'm not sure.

pyaerocom/aeroval/collections.py
lewisblake: I would say for the time being yes while I test this with colocated data I've been provided. If it goes smoothly we can remove. I will add a TODO

pyaerocom/aeroval/_processing_base.py
lewisblake: Same as above.

pyaerocom/aeroval/obsentry.py
heikoklein: The intention is to make obsentry immutable, which my example would also satisfy.

lewisblake: I agree making `ObsEntry` immutable is great. Now implemented.

thorbjoernl (approved):
LGTM :goat:

pyaerocom/aeroval/obsentry.py
lewisblake: Sadly, not as easy as I thought, but worth coming back to. #1367

PR #1397, opened 2024-11-08
test VarWebScaleAndColormap with wrong input

### LINKED ISSUE #1291: Arguments given in  VarWebScaleAndColormap may not be going through pydantic
**Describe the bug**

There are two `__init__` methods in this class. Only the second one is being used. This means that the pydantic checking is not being used. One way to test this is to try and give `VarWebScaleAndColormap` arguments which are not `scale` and `colmap`

 - Pyaerocom version: 0.21.dev0
 - Computing platform: All

Relevant PR: #1256

heikoklein:
## Change Summary

testing VarWebScaleAndColormap to fail on wrong input

## Related issue number
fix #1291

## Checklist

* [ ] Start with a draft-PR
* [x] Tests pass locally
* [x] Tests pass on CI
* [x] Make PR ready to review

### DIFF
--- a/pyaerocom/aeroval/glob_defaults.py
+++ b/pyaerocom/aeroval/glob_defaults.py
@@ -34,13 +34,14 @@ class ScaleAndColmap(dict[str, str | list[float]]):
 class VarWebScaleAndColormap(dict[str, ScaleAndColmap]):
     def __init__(self, config_file: str = "", **kwargs):
         """This class contains scale and colmap informations and is implemented as dict to allow
-        json serialization. It reads it inital data from data/var_scale_colmap.ini.
+        json serialization. It reads it inital data from data/var_scale_colmap.ini. kwargs will be send to update.
 
         :param config_file: filename to additional or updated information, defaults to None
         """
         super().__init__()
         with resources.path("pyaerocom.aeroval.data", "var_scale_colmap.ini") as file:
             self.update_from_ini(file)
+        config_file = kwargs.pop("config_file", config_file)
         if config_file != "":
             logger.info(f"Reading additional web-scales from '{config_file}'")
             if not os.path.exists(config_file):
@@ -51,6 +52,9 @@ class VarWebScaleAndColormap(dict[str, ScaleAndColmap]):
         self.update(**kwargs)
 
     def update(self, **kwargs):
+        """update/add scale and colormaps by kwargs, e.g.
+        update(concso2={"scale"=[0.2,1], "colormap"="bluewhite"})
+        """
         wvsc = _VarWebScaleAndColormap(scale_colmaps=kwargs)
         super().update(**{x: y._asdict() for x, y in wvsc.scale_colmaps.items()})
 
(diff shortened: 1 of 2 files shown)


pyaerocom/aeroval/glob_defaults.py
lewisblake, line 1: I have to admit I find the use of helper classes in this case unnecessarily convoluted and confusing. I think we could accomplish something similar more efficiently making more obvious use of `BaseModel`. For example, here is some untested code which communicates the idea but uses 2/3 as many lines of code.


```python
class ScaleAndColmap(BaseModel):
    scale: list[float] = Field(...)
    colmap: str = Field(...)


class VarWebScaleAndColormap(BaseModel):
    scale_colmaps: dict[str, ScaleAndColmap] = Field(...)

    @classmethod
    def from_ini(cls, filename: str):
        cfg = ConfigParser()
        if not os.path.exists(filename):
            raise FileNotFoundError(
                f"{cls.__name__} from_ini('{filename}') which does not exist"
            )
        cfg.read(filename)
        # remove configparser default
        cfg_dict = dict()
        keys = ("scale", "colmap")
        for s in cfg.sections():
            items = []
            for k in keys:
                if k in cfg[s]:
                    if k == "scale":
                        try:
                            items.append(json.loads(cfg[s][k]))
                        except Exception as ex:
                            raise KeyError(
                                f"wrong value for '{k}' of var '{s}' in {filename}: {ex}"
                            )
                    else:
                        items.append(cfg[s][k])
                else:
                    raise KeyError(f"missing '{k}' for var '{s}' in {filename}")
            cfg_dict[s] = {"scale": items[0], "colmap": items[1]}
        return cls(scale_colmaps=cfg_dict)

    def update_from_ini(self, filename: str):
        updated_model = self.from_ini(filename)
        self.scale_colmaps.update(updated_model.scale_colmaps)

    def write(self, filename: str):
        """write the scales and colormaps to a configuration file

        :param filename: path to a filename, currently a .ini file
        """
        cfg = ConfigParser()
        cfg.read_dict(self.model_dump())
        with open(filename, "w") as fh:
            cfg.write(fh)
            
```

lewisblake:
```python
# usage
vwsac = VarWebScaleAndColormap.from_ini("var_scale_colmap.ini")
vwsac.update_from_ini("other_file.ini")
vwsac.write("output.ini")
```

tests/aeroval/test_glob_defaults.py
lewisblake, line 1: Thanks for adding these tests!

pyaerocom/aeroval/glob_defaults.py
heikoklein: pydantic.BaseModel is a tool, not a solution. Not using BaseModel as an exposed part of the API is a clean design choice. We might switch to another validator at some time, which will be hard if a user relies on anything from BaseModel.

lewisblake (approved):

PR #1406, opened 2024-11-15
Some work for reading ColocatedData provided by CAMS2_82 partners

lewisblake:
## Change Summary

- If `var_name_input` not in `coldata.metadata`, first attempt to get it from `obs_vars` before calling UNDEFINED. Raise a warning.
- Fix `json_repr` for `BaseCollection`

NB: the first of these is a hack which could be removed if a checker is provided to project partners

## Checklist

* [x] Start with a draft-PR
* [x] Tests pass locally
* [x] Tests pass on CI
* [x] Make PR ready to review

### DIFF
--- a/pyaerocom/aeroval/collections.py
+++ b/pyaerocom/aeroval/collections.py
@@ -7,8 +7,6 @@ from pyaerocom.exceptions import EntryNotAvailable
 
 
 class BaseCollection(abc.ABC):
-    IGNORE_JSON = []
-
     def __init__(self):
         """
         Initialize an instance of BaseCollection.
@@ -111,9 +109,9 @@ class BaseCollection(abc.ABC):
         """
         return self.keylist()
 
-    def json_repr(self) -> dict:
+    def as_dict(self) -> dict:
         """
-        Convert object to serializable json dict
+        Convert object to serializable dict
 
         Returns
         -------
@@ -123,8 +121,6 @@ class BaseCollection(abc.ABC):
         """
         output = {}
         for key, val in self._entries.items():
-            if key in self.IGNORE_JSON:
-                continue
             if hasattr(val, "json_repr"):
                 val = val.json_repr()
             output[key] = val
--- a/pyaerocom/aeroval/collections.py
+++ b/pyaerocom/aeroval/collections.py
@@ -1,6 +1,5 @@
 import abc
 from fnmatch import fnmatch
-import json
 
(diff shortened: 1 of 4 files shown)


pyaerocom/aeroval/collections.py
```diff
@@ -110,9 +111,24 @@ def web_interface_names(self) -> list:
         """
         return self.keylist()
 
-    def to_json(self) -> str:
-        """Serialize ModelCollection to a JSON string."""
-        return json.dumps({k: v.dict() for k, v in self._entries.items()}, default=str)
+    def json_repr(self) -> dict:
```
heikoklein: I would call this function `as_dict` or `to_dict`. That dicts are easier to use with the json-module is only implicit.

pyaerocom/aeroval/collections.py
```diff
@@ -110,9 +111,24 @@ def web_interface_names(self) -> list:
...
+        """
+        Convert object to serializable json dict
+
```
heikoklein, line 115: you return a python dict, not a json dict.

pyaerocom/aeroval/setup_classes.py
```diff
@@ -511,7 +511,7 @@ def obs_cfg(self) -> ObsCollection:
 
     @field_serializer("obs_cfg")
     def serialize_obs_cfg(self, obs_cfg: ObsCollection):
-        return obs_cfg.to_json()
+        return obs_cfg.json_repr()
```
heikoklein: serialization is something as a string, you get a dict here.

pyaerocom/aeroval/setup_classes.py
```diff
@@ -523,7 +523,7 @@ def model_cfg(self) -> ModelCollection:
 
     @field_serializer("model_cfg")
     def serialize_model_cfg(self, model_cfg: ModelCollection):
-        return model_cfg.to_json()
+        return model_cfg.json_repr()
```
heikoklein: same as above

heikoklein (approved):
The part for the ColocatedData is fine.
I don't understand the BaseCollection: It is something which behaves like a dict and should serialize like a dict, but you try hard not to use a dict as a baseclass and rather put everything into a dict which is hidden?
Not sure how the BaseCollection changes correspond to the title and the rest of this PR.
Approving anyway since the code looks good even if I don't understand the concept.

PR #1410, opened 2024-11-18
Lowlevel helpers clean up

lewisblake:
## Change Summary

- Implement `_AuxReadSpec` in terms of pydantic
- Consequently, remove unused `Validator` child classes `StrWithDefault`, `FlexList`, `EitherOf`, `ListOfStrings`

## Related issue number

N/A 

## Checklist

* [x] Start with a draft-PR
* [x] Tests pass locally
* [x] Tests pass on CI
* [x] Make PR ready to review

### DIFF
--- a/pyaerocom/aeroval/aux_io_helpers.py
+++ b/pyaerocom/aeroval/aux_io_helpers.py
@@ -66,9 +66,10 @@ class _AuxReadSpec(BaseModel):
     @model_validator(mode="after")
     def validate_fun(self) -> None:
         if callable(self.fun):
-            return
+            return self
         elif isinstance(self.fun, str):
             self.fun = self.funcs[self.fun]
+            return self
         else:
             raise ValueError("failed to retrieve aux func")
 
--- a/pyaerocom/aeroval/aux_io_helpers.py
+++ b/pyaerocom/aeroval/aux_io_helpers.py
@@ -8,6 +8,8 @@ if sys.version_info >= (3, 11):
 else:
     from typing_extensions import Self
 
+from typing import TYPE_CHECKING
+
 from pydantic import (
     BaseModel,
     model_validator,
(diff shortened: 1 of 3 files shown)


pyaerocom/aeroval/aux_io_helpers.py
```diff
@@ -53,39 +59,19 @@ class _AuxReadSpec:
...
-            return fun
-        elif isinstance(fun, str):
-            return funcs[fun]
-        raise ValueError("failed to retrieve aux func")
+    fun: str | Callable
+    vars_required: list[str]
+    funcs: dict[str, Callable]
+
+    @model_validator(mode="after")
+    def validate_fun(self) -> None:
```
thorbjoernl: Return type should be `_AuxReadSpec`

pyaerocom/aeroval/aux_io_helpers.py
```diff
@@ -53,39 +59,19 @@ class _AuxReadSpec:
...
-        callable
-            callable function object.
-
-        """
-        if callable(fun):
-            return fun
-        elif isinstance(fun, str):
-            return funcs[fun]
-        raise ValueError("failed to retrieve aux func")
+    fun: str | Callable
```
thorbjoernl: Is it possible to use something like [this](https://github.com/pydantic/pydantic/discussions/4823) to make `fun` typehint as str only during typechecking, since that's what we expect it to be after validation? I think this would work better with tools such as mypy and code autocompletion.

pyaerocom/aeroval/aux_io_helpers.py
lewisblake: Can return `Self`

pyaerocom/aeroval/aux_io_helpers.py
lewisblake: Actually the opposite. After validation we expect it to be a Callable. But you can give it a string and `funcs` and a str -> Callable mapping which will set `fun` to be the value associated to the key given as input.

thorbjoernl (approved):

Issue #1414, opened 2024-11-20
Configuration with pyaro repeats name a lot


magnusuMET:
The configuration when using `pyaro` has needlessy many occurences of the same name, as seen in the following configuration for a run using the EEA pyaro-reader.
```json
{
  "obs_cfg": {
    "EEA-unverified": {
      "obs_id": "EEA-unverified",
      "obs_config": {
        "name": "EEA-unverified",
        "data_id": "eeareader",
      },
      "obs_name": "EEA-unverified"
    }
}
```

We should find a new config setup before pyaro is used in too many places

lewisblake:
Agreed some clean up would be nice. Here is a short explanation of what things currently mean
```python
{
  "obs_cfg": {
    "EEA-unverified": { # this is user-defined and can be anything you want. shown on webpage if `var_web_interface` not defined
      "obs_id": "EEA-unverified", # this is required by ObsEntry. In vanilla pyaerocom, maps to a obs network defined in paths.ini
      "obs_config": { # pyaro stuff, name matching obs_id might be redundant, probably should talk with Daniel
        "name": "EEA-unverified",
        "data_id": "eeareader", # maybe pyaro_id would be clearer
      },
      "obs_name": "EEA-unverified" # this get created when using a getter on a ObsEntry instance. Either is the var_web.interface name if provided, or it's the key supplied by the user
    }
}
```

lewisblake:
Can we close this now or is there additional renaming which would make things more clear?

PR #1428, opened 2024-11-26
Rename pyaro related fields

magnusuMET:
## Change Summary

It might be confusing having `obs_config` nested in `obs_cfg`, so I suggest we rename this inner field to the more descriptive `pyaro_config`.

`data_id` is also not very descriptive, `reader_id` might give a better idea that it is related to the entrypoint in `pyaro`?

## Related issue number

Related to #1414 

## Checklist

* [x] Start with a draft-PR
* [x] Tests pass locally
* [x] Tests pass on CI
* [x] Make PR ready to review

### DIFF
--- a/pyaerocom/colocation/colocation_setup.py
+++ b/pyaerocom/colocation/colocation_setup.py
@@ -69,7 +69,7 @@ class ColocationSetup(BaseModel):
     model_id : str
         ID of model to be used.
 
-    obs_config: PyaroConfig
+    pyaro_config: PyaroConfig
         In the case Pyaro is used, a config must be provided. In that case obs_id(see below)
         is ignored and only the config is used.
     obs_id : str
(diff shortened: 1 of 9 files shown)


pyaerocom/aeroval/setup_classes.py
```diff
@@ -357,6 +358,8 @@ class EvalSetup(BaseModel):
 
     _aux_funs: dict = {}
 
+    obs_pyaro_readers: list[PyaroConfig] | None = None
```
lewisblake: Could this be `tuple[PyaroConfig, ...] | None = None`

thorbjoernl (commented):
Have you checked if any of the base_configs need to be changed as a result of this change?

magnusuMET:
> Have you checked if any of the base_configs need to be changed as a result of this change?

Grepping for `filename_or_obj_or_url`, which is a required field for `pyaro`, shows no base_configs have started using `pyaro`

thorbjoernl (approved):
:rocket:

PR #1430, opened 2024-11-26
Obs-only: check model_list to see if empty instead

lewisblake:
## Change Summary

Change `if not self.cfg.model_cfg` in experiment_processor.py to instead look at `model_list`, which should evaluate to False if empty. Note `model_list` is created for every `model_name`, but will always be empty if no models are provided.

## Related issue number

N/A, chat with @dulte 

## Checklist

* [x] Start with a draft-PR
* [x] Tests pass locally
* [x] Tests pass on CI
* [x] Make PR ready to review

### DIFF
--- a/pyaerocom/aeroval/collections.py
+++ b/pyaerocom/aeroval/collections.py
@@ -277,3 +277,29 @@ class ModelCollection(BaseCollection):
             return self._entries[key]
         else:
             raise EntryNotAvailable(f"no such entry {key}")
+
+    def keylist(self, name_or_pattern: str = None) -> list[str]:
+        """Find model / obs names that match input search pattern(s)
+
+        Parameters
+        ----------
+        name_or_pattern : str, optional
+            Name or pattern specifying search string.
+
+        Returns
+        -------
+        list
+            list of keys in collection that match input requirements. If
+            `name_or_pattern` is None, all keys will be returned.
+
+        Raises
+        ------
+        KeyError
+            if no matches can be found
+        """
+        # Special case where the model cfg is empty, used for obs-only
+        if list(self._entries.keys()) == []:
+            return []
+
+        else:
+            return super().keylist(name_or_pattern)
--- a/pyaerocom/aeroval/collections.py
+++ b/pyaerocom/aeroval/collections.py
@@ -87,6 +87,10 @@ class BaseCollection(abc.ABC):
         KeyError
             if no matches can be found
         """
+        # Special case where the model cfg is empty, used for obs-only
+        if self._entries.keys() == []:
+            return []
+
         if name_or_pattern is None:
             name_or_pattern = "*"
 
@@ -94,6 +98,9 @@ class BaseCollection(abc.ABC):
         for key in self._entries.keys():
             if fnmatch(key, name_or_pattern) and key not in matches:
                 matches.append(key)
+
+        if len(matches) == 0:
+            raise KeyError(f"No matches could be found that match input {name_or_pattern}")
         return matches
 
     @property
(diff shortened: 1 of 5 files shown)


pyaerocom/helpers.py
```diff
@@ -1704,18 +1726,22 @@ def get_max_period_range(periods):
...
 ) -> iris.cube.Cube:
-    startstr = f"days since {start_yr}-01-01 00:00"
+    startstr = f"{TS_TYPE_TO_FREQ_NAME[freq]} since {start_yr}-01-01 00:00"
 
     if freq not in TS_TYPE_TO_PANDAS_FREQ.keys():
         raise ValueError(f"{freq} not a recognized frequency")
 
     start_str = f"{start_yr}-01-01 00:00"
-    stop_str = f"{stop_yr}-12-31 00:00"
+    stop_str = f"{int(stop_yr)+1}-01-01 00:00"
```
thorbjoernl: Might be wrong but `pd.date_range` is inclusive on both ends by default, so I think this includes one more element than we want? https://pandas.pydata.org/docs/reference/api/pandas.date_range.html

pyaerocom/aeroval/collections.py
```diff
@@ -87,13 +87,18 @@ def keylist(self, name_or_pattern: str = None) -> list[str]:
         KeyError
             if no matches can be found
         """
+        # Special case where the model cfg is empty, used for obs-only
+        if self._entries.keys() == []:
```
thorbjoernl, line 91: Shouldn't this be on `ModelCollection`? Now it applies to both `ObsCollection` and `ModelCollection` since it's on the base class.

pyaerocom/aeroval/experiment_processor.py
```diff
@@ -132,19 +132,19 @@ def run(self, model_name=None, obs_name=None, var_list=None, update_interface=Tr
...
 
         obs_list = self.cfg.obs_cfg.keylist(obs_name)
-        if not self.cfg.model_cfg:
+        model_list = self.cfg.model_cfg.keylist(model_name)
+        if not model_list:
             logger.info("No model found, will make dummy model data")
             self.cfg.webdisp_opts.hide_charts = ["scatterplot"]
             self.cfg.webdisp_opts.pages = ["evaluation", "infos"]
             model_id = make_dummy_model(obs_list, self.cfg)
+            model_list.append("dummy")
```
thorbjoernl: What is the intent here? Does this relate to the preceeding line? Maybe a clarifying comment would be useful since this does not appear related to the preceeding lines?

thorbjoernl (changes requested):

pyaerocom/aeroval/experiment_processor.py
dulte: The line above adds the dummy to cfg, but not to the model_list, which is still empty. This list is used later. But I've added a comment

pyaerocom/aeroval/collections.py
dulte: Good observation, moved to ModelCollection

pyaerocom/helpers.py
dulte: True, this was added to bypass a bug (maybe) with reading the Cube. But I think I've fixed it, so this will be removed

pyaerocom/helpers.py
```diff
@@ -1738,7 +1738,7 @@ def make_dummy_cube(
         raise ValueError(f"{freq} not a recognized frequency")
 
     start_str = f"{start_yr}-01-01 00:00"
-    stop_str = f"{int(stop_yr)+1}-01-01 00:00"
+    stop_str = f"{int(stop_yr)}-12-31 00:00"
```
thorbjoernl:
```suggestion
    stop_str = f"{int(stop_yr)}-12-31 23:00"
```

thorbjoernl (approved):

thorbjoernl (approved):
Lite forslag til endring, ellers LGTM

PR #1434, opened 2024-11-28
only_model_maps=True & pixel maps fixes

### LINKED ISSUE #1426: runs with `only_model_maps=True` get a "WARNING:no output available for experiment.." warning which is misleading
**Describe the bug**
this condition https://github.com/metno/pyaerocom/blob/232a3c956b98fd1f54e2b73792dacfb244107ab2/pyaerocom/aeroval/experiment_output.py#L209    
is based on checking that the `map` subfolder is populated in the output folder https://github.com/metno/pyaerocom/blob/232a3c956b98fd1f54e2b73792dacfb244107ab2/pyaerocom/aeroval/experiment_output.py#L159

in the case of a run with `only_model_maps=True`, the only subfolder that will be populated is `contour`   
so this check will always fail and the rest of the operations in `update_interface` will never be executed, despite the output having been correctly written out to `contour`

**To Reproduce**
1. run an experiment with `only_model_maps=True` and a fresh output folder

**Expected behavior**
in the case of a `only_model_maps=True` it should still be possible to finish the run by writing out the cfg[..].json, menu.json and the rest of the stuff in update_interface

lewisblake:
## Change Summary

- Allows to run `only_model_maps=True` without having run a previous experiment. This allows us to just show maps in cases where collocation and computation of statistics is computationally prohibitive, such as can be the case in CAMS2_82. Requires many changes in the code to create `menu.json`, `ranges.json`, `regions.json`, and fill `ts` with the necessary model and obs data to be able to slide through time series. Overall `ExperimentOutput` had far too high of a reliance on the `map` directory, so this work attempts to generalize the processing to be able to work when this does not exist.
- fixes misleading warning from #1426
- pixel maps are fixed up to correct projection (ESPG: 3857)
- fixes "coarsest" showing up in final config json for `maps_freq`
- fixes validators for `plot_types` to expect sets
- Minor clean up / corrections where I saw them.

## Related issue number

Although not the primary reason this PR will close https://github.com/metno/pyaerocom/issues/1426

## Checklist

* [x] Start with a draft-PR
* [x] Tests pass locally
* [x] Tests pass on CI
* [x] Make PR ready to review

### DIFF
--- a/pyaerocom/aeroval/experiment_output.py
+++ b/pyaerocom/aeroval/experiment_output.py
@@ -338,7 +338,8 @@ class ExperimentOutput(ProjectOutput):
 
         return MapInfo(obs_network, obs_var, vert_code, mod_id, mod_var, time_period)
 
-    def _info_from_contour_dir_file(self, file: pathlib.PosixPath):
+    @staticmethod
+    def _info_from_contour_dir_file(file: pathlib.PosixPath):
         """
         Separate map filename into meta info on obs and model content
 
@@ -366,9 +367,9 @@ class ExperimentOutput(ProjectOutput):
 
         if len(spl) != 3:
             raise ValueError(
-                f"invalid map filename: {file}. Must "
+                f"invalid contour filename: {file}. Must "
                 f"contain exactly 2 underscores _ to separate "
-                f"obsinfo, vertical, model info, and periods"
+                f"name, vertical, and periods"
             )
         name = spl[0]
         var_name = spl[1]
--- a/pyaerocom/aeroval/experiment_output.py
+++ b/pyaerocom/aeroval/experiment_output.py
@@ -327,6 +327,43 @@ class ExperimentOutput(ProjectOutput):
 
         return MapInfo(obs_network, obs_var, vert_code, mod_id, mod_var, time_period)
 
+    def _info_from_contour_dir_file(self, file: pathlib.PosixPath):
+        """
+        Separate map filename into meta info on obs and model content
+
+        Parameters
+        ----------
+        filename : str
+            name of file in "contour" subdirectory of json output directory for
+            this experiment
+
+        Raises
+        ------
+        ValueError
+            if input filename is invalid
+
+        Returns
+        -------
+        str
+            name of model
+        str
+            name of variable
+        str
+            Time period
+        """
+        spl = os.path.basename(file.name).split(file.suffix)[0].split("_")
+
+        if len(spl) != 3:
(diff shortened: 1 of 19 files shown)


pyaerocom/aeroval/experiment_output.py
```diff
@@ -766,9 +841,75 @@ def _is_part_of_experiment(self, obs_name, obs_var, mod_name, mod_var) -> bool:
...
+        else:
+            files = self._get_json_output_files("map")
         for file in files:
-            (obs_name, obs_var, vert_code, mod_name, mod_var, per) = self._info_from_map_file(file)
+            if self.cfg.processing_opts.only_model_maps:
+                # Hack to build menu.json
+                # Key issue we need to get around is that the ExperimentOutput class
+                # expects that at this point in the processing, all information it needs
+                # to describe an experiment has been written to disc, traditionally in the map directory
+                # if only_model_maps = True, then we do not do colocation, and so the maps dir is empty,
```
heikoklein: the dir is named `map` not `maps` (it is confusing enough that collocation-results are written to a directory called map...)

pyaerocom/aeroval/experiment_output.py
```diff
@@ -766,9 +841,75 @@ def _is_part_of_experiment(self, obs_name, obs_var, mod_name, mod_var) -> bool:
...
+                        continue
+                    obs_name = first_with_mod_name[0]
+                    all_combinations.remove(first_with_mod_name)
+                else:
+                    raise ValueError("Failed to infer vert_code in a only_model_maps experiment")
+
+            else:
+                (obs_name, obs_var, vert_code, mod_name, mod_var, per) = self._info_from_map_file(
+                    file
+                )
```
heikoklein, line 913: I agree that this looks very hacky, just to get `model_maps_only`. Since splitting the output-parts into smaller bits is very welcome from a point of performance, maybe the default of trying to do everything is wrong?

heikoklein (approved):
Thanks, looks good (one typo you might want to fix).

We should think about a different output-strategy rather than: "write everything except..." to improve parallelization, and avoid the current hacks, but that is rather long-term.

pyaerocom/aeroval/experiment_output.py
lewisblake: It is very hacky and I wish it didn't need to be, but I just don't see a way around it with out much bigger changes in our processing of experiments. I agree, that splitting things into smaller parts is preferred. Definitely we should think about a better approach.

PR #1445, opened 2024-12-05
Add post-processing of observations from pyaro

magnusuMET:
## Change Summary

Adds transformation of variables to go from e.g. `concno` to `concNno`, which were previously done using `AUX_FUNS`

## Checklist

* [x] Start with a draft-PR
* [x] Tests pass locally
* [x] Tests pass on CI
* [x] Make PR ready to review

## Longer explanation/design considerations
Models and observations should be compared using the same kind of species, which means observations might have to be transformed to match the model or for fair comparison to runs from previous years. There exists some functionality for this already in `pyaerocom`, but this must be implemented per reader. The functionality of this PR tries to generalise a bit, and implements this as a pyaro `Reader` adaptive layer (sort of middleware) like `VariableNameChanger`.

This functionality has been added to `pyaerocom` instead of `pyaro` as it is tied to the convention (name/units) of variables defined in `pyaerocom`.

Adding it as a separate reader in `pyaro-readers` turned out to make the configuration very verbose.

### DIFF
--- a/pyaerocom/io/pyaro/postprocess.py
+++ b/pyaerocom/io/pyaro/postprocess.py
@@ -1,10 +1,11 @@
 import dataclasses
 
-from cf_units import Unit
 from pyaro.timeseries import (
     Reader,
     Data,
 )
+
+from pyaerocom.units_helpers import get_unit_conversion_fac
 
 
 @dataclasses.dataclass
@@ -157,9 +158,9 @@ class PostProcessingReader(Reader):
             transform = self.compute_vars[varname]
             if isinstance(transform, VariableScaling):
                 data = self.reader.data(transform.REQ_VAR)
-                original_unit = Unit(data.units)
-                unit_scaling = original_unit.convert(1, transform.IN_UNIT)
-                scaling = unit_scaling * transform.SCALING_FACTOR
+                scaling = transform.SCALING_FACTOR * get_unit_conversion_fac(
+                    from_unit=data.units, to_unit=transform.IN_UNIT, var_name=transform.REQ_VAR
+                )
                 return PostProcessingReaderData(
                     data, variable=varname, units=transform.OUT_UNIT, scaling=scaling
                 )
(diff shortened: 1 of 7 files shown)


pyaerocom/io/pyaro/pyaro_config.py
```diff
@@ -34,6 +34,7 @@ class PyaroConfig(BaseModel):
     filename_or_obj_or_url: str | list[str] | Path | list[Path]
     filters: dict[str, FilterArgs]
     name_map: dict[str, str] | None = None  # no Unit conversion option
+    post_processing: dict[str, str] | None = None  # No post-processing
```
magnusuMET: Not sure about the naming of this variable, nor it being a dictionary. Could make more sense to have this be a list over variables which should be added (e.g. `post_processing=["concNno_from_concno", ...]`)

pyaerocom/io/pyaro/postprocess.py
```diff
@@ -0,0 +1,143 @@
...
+    REQ_VAR: str
+    IN_UNIT: str
+    OUT_UNIT: str
+    SCALING_FACTOR: float
+
+
+M_N = 14.006
+M_O = 15.999
+
+TRANSFORMATIONS = {
```
magnusuMET, line 31: These transformations should be documented somewhere

lewisblake:
Should talk at Aerotools meeting about transferring more of this workflow to pre-processing. Could create and store static data sets for specific project needs.

heikoklein (approved):
Thanks for the implementation, this is the right place to put it, virtually between pyaro-reading and pyaerocom usage. A few comments:

- Please check the usage of cf_units - not all pyaerocom-units are cf-compliant and @jgriesfeller added some special unit-functions to handle the non-SI units
- `obs_var_funs` documents that it covers the same functionality, but the documentation is wrong in https://pyaerocom.readthedocs.io/en/latest/_modules/pyaerocom/config.html . Can you correct the documentation, and say that it only covers merging of 2 variables?
- the conversion of vmro3max looks strange, and might be wrong when having daily data, but ok, usage is for the implementer.
- which reminds me that we need a good documentation of this usage under tutorials or in the readthedocs

Heiko

PR #1449, opened 2024-12-09
switch to pixelmaps for all models except IFS

charlienegri:
## Change Summary

switch to pixelmaps for all models except IFS
the speed-up compared to contour is enough for this to be desirable for production  
the IFS model domain is different tho, so the boundaries that hold for the rest of the models cannot be applied to it, and we must stick to contour maps for the IFS... 
this might create some confusion on the user side, perhaps the default should be overlay then, can this be configurable?

## Checklist

* [x] Start with a draft-PR
* [ ] The PR title is a good summary of the changes
* [x] Tests pass locally
* [x] Tests pass on CI
* [ ] Make PR ready to review

### DIFF
--- a/pyaerocom/scripts/cams2_83/cli.py
+++ b/pyaerocom/scripts/cams2_83/cli.py
@@ -78,6 +78,8 @@ def make_config(
         periods=eval_type.periods(start_date, end_date),
         json_basedir=str(data_path),
         coldata_basedir=str(coldata_path),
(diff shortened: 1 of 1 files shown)


charlienegri:
note: testing this code I noticed that menu.json is empty, this is not a problem for production but it is a 'bug' I guess for onlymap kinda runs that are supposed to be fully self-standing

lewisblake (approved):

PR #1470, opened 2025-01-06
Model maps with ColocatedData objects

lewisblake:
## Change Summary
Implements the functionality to make "model" maps from a ColocatedData object.

The way to do this is to 

- provide the `model_data_dir` in the model entry, ensuring that the path contains the model name at the end. (i.e., `"/lustre/storeB/users/lewisb/projects/CAMS2_82/coldata_testing2x2_180/ECMWF-OSUITE"`)
- provide the `coldata_dir` in the obs entry, ensuring that the model name *does not* appear at the end (i.e., "/lustre/storeB/users/lewisb/projects/CAMS2_82/coldata_testing2x2_180")
- set `add_model_maps=True` and optionally (depending on the resolution of the colocated data object) `only_model_maps=True`
- optionally set `pages = ("maps",)` if you only want to show maps
- declare which obserations and/or models should be shown in the right menu on the maps page. (e.g.,  right_menu = ("MODIS-TERRA", ))
- depending on the resolution you may want to use overlay maps. The preferred way to do this is to specify this for each model and observation network. (e.g., `plot_types = {"ECMWF-OSUITE": "overlay", "MODIS-TERRA": "overlay"}`)

image

## Checklist

* [x] Start with a draft-PR
* [ ] Tests pass locally
* [x] Tests pass on CI
* [x] Make PR ready to review

### DIFF
--- a/pyaerocom/aeroval/modelmaps_engine.py
+++ b/pyaerocom/aeroval/modelmaps_engine.py
@@ -1,9 +1,10 @@
+import glob
 import logging
 import os
 import xarray as xr
 
 
-from pyaerocom import GriddedData, TsType, const, __version__
+from pyaerocom import GriddedData, TsType, const, __version__, ColocatedData
 from pyaerocom.aeroval._processing_base import DataImporter, ProcessingEngine
 from pyaerocom.aeroval.modelmaps_helpers import (
     calc_contour_json,
@@ -236,15 +237,35 @@ class ModelMapsEngine(ProcessingEngine, DataImporter):
             var (str): variable name
         """
 
-        try:
-            data = self.read_gridded_obsdata(model_name, var)
-        except EntryNotAvailable:
+        if self.cfg.processing_opts.only_json:  # we have colocated data
             try:
-                data = self._read_model_data(model_name, var)
-            except Exception as e:
-                raise ModelVarNotAvailable(
-                    f"Cannot read data for model {model_name} (variable {var}): {e}"
+                preprocessed_coldata_dir = self.cfg.model_cfg.get_entry(model_name).model_data_dir
+                mask = f"{preprocessed_coldata_dir}/*.nc"
+            except KeyError:
+                preprocessed_coldata_dir = self.cfg.obs_cfg.get_entry(model_name).coldata_dir
+                mask = f"{preprocessed_coldata_dir}/{model_name}/*.nc"
+            file_to_convert = glob.glob(mask)
+            if len(file_to_convert) != 1:
+                raise ValueError(
+                    "Can only handle one colocated data object for plotting for a given (model, obs, var). "
+                    "Note that when providing a colocated data object, it must be provided via the model_data_dir arugment in a ModelEntry instance. "
+                    "It must also be provided via the coldata_dir argument in the ObsEntry instance. "
+                    "Additionally, note that the coldatadir does not contain the model_name at the end of the directory, "
+                    "whereas the coldata_dir does not."
                 )
+            coldata = ColocatedData(data=file_to_convert[0])
+            data = coldata.data.sel(data_source=model_name)
+            # data = GriddedData(data.to_iris())
+        else:
+            try:
+                data = self.read_gridded_obsdata(model_name, var)
+            except EntryNotAvailable:
+                try:
+                    data = self._read_model_data(model_name, var)
+                except Exception as e:
+                    raise ModelVarNotAvailable(
+                        f"Cannot read data for model {model_name} (variable {var}): {e}"
+                    )
 
         var_ranges_defaults = self.cfg.var_scale_colmap
 
@@ -255,7 +276,8 @@ class ModelMapsEngine(ProcessingEngine, DataImporter):
             cmapinfo = var_ranges_defaults["default"]
             varinfo = VarinfoWeb(var, cmap=cmapinfo["colmap"], cmap_bins=cmapinfo["scale"])
 
-        data = self._check_dimensions(data)
+        if not self.cfg.processing_opts.only_json:
+            data = self._check_dimensions(data)
 
         outdir = self.cfg.path_manager.get_json_output_dirs()["contour/overlay"]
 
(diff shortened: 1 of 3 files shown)


pyaerocom/aeroval/modelmaps_engine.py
```diff
@@ -236,16 +238,49 @@ def _process_overlay_map_var(self, model_name, var, reanalyse_existing):  # prag
...
-            data = self.read_gridded_obsdata(model_name, var)
-        except EntryNotAvailable:
+        if self.cfg.processing_opts.only_json:  # we have colocated data
             try:
-                data = self._read_model_data(model_name, var)
-            except Exception as e:
-                raise ModelVarNotAvailable(
-                    f"Cannot read data for model {model_name} (variable {var}): {e}"
+                preprocessed_coldata_dir = self.cfg.model_cfg.get_entry(model_name).model_data_dir
+                mask = f"{preprocessed_coldata_dir}/*.nc"
```
thorbjoernl: `preprocessed_coldata_dir` should probably be `glob.escape()`'d, just to be on the safe side (unless we are certain that directories we use will never include special characters: https://docs.python.org/3/library/glob.html#glob.escape).

pyaerocom/aeroval/modelmaps_engine.py
```diff
@@ -236,16 +238,49 @@ def _process_overlay_map_var(self, model_name, var, reanalyse_existing):  # prag
...
+                file_to_convert = glob.glob(mask)
+            except KeyError:
+                preprocessed_coldata_dir = self.cfg.obs_cfg.get_entry(model_name).coldata_dir
+                mask = f"{preprocessed_coldata_dir}/{model_name}/*.nc"
```
thorbjoernl: Same as above

pyaerocom/aeroval/modelmaps_helpers.py
```diff
@@ -143,3 +155,13 @@ def plot_overlay_pixel_maps(
     plt.close("all")
 
     return image
+
+
+def search_directory_recursively_for_netcdf_filenames_containing_strings(directory, strings):
+    matching_files = []
+    # Use glob to find all NetCDF files recursively
+    for nc_file in glob.iglob(os.path.join(directory, "**", "*.nc"), recursive=True):
```
thorbjoernl: Escape `directory`

pyaerocom/aeroval/modelmaps_engine.py
```diff
@@ -236,16 +238,49 @@ def _process_overlay_map_var(self, model_name, var, reanalyse_existing):  # prag
...
+                matching_files = (
+                    search_directory_recursively_for_netcdf_filenames_containing_strings(
```
thorbjoernl: Is there a more concise name possible? I'd at least remove directory from the name, as it is clear that a dir is searched from function args.

(Edit: Maybe this should be two functions, one that fetches nc-files, and one that filters them)

pyaerocom/aeroval/modelmaps_engine.py
```diff
@@ -236,16 +238,49 @@ def _process_overlay_map_var(self, model_name, var, reanalyse_existing):  # prag
...
+                    )
                 )
 
+                if len(matching_files) > 1:
+                    logger.info(
+                        f"Found more than one colocated data file for {model_name=} {var=}. Using the first one found - this theoretically should be consistent across files."
+                    )
+                file_to_convert = matching_files[:1]
+            if len(file_to_convert) != 1:
+                raise ValueError(
```
thorbjoernl: Is this documented anywhere outside of this exception?

pyaerocom/aeroval/modelmaps_engine.py
```diff
@@ -255,7 +290,8 @@ def _process_overlay_map_var(self, model_name, var, reanalyse_existing):  # prag
             cmapinfo = var_ranges_defaults["default"]
             varinfo = VarinfoWeb(var, cmap=cmapinfo["colmap"], cmap_bins=cmapinfo["scale"])
 
-        data = self._check_dimensions(data)
+        if not self.cfg.processing_opts.only_json:
+            data = self._check_dimensions(data)
 
         outdir = self.cfg.path_manager.get_json_output_dirs()["contour/overlay"]
```
thorbjoernl, line 266: Is this still needed?

pyaerocom/aeroval/modelmaps_engine.py
```diff
@@ -236,16 +238,53 @@ def _process_overlay_map_var(self, model_name, var, reanalyse_existing):  # prag
...
+        else:
+            try:
+                data = self.read_gridded_obsdata(model_name, var)
+            except EntryNotAvailable:
+                try:
+                    data = self._read_model_data(model_name, var)
+                except Exception as e:
+                    raise ModelVarNotAvailable(
+                        f"Cannot read data for model {model_name} (variable {var}): {e}"
+                    )
```
thorbjoernl: It looks like it may make sense to break this function up into two functions, one for the `only_json` path and one for the other. This simplifies the type that data can be (`DataArray` in one function and `GriddedData` in the other) which reduces repeated conditionals for type checking, and I'd argue makes it clearer what's happening. It increases code duplication but I think it is worth it for readability.

thorbjoernl (changes requested):

pyaerocom/aeroval/modelmaps_engine.py
thorbjoernl: (Looks like this covers for missing functionality in aerovaldb, so leave it for now)

pyaerocom/aeroval/modelmaps_engine.py
lewisblake: Changed to `find_netcdf_files`. It's a pretty niche function though, and I'd rather keep it all in one place since it's functionality is pretty small and limited

pyaerocom/aeroval/modelmaps_engine.py
lewisblake: Documented above

pyaerocom/aeroval/modelmaps_engine.py
lewisblake: Agreed it's kind of a mess. How about this: instead of having two separate functions, I moved the processing part for the `only_json` case into its own helper function. I think this cleans things up quite a bit.

pyaerocom/aeroval/modelmaps_engine.py
lewisblake: See comment below

thorbjoernl (approved):
LGTM

Issue #1513, opened 2025-02-10
fix broken paths, check and remove unused return(s) in the ModelMapsEngine


charlienegri:
this part of `_process_contour_map_var`  
```
        outdir = self.cfg.path_manager.get_json_output_dirs()["contour"]
        outname = f"{var}_{model_name}"
        fp_geojson = os.path.join(outdir, f"{outname}.geojson")
```
(https://github.com/metno/pyaerocom/blob/main-dev/pyaerocom/aeroval/modelmaps_engine.py#L207) is outdated in view of the new aerovaldb routes:   
https://github.com/metno/aerovaldb/blob/88bb6865ad810ab2f13cedb233c2c7ad600b5ae4/src/aerovaldb/jsondb/jsonfiledb.py#L69  
so the function's return  
```
return fp_geojson
``` 
(https://github.com/metno/pyaerocom/blob/main-dev/pyaerocom/aeroval/modelmaps_engine.py#L237) is probably never used?   

the return is put into `_files` here https://github.com/metno/pyaerocom/blob/main-dev/pyaerocom/aeroval/modelmaps_engine.py#L126,   
which then leads to this return 
https://github.com/metno/pyaerocom/blob/main-dev/pyaerocom/aeroval/modelmaps_engine.py#L154  
which then goes to this `all_files` object https://github.com/metno/pyaerocom/blob/main-dev/pyaerocom/aeroval/modelmaps_engine.py#L65  
which is never used, at all, and this return https://github.com/metno/pyaerocom/blob/main-dev/pyaerocom/aeroval/modelmaps_engine.py#L71 makes no sense to me

charlienegri:
the run function of the ModelMapsEngine (https://github.com/metno/pyaerocom/blob/main-dev/pyaerocom/aeroval/modelmaps_engine.py#L53) is called only once in the experiment processor, that I can see, and the return is never made use of   
```
        if self.cfg.webdisp_opts.add_model_maps:
            engine = ModelMapsEngine(self.cfg)

            if isinstance(
                self.cfg.modelmaps_opts.plot_types, dict
            ):  # There may be additional obs networks to compute "model" maps for
                model_list = list(set(model_list) and set(self.cfg.modelmaps_opts.plot_types))

            engine.run(model_list=model_list, var_list=var_list)
```
https://github.com/metno/pyaerocom/blob/main-dev/pyaerocom/aeroval/experiment_processor.py#L164

charlienegri:
[...]

PR #1519, opened 2025-02-12
fix broken paths and remove unused returns in ModelMapsEngine 

charlienegri:
## Change Summary

output paths are hardcoded in `_process_contour_map_var` and `_process_overlay_map_var` in order to be able to use the `reanalyse_existing=False` option (which implies checking that files have been produced)  

these paths were broken after recent aerovaldb changes and are now fixed, but the way of checking existing files, especially for `_process_contour_map_var` where the contours computation is not broken up timestep by timestep, is not very elegant... 
this can be improved at some point if https://github.com/metno/aerovaldb/issues/117 is tackled

the return of  various "files" object  has been removed because it is anyway broken as long as the paths are wrong
it does not appear to be used for anything else than raising exceptions that can be handled differently  

note: the behavior of the option `reanalyse_existing=False` is different for `_process_contour_map_var` and `_process_overlay_map_var` regarding the existence of files:   
for the latter the calculation is done time by time, i.e. `plot_overlay_pixel_maps` is invoked in a loop over timesteps, which allows for recalculating only the specific times missing among the output var_name_[time].webp   
for the former if even just one time is missing among the expected output var_name_[time].geojson files, all are recalculated  
changing this to be coherent with the overlays requires changing the behavior of `calc_contour_json`, which is beyond the scope of this PR

## Related issue number
try tackle #1513 

## Checklist

* [x] Start with a draft-PR
* [x] Tests pass locally
* [x] Tests pass on CI
* [x] Make PR ready to review

### DIFF
--- a/pyaerocom/aeroval/modelmaps_engine.py
+++ b/pyaerocom/aeroval/modelmaps_engine.py
@@ -175,7 +175,7 @@ class ModelMapsEngine(ProcessingEngine, DataImporter):
         try:
             data = self._read_model_data(model_name, var)
         except Exception as e:
-            raise VarNotAvailableError(
+            raise ModelVarNotAvailable(
                 f"Cannot read data for model {model_name} (variable {var}): {e}"
             )
 
--- a/pyaerocom/aeroval/modelmaps_engine.py
+++ b/pyaerocom/aeroval/modelmaps_engine.py
@@ -1,32 +1,31 @@
 import glob
 import logging
 import os
+
 import xarray as xr
 
-
-from pyaerocom import GriddedData, TsType, const, __version__, ColocatedData
+from pyaerocom import ColocatedData, GriddedData, TsType, __version__, const
 from pyaerocom.aeroval._processing_base import DataImporter, ProcessingEngine
+from pyaerocom.aeroval.json_utils import round_floats
 from pyaerocom.aeroval.modelmaps_helpers import (
-    calc_contour_json,
-    plot_overlay_pixel_maps,
-    _jsdate_list,
     CONTOUR,
     OVERLAY,
+    _jsdate_list,
+    calc_contour_json,
     find_netcdf_files,
+    plot_overlay_pixel_maps,
 )
-from pyaerocom.aeroval.json_utils import round_floats
-from pyaerocom.colocation.colocator import Colocator
-
 from pyaerocom.aeroval.varinfo_web import VarinfoWeb
+from pyaerocom.colocation.colocator import Colocator
 from pyaerocom.exceptions import (
     DataCoverageError,
     DataDimensionError,
     DataQueryError,
+    EntryNotAvailable,
     ModelVarNotAvailable,
     TemporalResolutionError,
     VariableDefinitionError,
     VarNotAvailableError,
-    EntryNotAvailable,
 )
 
 logger = logging.getLogger(__name__)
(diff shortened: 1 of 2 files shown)


charlienegri:
@lewis could you try this branch with some of your overlay map use cases including the only_model_maps=True at some point?  (or then point me at how to do that)

lewisblake:
`/lustre/storeB/users/lewisb/Python/Evaluations/aeroval/cfg_pixel_maps_testing.py`

charlienegri:
> `/lustre/storeB/users/lewisb/Python/Evaluations/aeroval/cfg_pixel_maps_testing.py`

not world readable

tests/aeroval/test_modelmaps_engine.py
```diff
@@ -34,11 +36,41 @@ def test__run(caplog, cfg: dict):
...
+    cfg["periods"] = ["20100615"]
+    cfg["main_freq"] = "daily"
+
+    # create expected geojson output file (empty file is ok:
+    # in the case reanalyse_existing=False content is not checked, only existence)
+    json_basedir = tmp_path / "data"
+    output_file = (
+        json_basedir
+        / "test/exp1/contour/od550aer_TM5-AP3-CTRL/od550aer_TM5-AP3-CTRL_1277942400000.geojson"
+    )
```
thorbjoernl: Consider using aerovaldb to write this file. It will ensure that this doesn't break when aerovaldb updates.

Something roughly like:
```python
with aerovaldb.open(f"json_files:{str(tmp_path / 'data'}") as db:
    db.put_contour("", "test", "exp1", "od550aer", "TM5-AP3-CTRL", timestep="1277942400000")
```

pyaerocom/aeroval/modelmaps_engine.py
```diff
@@ -322,8 +315,7 @@ def _process_overlay_map_var(self, model_name, var, reanalyse_existing):  # prag
                     write_var_name,
                     date,
                 )
-            files.append(fp_overlay + "." + self.cfg.modelmaps_opts.overlay_save_format)
-        return files
+        return
```
thorbjoernl: Unneeded return

pyaerocom/aeroval/modelmaps_engine.py
```diff
@@ -233,8 +227,7 @@ def _process_contour_map_var(self, model_name, var, reanalyse_existing):  # prag
                     model_name,
                     timestep=time,
                 )
-
-        return fp_geojson
+        return
```
thorbjoernl: Unneeded return

pyaerocom/aeroval/modelmaps_engine.py
```diff
@@ -53,22 +52,21 @@ def _get_run_kwargs(self, **kwargs):
...
             try:
-                files = self._run_model(model, var_list)
+                self._run_model(model, var_list)
             except VarNotAvailableError:
-                files = []
-            if not files:
+                success = False
+            if not success:
                 logger.warning(f"no data for model {model}, skipping")
                 continue
```
thorbjoernl: Can the logging and continue not simply be part of the except clause? I don't think the success variable is necessary.

pyaerocom/aeroval/modelmaps_engine.py
```diff
@@ -53,22 +52,21 @@ def _get_run_kwargs(self, **kwargs):
...
-            all_files.extend(files)
 
         self.cfg.modelmaps_opts.maps_freq = (
             self._get_maps_freq()
         )  # if needed, reassign "coarsest" to actual coarsest frequency
 
-        return files
+        return
```
thorbjoernl: Unneeded return

pyaerocom/aeroval/modelmaps_engine.py
```diff
@@ -150,8 +138,7 @@ def _run_model(self, model_name: str, var_list):
                 if self.raise_exceptions:
                     raise
                 logger.warning(f"Failed to process maps for {model_name} {var} data. Reason: {e}.")
-
-        return files
+        return
```
thorbjoernl, line 136: Unneeded return

thorbjoernl (changes requested):

pyaerocom/aeroval/modelmaps_engine.py
charlienegri: yeah this was the previous logic, I kept it while debugging a failed test because I was unsure but now it can go :+1:

tests/aeroval/test_modelmaps_engine.py
charlienegri: good idea thanks

thorbjoernl (approved):
LGTM

Issue #1523, opened 2025-02-18
Add station type to stations in aeroval


heikoklein:
**Is your feature request related to a problem? Please describe.**
Fairmode plots should differ between rural and urban sites. This information exists for ebas and EEA observations and should be made availabe to aeroval to plot sites with different type in different shapes.

heikoklein:
The information for EEA stations is available in the csv-data as: /lustre/storeB/project/aerocom/aerocom1/AEROCOM_OBSDATA/EEA_AQeRep.v2/download/metadata.csv

```
Countrycode     Timezone        Namespace       AirQualityNetwork       AirQualityStation       AirQualityStationEoICode        AirQualityStationNatCode        SamplingPoint   SamplingProces  Sample  AirPollutantCode        ObservationDateBegin    ObservationDateEnd      Projection      Longitude       Latitude        Altitude        MeasurementType AirQualityStationType   AirQualityStationArea   EquivalenceDemonstrated MeasurementEquipment    InletHeight     BuildingDistance        KerbDistance

NO      http://dd.eionet.europa.eu/vocabulary/aq/timezone/UTC+01        NO.NILU.AQD     NET_NO002R      STA_NO0002R     NO0002R 812     SPO_NO0002R_1_1569      SPP_NO0002R_1_1569      SAM_NO0002R_1_1569      http://dd.eionet.europa.eu/vocabulary/aq/pollutant/1    2012-10-01T08:00:00             EPSG:4979       8.251900000000001       58.38845999952113       200     remote  background      rural-regional  yes     http://dd.eionet.europa.eu/vocabulary/aq/measurementequipment/unknown   3       -9900   -999
```

heikoklein:
For parquet files, the metadata is in /lustre/storeB/project/aerocom/aerocom1/AEROCOM_OBSDATA/EEA-AQDS/download/metadata.csv

```
[...]

PR #1525, opened 2025-02-20
station_type in output

charlienegri:
## Change Summary

- station_type read from a (basically hardcoded) metadata file is added to the UngriddedData object read via the ReadCAMS2_83 reader
- station_type is added to StationData if found in the ungridded data and written to the colocated data too
- station_type is also handled in the colocated data -> json pipeline so to land in the map files (where the fairmode stats are also written, for now) like this:
  ```
  {
        "station_name": "SK0066A",
        "station_type": "sub",
        "latitude": 49.119,
        "longitude": 18.325,
        "altitude": 265.0,
        "region": [
            "Slovakia"
        ],
        ...
  ```
- none of this is pretty and it probably breaks some things 

## Related issue number
tackles #1523 but it's worth keeping the possibility open that that issue could be solved by a look-up table kind of operation on the frontend side perhaps

## Checklist

* [x] Start with a draft-PR
* [ ] Tests pass locally
* [x] Tests pass on CI
* [x] Make PR ready to review

### DIFF
--- a/pyaerocom/aeroval/coldatatojson_helpers.py
+++ b/pyaerocom/aeroval/coldatatojson_helpers.py
@@ -490,7 +490,7 @@ def _init_site_coord_arrays(data):
             if "station_type" in cd.data.coords:
                 sites_types = cd.data.station_type.values
             else:
-                sites_types = [None] * len(sites)
+                sites_types = [""] * len(sites)
             lats = cd.data.latitude.values.astype(np.float64)
             lons = cd.data.longitude.values.astype(np.float64)
             if "altitude" in cd.data.coords:
--- a/pyaerocom/aeroval/coldatatojson_helpers.py
+++ b/pyaerocom/aeroval/coldatatojson_helpers.py
@@ -487,6 +487,10 @@ def _init_site_coord_arrays(data):
             continue
         elif not found:
             sites = cd.data.station_name.values
+            if "station_type" in cd.data.coords:
+                sites_types = cd.data.station_type.values
+            else:
+                sites_types = [None] * len(sites)
             lats = cd.data.latitude.values.astype(np.float64)
             lons = cd.data.longitude.values.astype(np.float64)
             if "altitude" in cd.data.coords:
@@ -502,7 +506,7 @@ def _init_site_coord_arrays(data):
         else:
             assert all(cd.data.station_name.values == sites)
         jsdates[freq] = cd.data.jsdate.values.tolist()
-    return (sites, lats, lons, alts, countries, jsdates)
+    return (sites, sites_types, lats, lons, alts, countries, jsdates)
 
 
 def _get_stat_regions(lats, lons, regions, **kwargs):
@@ -516,7 +520,7 @@ def _get_stat_regions(lats, lons, regions, **kwargs):
 
 def _process_sites(data, regions, regions_how, meta_glob):
     freqs = list(data)
-    (sites, lats, lons, alts, countries, jsdates) = _init_site_coord_arrays(data)
+    (sites, site_types, lats, lons, alts, countries, jsdates) = _init_site_coord_arrays(data)
     if regions_how == "country":
         regs = countries
     elif regions_how == "htap":
@@ -532,6 +536,7 @@ def _process_sites(data, regions, regions_how, meta_glob):
         # init empty timeseries data object
         site_meta = {
             "station_name": str(site),
+            "station_type": site_types[i],
             "latitude": lats[i],
             "longitude": lons[i],
             "altitude": alts[i],
--- a/pyaerocom/colocation/colocation_utils.py
+++ b/pyaerocom/colocation/colocation_utils.py
@@ -835,6 +835,7 @@ def colocate_gridded_ungridded(
     lats = [np.nan] * stat_num
     alts = [np.nan] * stat_num
     station_names = [""] * stat_num
+    station_types = [""] * stat_num
 
     data_ref_unit = None
     ts_type_src_ref = None
@@ -850,6 +851,7 @@ def colocate_gridded_ungridded(
         lats[i] = obs_stat.latitude
         alts[i] = obs_stat.altitude
         station_names[i] = obs_stat.station_name
+        station_types[i] = obs_stat.station_type
 
         # ToDo: consider removing to keep ts_type_src_ref (this was probably
         # introduced for EBAS were the original data frequency is not constant
@@ -977,6 +979,7 @@ def colocate_gridded_ungridded(
         "data_source": meta["data_source"],
         "time": time_idx,
         "station_name": station_names,
+        "station_type": ("station_name", station_types),
         "latitude": ("station_name", lats),
         "longitude": ("station_name", lons),
         "altitude": ("station_name", alts),
--- a/pyaerocom/colocation/colocation_utils.py
+++ b/pyaerocom/colocation/colocation_utils.py
@@ -851,7 +851,9 @@ def colocate_gridded_ungridded(
         lats[i] = obs_stat.latitude
         alts[i] = obs_stat.altitude
         station_names[i] = obs_stat.station_name
-        station_types[i] = obs_stat.station_type
+        station_types[i] = None
+        if hasattr(obs_stat, "station_type"):
+            station_types[i] = obs_stat.station_type
 
         # ToDo: consider removing to keep ts_type_src_ref (this was probably
         # introduced for EBAS were the original data frequency is not constant
--- a/pyaerocom/colocation/colocation_utils.py
+++ b/pyaerocom/colocation/colocation_utils.py
@@ -13,8 +13,8 @@ from geonum.atmosphere import pressure
 
 from pyaerocom import __version__ as pya_ver
 from pyaerocom import const
-from pyaerocom.climatology_config import ClimatologyConfig
 from pyaerocom._lowlevel_helpers import RegridResDeg
+from pyaerocom.climatology_config import ClimatologyConfig
 from pyaerocom.exceptions import (
     DataUnitError,
     DimensionOrderError,
@@ -859,9 +859,7 @@ def colocate_gridded_ungridded(
         lats[i] = obs_stat.latitude
         alts[i] = obs_stat.altitude
         station_names[i] = obs_stat.station_name
-        station_types[i] = None
-        if hasattr(obs_stat, "station_type"):
-            station_types[i] = obs_stat.station_type
+        station_types[i] = getattr(obs_stat, "station_type", "")
 
         # ToDo: consider removing to keep ts_type_src_ref (this was probably
         # introduced for EBAS were the original data frequency is not constant
@@ -978,7 +976,7 @@ def colocate_gridded_ungridded(
         "from_files": files,
         "from_files_ref": None,
         "colocate_time": colocate_time,
-        "obs_is_clim": True if isinstance(use_climatology_ref, ClimatologyConfig) else False,
+        "obs_is_clim": (True if isinstance(use_climatology_ref, ClimatologyConfig) else False),
         "pyaerocom": pya_ver,
         "min_num_obs": min_num_obs,
         "resample_how": resample_how,
--- a/pyaerocom/colocation/colocation_utils.py
+++ b/pyaerocom/colocation/colocation_utils.py
@@ -4,6 +4,7 @@ Methods and / or classes to perform colocation
 
 import logging
 import os
+from typing import TYPE_CHECKING
 
 import numpy as np
 import pandas as pd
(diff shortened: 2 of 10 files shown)


heikoklein (approved):
Looks good. 
Can you also update the tutorial on colocated dataobjects to include `station_type`?
Did you discuss with @AugustinMortier which `station_type`'s he want's to support, or if he supports all which we give him, i.e. everything which is in the obsdata?

charlienegri:
- to do: this PR should be tested with an experiment that does not read the station_type into the UngriddedData object, to see if the netcdf writing works
- if/once merged: this too should be merged once approved https://gitlab.met.no/alvarov/mos_reader/-/merge_requests/4

pyaerocom/aeroval/coldatatojson_helpers.py
```diff
@@ -487,6 +487,10 @@ def _init_site_coord_arrays(data):
             continue
         elif not found:
             sites = cd.data.station_name.values
+            if "station_type" in cd.data.coords:
+                sites_types = cd.data.station_type.values
+            else:
+                sites_types = [None] * len(sites)
```
avaldebe: What is the indented type? `list[str] | list[None]` or `list[str | None]`?
`None` represent a missing `station_type`, right?
Then `list[str] | list[None]` means that all `station_type` are present (`list[str]`) or all are missing (`list[None]`).

What about missing values in `cd.data.station_type`?

pyaerocom/colocation/colocation_utils.py
```diff
@@ -835,21 +834,24 @@ def colocate_gridded_ungridded(
     lats = [np.nan] * stat_num
     alts = [np.nan] * stat_num
     station_names = [""] * stat_num
+    station_types = [""] * stat_num
```
avaldebe, line 847: Which is the intended type of `station_types`? And for missing `station_types`?
`list[str]` with `""` for missing or `list[str | None]` with `None` for missing?

Missing `station_type` are represented as an empty string (`""`), on a different place it is represented as `None`.

pyaerocom/colocation/colocator.py
```diff
@@ -1035,7 +1036,14 @@ def _prepare_colocation_args(self, model_var: str, obs_var: str):
             min_num_obs=self.colocation_setup.min_num_obs,
             colocate_time=self.colocation_setup.colocate_time,
             resample_how=rshow,
+            add_meta_keys=[],
         )
+        # check if the station_type key has been passed to the ungridded data object
+        # (not all readers may do that, currently only the CAMS2_83 reader does)
+        if isinstance(obs_data, UngriddedData):
+            if all("station_type" in dict.keys() for dict in obs_data.metadata.values()):
+                args.update(add_meta_keys=["station_type"])
```
avaldebe: Is it possible for only some metadata have `station_type`?

pyaerocom/io/cams2_83/obs.py
```diff
@@ -71,4 +74,29 @@ def read_csv(
...
         logger.warning("found negative obs")
         df = df[df.conc > 0]
+    metadata_file_path = Path(path).parent.parent / DEFAULT_METADATA_NAME
+    if metadata_file_path.is_file():
+        df_metadata = read_metadata(metadata_file_path)
+        if not df_metadata.empty:
+            df = df.merge(df_metadata, on="station", how="left")
+            return df["station lat lon alt time poll conc station_type".split()]
+        else:
+            logger.warning(f"Empty metadata from {metadata_file_path}")
```
avaldebe: There is no return value here.
When `metadata_file_path.is_file()` and `df_metadata.empty` something when really wrong. Should it no be better to raise a ValueError?

pyaerocom/io/cams2_83/obs.py
```diff
@@ -71,4 +74,29 @@ def read_csv(
...
+    try:
+        df = pd.read_csv(
+            path,
+            sep=",",
+            header=0,
+            names="station station_type".split(),
+            usecols=[0, 2],
+        )
+    except ParserError as e:
+        logger.warning(f"Invalid metadata file {path}, {e}")
```
avaldebe, line 88: the empty data frame has columns, so `read_csv` is going to crash at
``` python
    return df["station lat lon alt time poll conc".split()]
```
I think that it is better to not mask the `ParserError` or raise a custom error.

avaldebe (changes requested):
Please look at my comments about the `sites_type` type and possible missing vaules

pyaerocom/colocation/colocator.py
charlienegri: not for the cams283 reader so far at least, in general the readers that will want to pass station_type to the ungriddeddata object should not be written such that they key can be there or not, values can be empty, but they key should be either there for all or none, right?

pyaerocom/io/cams2_83/obs.py
charlienegri: the return is in the line after, can be moved perhaps... but the idea was, in this case just do as before, no metadata, return the original unmerged df

pyaerocom/colocation/colocation_utils.py
charlienegri: not sure what's best but it should be coherent yeah

pyaerocom/aeroval/coldatatojson_helpers.py
charlienegri: the idea here was that at this stage we just treat whatever came from the ungriddeddata object at face-value, so we should add whatever the readers have claimed station_type is... 
the type should be string tho, agree, not sure where/how to force it in the general logic... 
they are strings at the level of the the cams283 reader, even if null  
I'll change from None to "" here for coherence with station_name and colocation_utils,py

pyaerocom/io/cams2_83/obs.py
charlienegri: mmh I thought df.empty would take care of this case, I have even written a test for it...   
originally the try/except block was outside the read_metadata function... 
I moved it here because I thought it was cleaner but I'll see how to improve this part

avaldebe: maybe I'm wrong, will double check

pyaerocom/colocation/colocator.py
avaldebe: If it is for the reader to implement, I would assume that it could be missing for some.

pyaerocom/colocation/colocation_utils.py
```diff
@@ -845,21 +844,24 @@ def colocate_gridded_ungridded(
...
     # loop over all stations and append to colocated data object
     for i, obs_stat in enumerate(obs_stat_data):
         # Add coordinates to arrays required for xarray.DataArray below
         lons[i] = obs_stat.longitude
         lats[i] = obs_stat.latitude
         alts[i] = obs_stat.altitude
         station_names[i] = obs_stat.station_name
+        station_types[i] = None
+        if hasattr(obs_stat, "station_type"):
+            station_types[i] = obs_stat.station_type
```
avaldebe: I think this could be written using [getattr](https://docs.python.org/3/library/functions.html#getattr)
``` python
        station_types[i] = getattr(obs_stat, "station_type", None)
```

tests/io/cams2_83/test_read_obs.py
```diff
@@ -39,3 +40,144 @@ def test_init():
...
+
+def test_obs_no_metadata_file(tmp_path, caplog):
+    tmp_file = tmp_path / "obs.csv"
+    tmp_file.write_text(
+        """
+            STATION;LAT;LON;ALT(m);PARAMETER;YEAR;MONTH;DAY;HOUR;AVERAGING_PERIOD(h);CONCENTRATION(kg/m3)
+            AT0ENK1;48.392; 13.671;0525;o3;2025;02;12;01;1; 4.27600e-08
+            AT0ILL1;47.770; 16.766;0117;o3;2025;02;12;01;1; 6.56700e-08
+    """
+    )
```
avaldebe: it looks like you could create the fake obs file on a fixture, as it will be used on several tests

pyaerocom/colocation/colocator.py
charlienegri: are you saying that you would prefer an `any` here?

pyaerocom/io/cams2_83/obs.py
```diff
@@ -71,4 +74,28 @@ def read_csv(
...
+        except ParserError as e:
+            logger.warning(f"Invalid metadata file {path}, {e}")
+            df_metadata = pd.DataFrame()
+        if not df_metadata.empty:
+            df = df.merge(df_metadata, on="station", how="left")
+            return df["station lat lon alt time poll conc station_type".split()]
+        else:
+            logger.warning("Empty metadata")
+    else:
+        logger.warning(f"Metadata file {metadata_file_path} does not exist")
```
avaldebe: The logic with/without metadata is a bit hard to follow, maybe refactoring with a guarding clauses will make it easier to follow

``` python
    df = df["station lat lon alt time poll conc".split()]

    metadata_file_path = Path(path).parent.parent / DEFAULT_METADATA_NAME
    if not metadata_file_path.is_file():
        logger.warning(f"Metadata file {metadata_file_path} does not exist")
        return df

    try:
        df_metadata = read_metadata(metadata_file_path)
    except ParserError as e:
        logger.warning(f"Invalid metadata file {path}, {e}")
        return df
    
    if df_metadata.empty:
       logger.warning("Empty metadata")
       return df
            
    return df.merge(df_metadata, on="station", how="left")
```

avaldebe (changes requested):
Thanks for addressing my typing comments.
Please consider my suggestions about improving readability and tests fixtures

pyaerocom/colocation/colocator.py
avaldebe: yes, I meant
``` python
            if any("station_type" in dict for dict in obs_data.metadata.values()):
                args.update(add_meta_keys=["station_type"])
```
but that would add complexity somewhere else

maybe it is possible a [match statement](https://docs.python.org/3.10/tutorial/controlflow.html#match-statements) (Python 3.10+) to raise a ValueError when the reader is inconsistent and deal with the problem of inconsistent readers if/when needed
``` python
            match {"station_type" in dict for dict in obs_data.metadata.values()}:
                case {True}: # all stations have `station_type`
                    args.update(add_meta_keys=["station_type"])
                case {True, False}:
                    raise ValueError("some stations have `station_type` metadata while others do not")
```

pyaerocom/io/cams2_83/obs.py
charlienegri: yeah sure, thanks for the suggestion, this reads better

pyaerocom/colocation/colocator.py
charlienegri: mmh sure I will consider this and test a bit

charlienegri: mmh I can't see how to make the match/case work with an iterable, I have added the ValueError as an elif case, which is pretty ugly... 
(but added a test so we still have 100% coverage)

pyaerocom/colocation/colocation_utils.py
```diff
@@ -976,7 +976,7 @@ def colocate_gridded_ungridded(
         "from_files": files,
         "from_files_ref": None,
         "colocate_time": colocate_time,
-        "obs_is_clim": True if isinstance(use_climatology_ref, ClimatologyConfig) else False,
+        "obs_is_clim": (True if isinstance(use_climatology_ref, ClimatologyConfig) else False),
```
avaldebe, line 979: `isinstance` returns a boolean, the `True if X else False` construct is redundant. You could write it like this

``` python
        "obs_is_clim": isinstance(use_climatology_ref, ClimatologyConfig),
```

pyaerocom/colocation/colocator.py
avaldebe: Looks like you wrote the `set` part twice

You wrote something like:
``` python
            match {
                set(
                    "station_type" in dict.keys() for dict in obs_data.metadata.values()
                )
            }:
            ...
```
instead of writing 
``` python
            match {"station_type" in dict for dict in obs_data.metadata.values()}:
            ...
```

avaldebe (commented):
sorry, I still have comments

pyaerocom/colocation/colocation_utils.py
charlienegri: this is not my code,,,  I must have messed up the rebase with main-dev, sorry.... 
this is from main-dev anyway

charlienegri:
> sorry, I still have comments

they are much appreciated!

pyaerocom/colocation/colocator.py
charlienegri: I tried something like that beforehand....
I cannot see how to make the match case structure work with an iterable directly, we need a set-like operation somewhere, at the place of all/any.., but I don't see how the match structure is compatible with sets, directly, I will go with len(set([])), this however defies the point of using match/case...

avaldebe (approved):
perfect

charlienegri:
in Thorbjørn's test I see the station_type (empty as it should) both in the colocated data and the map files  
I will merge this now

Issue #1553, opened 2025-03-21
Add tests for units conversion of standard-deviations


heikoklein:
**Is your feature request related to a problem? Please describe.**
Whenever units are converted, not only the values, but also the standard-deviations / errors need to be converted as well. Pyaerocom should test the conversion of standard-deviations whereever units are converted

**Additional context**
https://github.com/metno/pyaerocom/blob/fdbec6a0c579e2f2e0b4b8e788c9209c557ea18a/pyaerocom/ungriddeddata.py#L1531 clearly converts units and values, but not errors. I'm not sure if this method is ever used, though.

PR #1560, opened 2025-03-26
More units refactoring

thorbjoernl:
## Change Summary

- Implement a `Units` wrapper on `cf_units.Unit` which does adds some functionality previously done in `unit_helpers.py`, namely:
  - Takes two additional optional arguments during initialization, the `ts_type` and the `aerocom_varname`.
  - Does the handling of non-standard units eg `ug S m-3` to `ug m-3`.
    - This is done by including a scalar factor in the cfunit: in this case `ug S m-3` for `concso4` may be represented as `2.995870887260253 ug m-3` internally which makes cf_units unit conversion work largely unaltered when converting to `ug m^-3`.
  - Does the addition of implicit temporal frequencies based on `TsType` for rate variables.
  - `.convert` handles `pd.Series` and improves typehinting.
  - `.convert` takes an optional callback function that is called during conversion. My intention is to use this later to do logging using this (eg. print log message only if `factor != 1`).
- Bunch of typehints (as always :)).
- Use `TsType` instead of `str` in more places.
- Removes references to `UALIASES` outside `pyaerocom.units`.

This one goes deeper into the unit conversion so has more possibility to break things, so probably best with multiple pairs of eyes on this one. End goal is that basically all unit conversion outside of `pyaerocom.units` is done using a high level api (for instance `convert_unit` and similar functions).

## Checklist

* [x] Start with a draft-PR
* [x] Tests pass locally
* [x] Tests pass on CI
* [x] Make PR ready to review


pyaerocom/io/ghost/reader.py
```diff
@@ -362,9 +363,9 @@ def read_file(self, filename, var_to_read=None, invalidate_flags=None, var_to_wr
             for meta_key, to_unit in self.CONVERT_UNITS_META.items():
                 from_unit = ds[meta_key].attrs["units"]
 
-                if from_unit != to_unit:
-                    cfac = cf_units.Unit(from_unit).convert(1, to_unit)
-                    meta_glob[meta_key] *= cfac
+                meta_glob[meta_key] = convert_unit(
```
lewisblake, line 366: Why get rid of the if statement?

pyaerocom/units/unit.py
```diff
@@ -0,0 +1,375 @@
...
+    Third, cf_units.Unit does not natively support conversion of eg. pd.Series.
+    This wrapper allows conversion of any data structure that supports __mul__.
+    """
+
+    #: Custom unit conversion factors for certain variables
+    #: columns: variable -> from unit -> to_unit -> conversion
+    #: factor
+    _UCONV_MUL_FACS = pd.DataFrame(
+        [
+            # ["dryso4", "mg/m2/d", "mgS m-2 d-1", M_S / M_SO4],
```
lewisblake: Is this dead code from long ago?

lewisblake (approved):
Seems pretty reasonable to be. I think it could also be good to test this with an experiment here we have to do unit conversions to see how this effects things. It would give more confidence that it works in practice before merging and reduce potential frustration if we miss some behavior not covered by tests.

pyaerocom/units/unit.py
```diff
@@ -0,0 +1,375 @@
...
+    from_aerocom_var: str | None
+    from_ts_type: TsType | None
+    from_cf_unit: cf_units.Unit
+    to_cf_unit: cf_units.Unit
+
+
+UnitConversionCallbackHandler = Callable[[UnitConversionCallbackInfo], None]
+
+
+class PyaerocomUnit:
```
heikoklein: Since this is the only Unit which should be used within pyaerocom, please call it Unit, not pyaerocom.unit.PyaerocomUnit.

pyaerocom/io/ghost/reader.py
thorbjoernl: Converting between the same units should be equivalent to multiplying by 1, so shouldn't change anything. Saves a line :)

pyaerocom/io/aux_components_fun.py
```diff
@@ -12,6 +12,9 @@
 )
 from pyaerocom.units.molecular_mass import get_molmass
 
+from pyaerocom.units import PyaerocomUnit
+
```
heikoklein, line 15: Why is cf_units not removed here?

pyaerocom/io/iris_io.py
```diff
@@ -15,6 +15,8 @@
 import iris.coords
 import iris.util
 
+from pyaerocom.units import PyaerocomUnit
+
```
heikoklein: PyaerocomUnit added but cf_units not removed?

pyaerocom/stationdata.py
```diff
@@ -231,10 +235,10 @@ def check_unit(self, var_name: str, unit: str | None = None):
         if unit is None:
             unit = const.VARS[var_name].units
         u = self.get_unit(var_name)
-        if not get_unit_conversion_fac(u, unit, var_name) == 1:
+        if not convert_unit(1, u, unit, var_name) == 1:
             raise DataUnitError(f"Invalid unit {u} (expected {unit})")
 
-    def convert_unit(self, var_name: str, to_unit: str):
+    def convert_unit(self, var_name: str, to_unit: str) -> None:
```
heikoklein, line 241: This function only converts data, but not data-stdev. See https://github.com/metno/pyaerocom/issues/1553

pyaerocom/units/unit.py
```diff
@@ -0,0 +1,375 @@
...
+    """Pyaerocom specific encapsulation of cf_units.Unit that extends it
+    with additional needed behaviour.
```
heikoklein: We should be able to switch to another units package (or write our own) at one point in the future. No need to mention cf_units in the docs.

pyaerocom/units/unit.py
```diff
@@ -0,0 +1,375 @@
...
+                "from_aerocom_var": self._aerocom_var,
+                "from_ts_type": self._ts_type,
+                "from_cf_unit": str(self._cfunit),
+                "to_cf_unit": str(to_unit),
+            }
+            callback(info)
+        return value
+
+    @property
+    def cftime_unit(self) -> str:
```
heikoklein: why is cftime exposed here?

heikoklein (commented):
Very nice refactoring. The unit.py class will be one of the most used classes of pyaerocom. A higher test-coverage is desirable, and would make the PR better readable.

All public functions and properties in unit.py need documentation.

unit.py should be named units.py

pyaerocom/units/unit.py
thorbjoernl: This class is written to mirror the interface of `cf_units.Unit` except for the additional functionality. Is there a reason why this shouldn't be exposed?

pyaerocom/units/unit.py
thorbjoernl: This class is pretty tightly coupled with cf_units. It mirrors it's interface and function signatures are almost identical to cf_units. Many variables are set with values defined by cf_units for instance (eg. calendar strings).

pyaerocom/units/unit.py
heikoklein: Maybe, at one point, we want to use another base-package for units. We don't need to expose everything just because cf_units does it. Cutting down on a interface might be useful, and only exposing function when really needed.

heikoklein (approved):

PR #1566, opened 2025-03-28
Pin pydantic <3.0 and ignore pydantic deprecation warning

magnusuMET:
## Checklist

* [ ] Start with a draft-PR
* [x] Tests pass locally
* [ ] Tests pass on CI
* [ ] Make PR ready to review

### DIFF
--- a/pyaerocom_env.yml
+++ b/pyaerocom_env.yml
@@ -18,7 +18,7 @@ dependencies:
   - tqdm
   - openpyxl
   - typer >=0.7.0
-  - pydantic >= 2.7.1
+  - pydantic ~= 2.7
   - pyproj >= 3.0.0
   - pooch >=1.7.0
   - psutil >= 5.0.0
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -237,7 +237,7 @@ deps =
     importlib-resources ==5.10; python_version < "3.11"
     typing-extensions ==4.6.1; python_version < "3.11"
     cf-units ==3.3.0; python_version < "3.11"
-    pydantic ==2.7.1; python_version < "3.11"
+    pydantic ==2.7.0; python_version < "3.11"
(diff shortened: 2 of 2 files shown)


magnusuMET:
Alternative to #1565

pyaerocom_env.yml
```diff
@@ -18,7 +18,7 @@ dependencies:
   - tqdm
   - openpyxl
   - typer >=0.7.0
-  - pydantic >= 2.7.1
+  - pydantic ~= 2.7
```
thorbjoernl, line 21: Also change minimum dependency version tests below to 2.0.0 (though I suspect this will break). I'd use `>= 2.7.0, < 3.0.0`

thorbjoernl (changes requested):

pyaerocom_env.yml
thorbjoernl: Actually, this matched only 2.7.x right? That's not what we want.

magnusuMET: `~=2.7` is equivalent to `>=2.7.0, < 3.0.0` but I can use the explicit dep limits if you wish

thorbjoernl: No no need, I thought it was within a minor version.

But do check that the lowest version runs by adding in minimal depenencies tests

magnusuMET: As in add a `pip install -e ./ pydantic==2.7.0` test in CI?

thorbjoernl: Here: https://github.com/metno/pyaerocom/blob/e392ae4138873fe7b2e3267fdb8c69a02fb828fa/pyproject.toml#L238

magnusuMET: Did not see that one, thanks, added and we will see if CI complains

thorbjoernl (approved):
LGTM, assuming tests pass

PR #1556, opened 2025-03-24
try parallellized tests

thorbjoernl:
## Change Summary

- Removes plotting (see comment below)
- Uses pytest-xdist to parallellize tests in ci (does work, but not too much of a speed up in ci).

## Related issue number

closes #1554

## Checklist

* [x] Start with a draft-PR
* [x] The PR title is a good summary of the changes
* [x] PR is set to AeroTools and a tentative milestone
* [x] Documentation reflects the changes where applicable
* [x] Tests for the changes exist where applicable
* [x] Tests pass locally
* [x] Tests pass on CI
* [x] At least 1 reviewer is selected
* [x] Make PR ready to review

Issue #1411, opened 2024-11-19
nice and unique units in pyaerocom output for aeroval


heikoklein:
**Is your feature request related to a problem? Please describe.**
The maps endpoint/json files and the ts endpoint/json-files contain units of observations and models. Even if they are run from the same config the units of these endpoints can be slightly different, e.g. https://aeroval.met.no/pages/maps/?project=emep&experiment=2024-reporting and https://aeroval.met.no/pages/evaluation/?project=emep&experiment=2024-reporting show the same but differently notated units, e.g. `ug N m-3`
 vs `µg N / m3`. 

This is both confusing for end-users and for also for data-checks like: https://github.com/metno/AeroToolsIssues/issues/62

The units should be written using the same notation from the configuration. The units should be both CF-compliant and nice. udunits as basis of the CF-Convention allows nowadays superscripts and µ, e.g. `µg N / m³` is nice and accepted

**Describe the solution you would like to see**

Units of a component should come from one configuration and support µ and ³.

lewisblake:
One thought about this: this would potentially introduce another alias. In my opinion, it is preferable to do the work related to organizing the unit-business into a single module, and then implement changed like this in there.

lewisblake:
Now that #1552 is merged we can approach this more systematically.

PR #1575, opened 2025-03-31
[Units] - Part 3

### LINKED ISSUE #1512: Introduce a units module
pyaerocom should have a `units` module where all functions related to unit declarations and conversions are stored. Ideally a target unit would be provided for each variable, and then if not available the default behavior of converting to the units of the observations could continue.

This is important for a number of reasons. Most recently it came up because super-observations where networks report the same variable in different units can not currently be created correctly.

 - Would you be able to work on this solution yourself?
 - How can the Pyaerocom development team assist you?

**Additional context**
Related issues inlcude:
- #1499
- #1021 
- #1411

### LINKED ISSUE #1553: Add tests for units conversion of standard-deviations

thorbjoernl:
## Change Summary

- Ensure units module shows up on docs: https://pyaerocom--1575.org.readthedocs.build/en/1575/api.html#units
- Remove custom logic in `GriddedData` and just use units module for conversion.
- Make use of callback pattern for logging in some places with unit conversion.
- Users can specify a units dict in their config, which will override units specified in `variables.ini`.

### Unrelated / minor fixes.
- Removes plotting related documentation (code was removed as part of #1556, but docs lingered)
- fix: Mapmodels engine producing no output if both overlays and contours were requested.
- fix: Use of wrong Exception types in some places.

## Related issue number

- closes #1512
- closes #1553

## Checklist

* [x] Start with a draft-PR
* [x] Tests pass locally
* [x] Tests pass on CI
* [x] Make PR ready to review


thorbjoernl:
I think this is ready for review now, though I'm waiting for a run to go through to see if #1411 is fixed. Therefore leaving it as draft for now.

Last EMEP test run on this branch: https://aeroval-test.met.no/thlun/pages/evaluation/?project=emep2024-test&experiment=emep2024-test-20250409T143105.703841

pyaerocom/aeroval/modelmaps_engine.py
```diff
@@ -108,19 +108,14 @@ def _run_model(self, model_name: str, var_list):
...
                 if isinstance(self.cfg.modelmaps_opts.plot_types, dict):
-                    make_contour = CONTOUR in self.cfg.modelmaps_opts.plot_types.get(
-                        model_name, False
-                    )
-                    make_overlay = OVERLAY in self.cfg.modelmaps_opts.plot_types.get(
-                        model_name, False
-                    )
-                if self.cfg.modelmaps_opts.plot_types == {CONTOUR} or make_contour:
-                    self._process_contour_map_var(model_name, var, self.reanalyse_existing)
+                    plot_types = self.cfg.modelmaps_opts.plot_types.get(model_name, [])
```
lewisblake: This is much cleaner logic, thanks

pyaerocom/aeroval/experiment_output.py
```diff
@@ -32,8 +32,9 @@
 from pyaerocom.exceptions import EntryNotAvailable, VariableDefinitionError
 from pyaerocom.stats.mda8.const import MDA8_OUTPUT_VARS
 from pyaerocom.stats.stats import _init_stats_dummy
+from pyaerocom.units.helpers import get_standard_unit
```
lewisblake, line 35: Good naming

pyaerocom/_lowlevel_helpers.py
```diff
@@ -331,6 +331,25 @@ def pretty_str(self):
     def __str__(self):
         return str(self.to_dict())
 
+    def __eq__(self, other) -> bool:
```
lewisblake: One of the first code smells I encounter is defining a dunder for equality with `BrowseDict`. Maybe it'll become clearer to me later on, as I review, but in general I think we should stay away from expanding upon this class / using it more. There usually is a better data structure, and if you ever have to debug something with this class, you're in for a washing machine ride of recursion.

pyaerocom/aeroval/modelmaps_engine.py
```diff
@@ -620,4 +612,13 @@ def _process_only_json(self, model_name, var):  # pragma: no cover
         data = data.drop_vars("data_source")
         data = data.transpose("time", "latitude", "longitude")
         data = data.sortby(["latitude", "longitude"])
-        return data
+
+        out = GriddedData(data.to_iris())
+
+        # NOTE: For some reason, creating a GriddedData object as above results in a (seemingly valid)
```
lewisblake: This also needs to work on `ColocatedData` objects, which you may recall is under the hood an xarray DataArray which has a source dimension, hence the reason this is written to also work with xarray. Unfortunately this is not covered by testing. So your solution may be a good one, but it needs verification on features which are a bit cutting edge right now.

pyaerocom/aeroval/_processing_base.py
```diff
@@ -28,6 +30,7 @@ class HasConfig:
 
     def __init__(self, cfg: EvalSetup):
         self.cfg = cfg
+        set_unit_overrides(self.cfg.units_cfg.units)
```
lewisblake: Does this function do nothing if the units are not overridden?

pyaerocom/ungriddeddata_structured.py
```diff
@@ -486,7 +450,8 @@ def remove_outliers(
         else:
             new = self.copy()
 
-        new.check_convert_var_units(var_name, to_unit=unit_ref)
+        new.check_unit(var_name)
+        # new.check_convert_var_units(var_name, to_unit=unit_ref)
```
lewisblake: dead code

pyaerocom/units/logging/callbacks.py
```diff
@@ -0,0 +1,36 @@
+import logging
+from pyaerocom.units.units import UnitConversionCallbackInfo
+
+logger = logging.getLogger(__name__)
+
+
+class LoggingCallback:
```
lewisblake, line 7: Nice use of the callback pattern

pyaerocom/_lowlevel_helpers.py
thorbjoernl: I agree. I did this as a quick fix for a test failure which suddenly failed due the changes, but probably not the right fix. It is not used outside of tests.

pyaerocom/aeroval/_processing_base.py
thorbjoernl: Overrides should default to empty dict, in which case fallback is to use units defined in `variables.ini`

thorbjoernl:
Side note: I ran through the emep config before easter, and it looks like the unit overrides don't make their way through. Not sure why though:

- https://aeroval-test.met.no/thlun/pages/evaluation/?project=emep2024-test&experiment=emep2024-test-20250415T084547.085725&parameter=concpm10&tab=experiment

Here pm10 should have nicely formatted units (`μg / m³`) but they don's show up like that on the web. Need to investigate.

pyaerocom/aeroval/modelmaps_engine.py
thorbjoernl: I've reversed these changes, so hope the previous functionality is restored.

thorbjoernl:
Note: The CI test failure is due to [this change](https://github.com/metno/pyaerocom/pull/1613) from pyaro not being in a release yet. They work locally.

Before merging, a pyaro release should be made and minimum pyaro version adjusted.

mean_dryHNO3_2008_hm.tif
lewisblake, line 1: Can these tif files be organized in a better place? It looks to me like they're just sitting in the base of the directory tree

pyaerocom/griddeddata.py
```diff
@@ -45,10 +47,13 @@
 from pyaerocom.time_resampler import TimeResampler
 from pyaerocom.units.datetime import TsType
 from pyaerocom.units import Unit
-from pyaerocom.units.units_helpers import get_unit_conversion_fac
+from pyaerocom.units.helpers import get_standard_unit
```
lewisblake, line 50: Long term I think we should think of a more descriptive name than helpers

lewisblake (approved):

mean_dryHNO3_2008_hm.tif
lewisblake: Move these into a more appropriate place if you can, then merge when see fit.

thorbjoernl: Those aren't supposed to be part of the PR, so will remove.

pyaerocom/griddeddata.py
thorbjoernl: Definitely

Issue #1384, opened 2024-10-21
Change all paths in pyaerocom from strings to Path-like


lewisblake:
**Is your feature request related to a problem? Please describe.**
strings are used all over. A better choice would be to use e.g., `Path` from `pathlib`. Then pyaerocom can be used in a VRE, we can read data directly from Thredds. The code would then be agnostic to if it's reading a real path or a URL. 

**Describe the solution you would like to see**
Above
 - Would you be able to work on this solution yourself?
 - How can the Pyaerocom development team assist you?

**Additional context**

PR #1594, opened 2025-04-15
Implement EPROFILE reader

lewisblake:
## Change Summary

- Introduce a reader for the EPROFILE data. Data is 3D, so need to implement an internal reader in pyaerocom analogous to the EARLINET reader.
- Introduce extinction and backscatter variables at 1064nm
- Made and uploaded new testdata-minimal: `testdata-minimal.tar.gz.20250425`. Updated pyaerocom to make this is version to use.
- Introduce `EprofileFileError` (and `EarlinetFileError`)

### Minor / unrelated
- Fix decorators in `readungriddedbase.py` to their supported analogues
- Fix the EARLINET reader in many of the analogous ways to EPROFILE based on feedback
- Remove plotting functionality for `VerticalProfle` (was not removed during previous PRs) 
- Fix logging in `read_earlinet.py`. Made an issue to clean this up in parent class #1600

## Related issue number

Close https://github.com/metno/AeroToolsIssues/issues/97

## Checklist

* [x] Start with a draft-PR
* [x] Tests pass locally
* [x] Tests pass on CI
* [x] Make PR ready to review


lewisblake:
Ready for a round of feedback I'd say. I want to run some experiments with this next week to see how it performs / looks on Aeroval.

pyaerocom/io/read_eprofile.py
```diff
@@ -0,0 +1,547 @@
...
+            vars_to_retrieve = [vars_to_retrieve]
+        _vars = []
+
+        for var in vars_to_retrieve:
+            if var in self.VAR_NAMES_FILE:  # make sure to only read what is supported by this file
+                _vars.append(var)
+            elif var in self.AUX_REQUIRES:
+                _vars.append(var)
+            else:
+                raise ValueError(f"{var} is not supported")
```
thorbjoernl, line 159: Could be a more specific exception. I see we have `VarNotAvailableError` already defined, but not sure that is the appropriate one here.

pyaerocom/io/read_eprofile.py
```diff
@@ -0,0 +1,547 @@
...
+
+        #: files that are supposed to be excluded from reading
+        self.exclude_files = []
+
+        #: files that were actually excluded from reading
+        self.excluded_files = []
+
+        self.is_vertical_profile = True
+
+    def read_file(self, filename, vars_to_retrieve=None, remove_outliers=True) -> StationData:
```
thorbjoernl, line 128: It would be nice to use the `@override` decorator here, and check them with mypy. Ensures that function signatures are consistent with the base class.

```python
import sys

if sys.version_info >= (3, 12):
    from typing import override
else:
    from typing_extensions import override
```

Then run mypy with [--explicit-override](https://mypy.readthedocs.io/en/stable/error_code_list2.html#check-that-override-is-used-when-overriding-a-base-class-method-explicit-override) option.
(Applies to all overridden functions)

pyaerocom/io/read_eprofile.py
```diff
@@ -0,0 +1,547 @@
...
+                    elif "Number of" in line:
+                        num = int(line.split(":")[1].strip())
+                        indata = True
+
+            if not count == num:
+                raise Exception
+        self.exclude_files = list(dict.fromkeys(exclude))
+        return self.exclude_files
+
+    def get_file_list(self):
```
thorbjoernl: Should be private?

pyaerocom/io/read_eprofile.py
```diff
@@ -0,0 +1,547 @@
...
+
+        Returns
+        -------
+        list
+            list containing file paths
+        """
+        exclude_files = set([Path(file) for file in self._get_exclude_filelist()])
+        logger.info("Fetching EPROFILE data files...")
+
+        all_files = set(Path(self.data_dir).rglob(self._FILEMASK))
```
thorbjoernl: I think rglob includes paths for folders, and it is theoretically possible for someone to name a folder ending with .nc so I would include a `.isfile()` check

pyaerocom/io/read_eprofile.py
```diff
@@ -0,0 +1,547 @@
...
+
+        # shorten data_obj._data to the right number of points
+        data_obj._data = data_obj._data[:idx]
+
+        return data_obj
+
+    def _get_exclude_filelist(self):  # pragma: no cover
+        """Get list of filenames that are supposed to be ignored"""
+        exclude = []
+        import glob
```
thorbjoernl: Please put imports at top of file (ref. [PEP-8](https://peps.python.org/pep-0008/#imports))

pyaerocom/io/read_eprofile.py
thorbjoernl:
```suggestion
    def get_file_list(self) -> list[pathlib.Path]:
```

pyaerocom/io/read_eprofile.py
```diff
@@ -0,0 +1,547 @@
...
+                self.logger.exception(
+                    f"Failed to read file {os.path.basename(_file)} (ERR: {repr(e)})"
+                )
+
+        # shorten data_obj._data to the right number of points
+        data_obj._data = data_obj._data[:idx]
+
+        return data_obj
+
+    def _get_exclude_filelist(self):  # pragma: no cover
```
thorbjoernl, line 501: In favour of consistency and maybe moving towards closing #1384 I think we should either use glob.glob with string paths, or use `Path.glob()`, not mix them.

pyaerocom/io/read_eprofile.py
```diff
@@ -0,0 +1,547 @@
...
+                for u in unames:
+                    if u in arr.attrs:
+                        unit = arr.attrs[u]
+                if unit is None:
+                    raise DataUnitError(f"Unit of {var} could not be accessed in file {filename}")
+                unit_fac = None
+                if len(unit) > 0:
+                    try:
+                        to_unit = self._var_info[var].units
+                        unit_fac = get_unit_conversion_fac(unit, to_unit)
```
thorbjoernl: Consider using `convert_unit()` instead.

pyaerocom/io/read_eprofile.py
lewisblake: Cleaned up quite a few errors raises in the EPROFILE reader and its cousin EARLINET

pyaerocom/io/read_eprofile.py
lewisblake: I think it should be public

thorbjoernl: Misunderstood the inheritance here. Agree.

pyaerocom/io/read_eprofile.py
lewisblake: Agreed, will change

thorbjoernl (approved):
LGTM

Issue #1600, opened 2025-04-28
Remove ReadUngriddedBase's own instance of logger class


lewisblake:
**Is your feature request related to a problem? Please describe.**
`ReadUngriddedBase` has it's own logger 

**Describe the solution you would like to see**
 - replace all occurances of `self.logger` with `logger` where `logger = logging.getLogger(__name__)`. That is the name of the logger is tied to the module.

**Additional context**
NA

PR #1601, opened 2025-04-28
Change all occurrences of self.logger to logger

lewisblake:
## Change Summary

Title. Classes should not need to maintain their own loggers. I took a poll and no developer is using / favors this functionality. Module level logging was preferred.

I additionally removed the "logger" in the colocator.

## Related issue number
#1600

## Checklist

* [x] Start with a draft-PR
* [x] Tests pass locally
* [x] Tests pass on CI
* [x] Make PR ready to review

### DIFF
--- a/pyaerocom/_lowlevel_helpers.py
+++ b/pyaerocom/_lowlevel_helpers.py
@@ -118,7 +118,7 @@ class StrType(Validator):
 class Loc(abc.ABC):
     """Abstract descriptor representing a path location
 
-    Descriptor???
+    Descriptor: TODO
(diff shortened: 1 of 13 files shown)


lewisblake:
Probably the colocator's logger should be scratched and the python logging used instead (since changing to another logging package is outside the scope of this PR). Here is an example of what the current logging files contain:

```
------------------ NEW ----------------
Timestamp: 06-02-2025 11:00

Analysis configuration
```

Safe to say we can probably get rid of it.

heikoklein (approved):
Thanks, and agreed, this should make debugging from log-files easier.

Issue #1581, opened 2025-04-03
Pyaerocom should be able to handle different stations with the same name


lewisblake:
**Describe the bug**
Please provide a clear and concise description of what the bug is.
 - Pyaerocom version: 0.29.0
 - Computing platform: All
 - Configuration file (if applicable):
 - Error message (if applicable):

There are two cities named Amherst (one in MA, the other in NY) in the AirNow data. Pyaerocom should not assume that station names are unique. Rather, each instance of `StationData` should have a primary key, which uniquely identifies the instance and the processing should iterate over those instead.

~/aerocomB/aerocom1/AEROCOM_OBSDATA/AirNow/allStations_20191224.csv                                                                                                                                                                    "360290002","AMHERST","42.993","-78.772","181",,"AUDUBON GOLF COURSE, 450 MAPLE ROAD","POINT (-78.77153 42.99328)","America/New_York","Residential","2015-04-29 21:33:43","urban",

"250150103","AMHERST","42.401","-72.523","81",,"N PLEASANT ST","POINT (-72.523109 42.400579)","America/New_York","Agricultural","2015-04-29 21:33:43","urban",                          

**To Reproduce**
1. Run an experiment with the new AirNow files

**Expected behavior**

Should be able to handle networks with stations with the same name

**Screenshots**

**Additional context**
@andagit may be able to provide more context / be a good contact person regarding some of the issues she faced.

magnusuMET:
I also want this for the parquet EEA files (the above is for airnow), where station names are very long. It would be better to treat these as primary keys and have some notion of `short_name` which could be used for e.g. visualisation

dulte:
Implementing some kind of hash internally, instead of station_name? Then at some point do the same for output

PR #1634, opened 2025-05-19
Add display_name for stations

magnusuMET:
## Change Summary

There are two primary motivations for separating `station_name` and `display_name`:

* Some readers have `station_name` which are used as index key, but are not helpful for the users of aeroval
* pyaerocom implicitly requires station names to be unique which may cause issues when merging different observation networks

A result can be seen at https://aeroval-test.met.no/magnusu/pages/evaluation/?project=eeareadertest&experiment=display-name (cf. [older experiment](https://aeroval-test.met.no/magnusu/pages/evaluation/?project=eeareadertest&experiment=2024-reporting-catalog-eea-reader))

An alternative would be to rename `station_name` to `station_id`, but this requires extensive changes over most of the code of pyaerocom and aeroval

## Related issue number

Potentially addressing https://github.com/metno/pyaerocom/issues/1581

## Checklist

* [x] Start with a draft-PR
* [x] Tests pass locally
* [x] Tests pass on CI
* [x] Make PR ready to review

### DIFF
--- a/pyaerocom/aeroval/coldatatojson_helpers.py
+++ b/pyaerocom/aeroval/coldatatojson_helpers.py
@@ -491,7 +491,7 @@ class CoordSiteArrays:
     alts: list[float]
     countries: list[str]
     jsdates: dict[str, list[str]]
-    display_names: list[str] | None
+    display_names: list[str | None] | None
 
 
 def _init_site_coord_arrays(data) -> CoordSiteArrays:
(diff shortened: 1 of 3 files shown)


heikoklein (changes requested):
This looks very nice, but it is very confusing that the upper left map works with different names than the other graphs:

image

I guess both names need to be given in the json-files in the `map`-folder.

And I guess, the `station_display_name` has already been implemented in aeroval? At least for the `ts`-folder?

heikoklein (changes requested):
I don't understand in this PR how the new metadata is transferred from the reader to the UngriddedData and to the StationData. 

A new metadata-field is introduced without documenting it anywhere properly.

magnusuMET:
The maps endpoint required whitelisting of the metadatakey in aerovaldb https://github.com/metno/aerovaldb/pull/136

Metadata is transferred through the colocated data object as done for `station_type`. It is not pretty

heikoklein (approved):
This is very complicated to understand, but ok.

PR #1638, opened 2025-05-21
EPROFILE: New format

lewisblake:
## Change Summary

Updated the EPROFILE reader to the new format which the data is being converted to. Notably the altitude is now given above ground level, so needed to add the station altitude to get altitude above sea level. The coordinates are now time dependent, although this shouldn't be an issue for us as we had to store them for each time step in `UngriddedData` anyway.

I updated the testdata-minimal with these new files, and updated the tests accordingly.

~I also removed the `__init__.py` file from the tests dir, as that was causing problems with pytest.~

I removed pyaro-readers from the testing env on CI and skipped tests which relied on the package. 
## Related issue number

NA

## Checklist

* [x] Start with a draft-PR
* [x] Tests pass locally
* [x] Tests pass on CI
* [x] Make PR ready to review

### DIFF
--- a/pyaerocom/io/read_eprofile.py
+++ b/pyaerocom/io/read_eprofile.py
@@ -84,9 +84,9 @@ class ReadEprofile(ReadUngriddedBase):
     #: metadata keys that are needed for reading (must be values in
     #: :attr:`META_NAMES_FILE`)
     META_NEEDED = [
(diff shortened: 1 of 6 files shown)


jgriesfeller (approved):
all good

PR #1656, opened 2025-06-18
SESAM + EPROFILE Fixes

### LINKED ISSUE #1639: menu.json not created when `only_model_maps=True`
**Describe the bug**
 - Pyaerocom version: >= 0.31.dev0
 - Computing platform: all
 - Configuration file (if applicable): Can share if needed
 - Error message (if applicable): None (fails silently)

**To Reproduce**
1. Run an experiment with  `only_model_maps=True`

**Expected behavior**
menu.json should be created. Otherwise Aeroval can not show the results.

**Screenshots**

**Additional context**
Originally addressed in https://github.com/metno/pyaerocom/pull/1434, but not properly tested, so probably some other changes to aerovaldb or something has broken how this file was made. All other files appear to be where they should be, which is promising.

lewisblake:
## Change Summary

This PR unfortunately contains the changes of two unrelated things which I needed to work on simultaneously. One aspect is related to SESAM, and some fixes needed for the Aeroval evaluation we're making for the EMEP report. The other is related to EPROFILE data, which is part of CAMS2_82.

- Fix up computation of menu.json 
- Fix bug related to model_list in `ExperimentProcessor` - was being over written 
- Implement the EPROFILE reader in terms of `UngriddedDataStructured` to use ~1/3 of the memory. `UngriddedDataStructured` needed to be generalized to handle vertical profiles.
- Other EPROFILE reader fixes  
- Generalize merging of `StationData` with 3d data to handle data not sampled at the same time
- Implement ColocationSetup validators for `profile_layer_limits`, `colocation_layer_limits`
- Some clean up in UngriddedDataStructured. For example, remove unnecessary conversion of each column to their nan-type. 
- Clean up `DynamicRecArray.data`.
- Emojis for key logging statements. Aids in readability.

## Related issue number
Close #1639

## Checklist

* [x] Start with a draft-PR
* [x] Tests pass locally
* [x] Tests pass on CI
* [x] Make PR ready to review


lewisblake:
Not sure what's causing these to fail right now - they all pass locally

lewisblake:
@heikoklein Could you please comment on whether the point in the change description,

"Some clean up in UngriddedDataStructured. For example, remove unnecessary conversion of each column to their nan-type.",

specifically the part about removing the conversion step, was in fact correct to do? I'm wondering now if this was needed for the memory optimization, but it did not appear to benefit the code and in fact caused me some issues so I removed it.

lewisblake:
Now I'm just back to the EPROFILE read test failing inexplicably. This is similar to an issue @jgriesfeller had, where probably GitHub CI is being a little too "intelligent", and using a cached version of the data. Not sure what the solution was to get around this. But I will mark the PR as ready for review because otherwise there is a lot here which would benefit from feedback.

dulte:
@lewisblake , in colocation_3d, where you check if the station_data is xarray or pandas, is there a reason that you don't just cast the pandas series into an xarray DataArray, instead of doing the complicated series aggregation?

pyaerocom/dynamic_rec_array.py
```diff
@@ -75,10 +75,8 @@ def data(self) -> np.array:
 
         :return: np.aarry of type dtype
         """
-        if self._capacity != self._length:
-            self._array = self._array[:][: self._length]
-            self._capacity = len(self._array)
-        return self._array
+        # NOTE: Commenting out the shrinking of the array might be needed in debug mode
+        return self._array[:][: self._length]
```
dulte: You note that "Commenting out the shrinking of the array might be needed in debug mode", but it seems that you have completely removed this feature. Is this on purpose?

dulte (commented):
As far as I can see, in `read_eprofile`, `remove_outliers` is an parameter for both `read_file` and `read`, but is never past between them, meaning `remove_outliers` does nothing

lewisblake:
> @lewisblake , in colocation_3d, where you check if the station_data is xarray or pandas, is there a reason that you don't just cast the pandas series into an xarray DataArray, instead of doing the complicated series aggregation?

The conversion to an xarray object with the proper format is similarly complicated, and `_colocated_site_data_helper` & `_colocate_site_data_helper_timecol` operate on pandas series anyway.

pyaerocom/dynamic_rec_array.py
lewisblake: Removed

lewisblake:
> As far as I can see, in `read_eprofile`, `remove_outliers` is an parameter for both `read_file` and `read`, but is never past between them, meaning `remove_outliers` does nothing

It's applied on line 311 in the function `read_file`, but you are correct that this never gets passed to `read_file` from `read` (through `_read_files_structured` and `_station_data_iterator`), so good catch! Should be passed correctly now.

lewisblake:
@jgriesfeller The lesson we learned from this PR is that paths.ini is dealt with differently in CI tests than regular pyaerocom. The `const.OBSLOCS_UNGRIDDED` registered during tests are those that are in `DataForTests`.  You may need to specify directly in your tests which element of `DataForTests` should be the obs_id in your reader in the tests.

dulte (approved):
With all the tests passing, this is ready for merging :champagne:

pyaerocom/ungriddeddata_structured.py
```diff
@@ -639,21 +675,22 @@ def append_station_data(
...
+                    uds = UngriddedDataStructured(num_points=len(values))
                 v_data = uds._dra._array  # access to raw numpy-array
                 v_data["meta_id"][:] = meta_idx
                 v_data["var_id"][:] = var_idx
                 v_data["data"] = values
                 v_data["start_time"] = times
                 # v_data["end_time"] not used
+                if altitude is not None:
+                    v_data["dataaltitude"] = altitude
                 # v_data["dataaltitude"] not used
```
heikoklein, line 694: @lewisblake looks like you started using `v_data["dataaltitude"]` for vertical profiles, so the comment should be removed or modified to `v_data["dataaltitude"] used for vertical profiles`

Issue #1590, opened 2025-04-10
Export of ungridded-data to stationdata should ensure that flags and errors are handled like values


heikoklein:
**Describe the bug**
stddev and flags are handled differently from values when ungridded data is exported/merged to StationData.

Code in https://github.com/metno/pyaerocom/blob/d4b9555e6524ce9a4af1e0727bd884a58f2bcf3b/pyaerocom/ungriddeddata_structured.py#L372 and https://github.com/metno/pyaerocom/blob/d4b9555e6524ce9a4af1e0727bd884a58f2bcf3b/pyaerocom/ungriddeddata.py#L791 sort values, but not errors and flags.

Code in https://github.com/metno/pyaerocom/blob/d4b9555e6524ce9a4af1e0727bd884a58f2bcf3b/pyaerocom/stationdata.py#L782 shows concat values, but not flags and errors

The general problem of StationData seems to be, that times and values are stored in one pandas.Series object, while flags and errors are in two  separate arrays.

**Expected behavior**

A test with data including flags and errors for different sources should identify the error, e.g. that if data is unsorted, stationdata-errors and flags should be sorted the same way as the data-values.

Another test should identify where merging of stations without merging of flags occurs.

**Additional context**
UngriddedDataStructured has some code to warn the inconsistent length of flags and errors, needed for EEA-evaluation: https://github.com/metno/pyaerocom/blob/683d0cf9ab8c746daf0eda6d2d56ba26957cce90/pyaerocom/ungriddeddata_structured.py#L261

thorbjoernl:
re: https://github.com/metno/pyaerocom/blob/d4b9555e6524ce9a4af1e0727bd884a58f2bcf3b/pyaerocom/stationdata.py#L782

This functionality relies on [temporal resampling of the time series](https://github.com/metno/pyaerocom/blob/d4b9555e6524ce9a4af1e0727bd884a58f2bcf3b/pyaerocom/stationdata.py#L752), which currently does not occur for errors (afaict the error values of the original data are [simply kept without resampling](https://github.com/metno/pyaerocom/blob/62a55d8287b15ecd24eb01213ecda3d26e6669bb/pyaerocom/time_resampler.py#L234)). Can the errors be resampled in the same way as the data, or is that statistically invalid?

heikoklein:
[...]

PR #1662, opened 2025-06-24
Make sorting of values and errors consistent (+ resample errors)

thorbjoernl:
## Change Summary

Make sorting of values and errors consistent (+ resample errors). This does not yet address flags, as I am not sure how exactly they should be handled during resampling? Maybe resample just based on valid data, then all data is valid after a resampling operation?

## Related issue number

#1590 

## Checklist

* [x] Start with a draft-PR
* [x] Tests pass locally
* [x] Tests pass on CI
* [x] Make PR ready to review

### DIFF
--- a/pyaerocom/helpers.py
+++ b/pyaerocom/helpers.py
@@ -806,7 +806,7 @@ def calc_climatology(s, start, stop, min_count=None, set_year=None, resample_how
 
 
 def resample_errors(ds: pd.Series, *args, **kwargs):
-    return math.sqrt((ds**2).sum(skipna=True)) / len(ds)
+    return math.sqrt((ds**2).sum(skipna=True)) / len(ds[ds.notna()])
 
 
 def resample_timeseries(ts, freq, how=None, min_num_obs=None):
--- a/pyaerocom/helpers.py
+++ b/pyaerocom/helpers.py
@@ -36,6 +36,7 @@ from pyaerocom.units.datetime import TsType, is_year, to_pandas_timestamp
 from pyaerocom.variable_helpers import get_variable
 
 from pyaerocom.units import Unit
+import math
 
 logger = logging.getLogger(__name__)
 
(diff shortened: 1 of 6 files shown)


pyaerocom/helpers.py
```diff
@@ -836,6 +841,12 @@ def resample_timeseries(ts, freq, how=None, min_num_obs=None):
         p = int(how.split("percentile")[0])
         how = lambda x: np.nanpercentile(x, p)  # noqa: E731
 
+    if how == "error":
+        if not isinstance(ts, pd.Series):
+            raise NotImplementedError("Not yet implemented.")
+
+        how = resample_errors
```
heikoklein, line 852: This looks wrong. `resample_errors` for a `mean` needs a `1/n` factor, not only the sqrt of the square-sums. With values like $ts_i \pm \sigma_i$:
mean = $\frac{\sum_{i=1}^n ts_i}{n}$
mean-error = $\frac{\sqrt{\sum_{i=1}^n \sigma_i^2}}{n}$

Error for percentiles: should be derived from both the distribution of values and their errors. @lewisblake : some statistics please...

But error-resampling depends on the data-resampling method and should be somehow connected.

heikoklein (commented):
Thanks, most problems with errors are solved. I'm still a bit unsure if the error-propagation during the resampling is correct, please double-check...

pyaerocom/helpers.py
```diff
@@ -806,7 +806,7 @@ def calc_climatology(s, start, stop, min_count=None, set_year=None, resample_how
 
 
 def resample_errors(ds: pd.Series, *args, **kwargs):
-    return math.sqrt((ds**2).sum(skipna=True))
+    return math.sqrt((ds**2).sum(skipna=True)) / len(ds)
```
heikoklein: close, but `len(ds)` does not `skipna` in case of `nan` values

heikoklein: and maybe, we should make sure that values which are `nan` have errors which are `nan`.

pyaerocom/helpers.py
lewisblake: Resolved by a conversation :)

heikoklein (approved):
Thanks, looks good.

PR #1665, opened 2025-06-26
Small units bug

thorbjoernl:
## Change Summary

AFAICT this fixes the issues with deposition vars (eg. `drynh3`) in the EMEP evaluation, ~~though I'm still waiting for a confirmation run to see if it also fixes `concso4c`~~.

This changes behavior in the EBAS reader to pick a column arbitrarily when multiple columns have compatible units. I haven't found a case where such a case is not simply two columns that represent the same data in different units (eg. `mg S m-2` vs `mg m-2`).

It removes a test that no longer causes a unique column failure due to the above change.

## Related issue number

NA

## Checklist

* [x] Start with a draft-PR
* [x] Tests pass locally
* [x] Tests pass on CI
* [x] Make PR ready to review

### DIFF
--- a/pyaerocom/io/read_ebas.py
+++ b/pyaerocom/io/read_ebas.py
@@ -851,6 +851,8 @@ class ReadEbas(ReadUngriddedBase):
                     continue
             if len(_cols) > 0:
                 result_col = _cols
+        if len(result_col) > 1:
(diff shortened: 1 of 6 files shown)


jgriesfeller (approved):
Looking good

PR #1668, opened 2025-06-27
add missing keys to user specific paths.ini

### LINKED ISSUE #1667: Introduce update of user specific `paths.ini` file
Currently, when we introduce e.g. a new observation network, the new entries in `paths.ini` have to be added to a pontentially existing user specific `paths.ini` by the user.

We should add missing keys to a user pespecific `paths.ini` automatically using the defaults from the distribution/

jgriesfeller:
- add missing keys to user specific paths.ini
- remove case sensitivenes for keys in ini files (since a rewrite of paths.ini converts all keys to lower case (as per "standard")

## Change Summary

see above

## Related issue number

fix #1667 

## Checklist

* [x] Start with a draft-PR
* [x] Tests pass locally
* [x] Tests pass on CI
* [x] Make PR ready to review


pyaerocom/config.py
```diff
@@ -302,18 +304,71 @@ def infer_config(self):
...
+        """
+        if user_file == default_file:
+            return
+
+        # try to determine if an update of the user file is needed...
+        user_config = configparser.ConfigParser()
+        user_config.read(user_file)
+        default_config = configparser.ConfigParser()
+        default_config.read(default_file)
+        user_keys = []
```
lewisblake: I think since keys should be unique, maybe we can improve this by using sets. Then the comparison in the if statement below could be a little more precise (hopefully allowing us to remove your comment on it). Consider something like

```python
user_keys = {f"{section}_{key}" for section in user_config.sections() for key in user_config.options(section)}
default_keys = {f"{section}_{key}" for section in default_config.sections() for key in default_config.options(section)}

missing_keys = default_keys - user_keys  
extra_keys = user_keys - default_keys   

if missing_keys or extra_keys:
    # log updating the update
    # create new config starting from default
    
    # copy user-specific sections and keys
    for section in user_config.sections():
                if section not in updated_config:
                    updated_config[section] = {}
                for key in user_config.options(section):
                    # only copy keys not present in default config to preserve user customizations
                    if f"{section}_{key}" in extra_keys:
                        updated_config[section][key] = user_config[section][key]
....
```

pyaerocom/config.py
```diff
@@ -302,18 +304,71 @@ def infer_config(self):
...
+        # config, but we ignore that for now
+        if len(user_keys) != len(default_keys):
+            updated_config = configparser.ConfigParser()
+            updated_config.read([default_file, user_file])
+            logger.info(
+                f"updating user config file {user_file} with additional keys from default file."
+            )
+            backup_file_name = user_file + ".backup" + datetime.today().strftime("%Y%m%d%H%M%S")
+            os.rename(user_file, backup_file_name)
+            with open(user_file, "w") as user_file:
```
lewisblake: Consider putting in a try/except block. Perhaps even most of the meat of the code should go in a try/except block

tests/test_config.py
```diff
@@ -138,6 +150,42 @@ def test_Config_read_config():
...
+        # remove 10 lines to call the addition of non-existing keys on CI
+        lines = lines[0:-10]
+    user_file = os.path.join(const.my_pyaerocom_dir, const.PATHS_INI_NAME)
+    # we don't want to interfere with the developer's paths.ini
+    # so move that away for a sec
+    file_exist_flag = False
+    if Path(user_file).exists():
+        tmp_file = user_file + ".CI.tmp"
+        os.rename(user_file, tmp_file)
+        file_exist_flag = True
```
thorbjoernl: I'm not a fan of manipulating the user's config files like this for a test. It is always possible that a test is aborted midway through. Can we use `pytest-mock` to set the ini file path on CONST to a temporary directory for the test? Similar to how I use it here: https://github.com/metno/aerovaldb/blob/df90e038786723a349093a3226ae9423f83c95ce/tests/jsondb/test_jsonfiledb.py#L94

pyaerocom/config.py
```diff
@@ -798,8 +853,10 @@ def read_config(
         if init_data_search_dirs:
             self._search_dirs = []
 
-        cr = ConfigParser()
-        cr.optionxform = str
+        cr = configparser.RawConfigParser()
+        # this makes that all keys are converted to lower case strings so that we can
+        # check with lower case strings only later on
+        cr.optionxform = lambda option: str(option).lower()
```
thorbjoernl, line 903: If we want to preserve the user's capitalization, optionxform should also allow doing that instead, by just setting `optionxform = str`. https://stackoverflow.com/questions/1611799/preserve-case-in-configparser

pyaerocom/config.py
```diff
@@ -302,18 +304,71 @@ def infer_config(self):
...
+
+        # try to determine if an update of the user file is needed...
+        user_config = configparser.ConfigParser()
+        user_config.read(user_file)
+        default_config = configparser.ConfigParser()
+        default_config.read(default_file)
+        user_keys = []
+        default_keys = []
+        for _section in user_config.sections():
+            for key in user_config.options(_section):
```
thorbjoernl: Is `.options` equivalent to `user_config[_section].keys()`. If so I find the latter more readable.

pyaerocom/config.py
thorbjoernl: Maybe in addition to try-except we should restore the backup and inform the user that something went wrong via the logs.

thorbjoernl (changes requested):

pyaerocom/config.py
jgriesfeller: That's what was in there before. The disadvantage is that one has to match both cases later on (or even all possible mixes). Do we really want to use regular expressions here?

tests/test_config.py
jgriesfeller: changed to testing on different file names

pyaerocom/config.py
jgriesfeller: done

pyaerocom/config.py
jgriesfeller: done

pyaerocom/config.py
jgriesfeller: because a user might use older pyaerocom versions I chose not to remove additional keys. The code for that has only been commented for now though.

thorbjoernl (approved):

PR #1673, opened 2025-07-02
add comment about use_meteorological_seasons in the docs

charlienegri:
`use_meteorological_seasons` plays a role only in the computations that make use of `_select_period_season_coldata`  

since the diurnal statistics are computed in a completely different/separate way, and they never make use of  this function, this means that at the moment `use_meteorological_seasons` has no effect on those.  
Unclear if this should be changed or not, but at least a note in the docs about this is desirable at this stage.

## Checklist

* [x] Start with a draft-PR
* [x] Tests pass locally
* [x] Tests pass on CI
* [x] Make PR ready to review

### DIFF
--- a/docs/_static/aeroval/cfg_examples_example1.py
+++ b/docs/_static/aeroval/cfg_examples_example1.py
@@ -157,7 +157,8 @@ GLOB_CFG = dict(
     # (if available) and Jan/Feb of the same year, while if use_meteorological_seasons=False,
     # it will refer to data from Jan/Feb and December of the same year. (The ['all'] (whole year) key
     # for the year in question will not be affected, so it will refer to data from January 1st to
-    # December 31st.
+    # December 31st. This applies at the moment only to the statistics that make use of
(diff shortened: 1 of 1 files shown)


thorbjoernl (approved):

charlienegri:
@thorbjoernl I have now looked more closely into `_create_diurnal_weekly_data_object`,   
the relevant code is:
```
for yr in yearkeys:
        data = data_allyears.where(data_allyears["time.year"] == yr, drop=True)
        for seas in seasons:
            rep_week_ds = xr.Dataset()
            if resolution == "seasonal":
                mon_slice = data.where(data["time.season"] == seas, drop=True)
            elif resolution == "yearly":
                mon_slice = data
```
so there is a loop on years and a nested loop on seasons, which means that the point of the PR stands, because for a given year `data_allyears.where(data_allyears["time.year"] == yr, drop=True)` will cut out the D relevant for DJF of that year (relevant if the intention is use_meteorological_seasons = True)

PR #1679, opened 2025-07-07
Support for MultiGriddedData

dulte:
## Change Summary
Adds possibility to read multiple model files, e.g. uEMEP gridded city simulation. This is done through the new MultiGriddedData, which holds a list of GriddedData files. As of now, only functions that are needed for colocation using colocate_gridded_ungridded are supported.

While trying to have as small a footprint as possible, some changes to the ungridded container was needed, to make sure the regridding works correctly for both gridded and ungridded data.

Assumes model file have the same projection (or lack of projection), vars, units, ts_type and ndim. The data must also be disjoint (no overlapping).

Is tested with uEMEP and EMEP data

- Full run started:
  - Output will (eventually) be available here: https://aeroval-test.met.no/thlun/pages/evaluation/?project=uemep&experiment=uemep-2025-08-29
  - Output dir: `/lustre/storeB/users/thlun8736/python/aeroval/data/uemep/uemep-2025-08-29`

## Checklist

* [x] Start with a draft-PR
* [x] Tests pass locally
* [x] Tests pass on CI
* [x] Make PR ready to review


dulte:
Tests fail for gridded_gridded (as expected), and for aeroval/utils. The functions in aeroval/utils are never used, so will probably remove this code and its tests :wastebasket:

pyaerocom/griddeddata_container.py
```diff
@@ -0,0 +1,552 @@
...
+                ts_type=ts_type,
+                vert_which=vert_which,
+                flex_ts_type=flex_ts_type,
+                **kwargs,
+            )
+
+            self.add_griddeddata(data)
+
+    def _create_latlon_info(self, data: GriddedData) -> dict:
+        return {
```
thorbjoernl: 1. This should deal with standard name not existing as some uEMEP files don't have it (?).

2. Not sure if this is an issue on my end, but when I ran uEMEP against EBAS in EMEP config, the dimensions are lat / lon, not latitude / longitude. Maybe something needs to be more flexible or there is an issue somewhere further up.

pyaerocom/griddeddata_container.py
```diff
@@ -0,0 +1,552 @@
...
+
+        self.ndim = data.ndim
+
+        self.latlon_info = self._create_latlon_info(data)
+
+        # self._lat_res = data.lat_res
+        # self._lon_res = data.lon_res
+
+        self._lat_points = data.latitude.points
+        self._lon_points = data.longitude.points
```
thorbjoernl: Needs to also account for lat / lon names; see other comment.

pyaerocom/griddeddata.py
```diff
@@ -993,6 +993,69 @@ def transpose(self, new_order):
...
+        Returns
+        -------
+        tuple[list[tuple[float, float]]]
+            Two list, one for lats and one for lons
+
+        """
+
+        latitude = self.latitude.points
+        longitude = self.longitude.points
+        lat_range = (np.min(latitude), np.max(latitude))
```
thorbjoernl: This will also need to take into account lat / lon possibility

pyaerocom/colocation/colocation_utils.py
```diff
@@ -771,25 +767,25 @@ def latlon_proj(lat, lon):
...
-        longitude = data.longitude.points
-        xrange = [np.min(longitude), np.max(longitude)]
-        yrange = [np.min(latitude), np.max(latitude)]
+
+        if isinstance(data, GriddedData):
+            latitude = data.latitude.points
+            longitude = data.longitude.points
+            xrange = [np.min(longitude), np.max(longitude)]
+            yrange = [np.min(latitude), np.max(latitude)]
+        elif isinstance(data, GriddedDataContainer):
```
heikoklein: Would be nice to have a `get_lonlat_ranges()` in a GriddedData/GriddedDataContainer and use it just here.

(Maybe we have that already for data with projections? `data.get_xyranges()` ?)

pyaerocom/griddeddata_container.py
```diff
@@ -0,0 +1,499 @@
...
+
+
+logger = logging.getLogger(__name__)
+
+
+class GriddedDataContainerException(Exception):
+    pass
+
+
+class GriddedDataContainer:
```
heikoklein, line 19: Would be nice to have a common interface for GriddedDataContainer/GriddedData so it is clear which functions have to exist in both.

pyaerocom/griddeddata_container.py
```diff
@@ -0,0 +1,499 @@
...
+    def __init__(self, data_id: str, data: GriddedData | None = None):
+        """
+        Class for holding and working with multiple GriddedData objects.
```
heikoklein, line 22: What is the difference between `MultiGriddedData` and `GriddedDataContainer`? Are both in use?

tests/aeroval/test__processing_base.py
```diff
@@ -2,7 +2,7 @@
 
 import pytest
 
-from pyaerocom import Colocator, GriddedData, UngriddedData
+from pyaerocom import Colocator, UngriddedData, GriddedDataContainer
```
heikoklein, line 5: Should only import UngriddedDataContainer interface and test against it, not one of the implementations, e.g. UngriddedData

heikoklein (approved):
Looks good. Some ideas of improving the interface, but that can also be done at a later stage.

heikoklein (changes requested):
The evaluation of https://aeroval-test.met.no/thlun/pages/evaluation/?project=uemep&parameter=concNno2&observation=EBAS&station=Cabauw+Wielsekade&experiment=uemep-2025-08-29# show again the wrong values of Cabauw Wielsekade, (1.09 ug N/m3 instead of 4.38 ug N/m3 as in the experiment 2025-08-28-2)

pyaerocom/griddeddata.py
```diff
@@ -531,7 +531,7 @@ def lon_res(self):
             raise AttributeError("Data does not contain longitude information")
         vals = np.diff(self.longitude.points)
         val = vals.mean()
-        if vals.std() / val > 0.0001:
+        if vals.std() / val > 0.1:  # NOTE: Increased from 0.0001
```
heikoklein: why is this needed for multigridded data with projection. Checking the resolution of the 2d longitude grid does not make any sense. For lat/lon data without projection, the noise-to-signal ratio should not be changed.

heikoklein (changes requested):

thorbjoernl:
# Status

- Current config I'm working with: `/lustre/storeB/project/fou/kl/emep/Reports/2025/SourceReceptor/Evaluation/cfg_emep-status_reporting2025_uemep.py`
	- I've left some explanatory comments. See comments starting with `NOTE: `
- Most recent evaluations: 
	- https://aeroval-test.met.no/thlun/pages/evaluation/?project=uemep&experiment=uemep-2025-08-29
	- https://aeroval-test.met.no/thlun/pages/evaluation/?project=uemep&experiment=uemep-2025-08-29-2
	- The only difference is that the *second* evaluation is done on only the Netherlands (`uemep/157`, `uemep/158`) tiles.
		- **Problem:** Wrong model values for stations and regions. Since the including of extra tiles changes the station values, I suspect that data that shouldn't be included is included. May be some operation that doesn't use proj_info correctly.
			- The issue is already present in the collocated data afaict so it happens during colocation.

pyaerocom/griddeddata_container.py
dulte: No. GriddedDataContainer is meant to take over from MultiGriddedData. So I will remove all traces of the latter

pyaerocom/griddeddata_container.py
dulte: I'll make that now that I have an overview over all the functions that needs to exist in both

pyaerocom/griddeddata.py
dulte: This is a problem/fix for the uEMEP 250m. When using 250m gridding this check is not passed. I think this is due to the more irregular coords to make it fit with the gridding. It might be errors in the precision of the lat/lon or something. I need to figure out how to handle the fine resolution, and then I will change this check back

dulte:
to_time_series now makes sure that the resulting array is sorted in the same way as the incoming coords, and thus the stations in the station list for obs and mod in `colocate_gridded_ungridded` should be the same

dulte:
@heikoklein, I've removed the interpolation of single coords we talked about, and instead implemented the simplest possible sorting of the list of stationdatas, to see if that worked

This will fail for e.g. LCS, where multiple stations can have the same coords :( So I will look into your idea (colocating each tile separately) when I'm back  

Here is the result: https://aeroval-test.met.no/danielh/pages/evaluation/?project=uemep&statistic=nmb&experiment=uemep-full-2&region=ALL#

pyaerocom/griddeddata_container.py
```diff
@@ -369,69 +369,31 @@ def to_time_series(
...
-        GriddedData
-            child with has coord in domain
+    def sort_stationdatas_by_coords(
+        self, sd: list[StationData], coords: dict[str, list[float]]
+    ) -> list[StationData]:
+        def coord_index(stationdata: StationData, points):
+            lat = float(stationdata["latitude"])
+            lon = float(stationdata["longitude"])
 
+            dists2 = [(point[0] - lat) ** 2 + (point[1] - lon) ** 2 for point in points]
```
heikoklein: You don't know here if you have x/y coordinates or lat/lon coordinates. A eukledian metric as used her is only valid for x/y coordinates, while you have to use in general the haversine formula to get it working for lat/lon.

The general problem is not that to_time_series destroys the order, but that you only get the positions of tiles, and the tiles don't have a order. You know already from the bounding-box test earlier which stations belong to which tile. You should try not loose that information (in colocation_utils.py I think).

dulte:
Due to the failings of the above sorting, I've switched to sorting obs stations based on which tile they belong. 

I tried to colocate each tile separately, but that increased the runtime to much (due to having to filter the obs once per tile). The new sorting is a compromise between the first sorting method and colocation each tile separately. 

https://aeroval-test.met.no/danielh/pages/evaluation/?project=uemep&statistic=data_mean&experiment=uemep-full-3&region=ALL#

dulte:
The last commit lowers the runtime for LCS and uemep from ~24 hours to about 1.5 hours

thorbjoernl:
[Here](https://aeroval-test.met.no/thlun/pages/evaluation/?project=multigrid-test-final) is a new status run and full uemep evaluation with the most recent push.

pyaerocom/griddeddata_container.py
```diff
@@ -0,0 +1,553 @@
...
+    ) -> tuple[GriddedData, list[tuple[float, float]], list[tuple[float, float]]]:
+        tiles = []
+        xranges = []
+        yranges = []
+        for data in self.children:
+            if self.proj_info is None:
+                latitude = data.latitude.points
+                longitude = data.longitude.points
+                xrange = [np.min(longitude), np.max(longitude)]
+                yrange = [np.min(latitude), np.max(latitude)]
```
heikoklein, line 259: in hires/uemep, latitude and longitude can be huge. And latitude/longitude is in lat-lon projection guaranteed to be monotonous. So use data.latitude[0]/[-1] and get min/max of those.

pyaerocom/griddeddata_container.py
```diff
@@ -0,0 +1,553 @@
...
+            else:
+                xrange, yrange = self._get_xyrange_child(data)
```
heikoklein: confusing name with `_child` since you are in the child-loop already. `_get_xyrange` or `_get_xyrange_data`

pyaerocom/griddeddata_container.py
```diff
@@ -0,0 +1,553 @@
...
+            xrange, yrange = self._get_xyrange_child(data)
+            if xrange is None or yrange is None:
+                raise ValueError(
+                    f"x/y axis not found in cube: {data.proj_info.x_axis}, {data.proj_info.y_axis}"
+                )
+            xranges.append(xrange)
+            yranges.append(yrange)
+        return xranges, yranges
+
+    def _get_xyrange_child(self, child: GriddedData) -> tuple[tuple[float, float]] | tuple[None]:
```
heikoklein: This method is designed to be independent of self, and just work on "child". Make it a staticmethod. (and change child to data)

pyaerocom/griddeddata_container.py
```diff
@@ -0,0 +1,553 @@
...
+        xrange = None
+        yrange = None
+        for coord in child.cube.dim_coords:
+            if coord.var_name == child.proj_info.x_axis:
+                vals = coord.points
+                xrange = (np.min(vals), np.max(vals))
```
heikoklein: Don't use min/max on all values, which are guaranteed to be monotonic.

Do we have access to axes bounds? Currently we might loose half grid-cells to the right and left.

pyaerocom/griddeddata_container.py
```diff
@@ -0,0 +1,553 @@
...
+            if coord.var_name == child.proj_info.y_axis:
+                vals = coord.points
+                yrange = (np.min(vals), np.max(vals))
```
heikoklein: Same as above

pyaerocom/colocation/colocation_utils.py
```diff
@@ -971,14 +962,60 @@ def latlon_proj(lat, lon):
...
+    yranges,
+    proj,
+    unsorted_obs_stat_data,
+) -> tuple[list[StationData], list[StationData]]:
+    obs_stat_pos = np.array(
+        [
+            [*proj(station.latitude, station.longitude), station.latitude, station.longitude]
+            for station in unsorted_obs_stat_data
+        ],
+    )
```
heikoklein, line 984: obs_stat_pos has 4 axes in float and those are addressed by 0-3. Would be much nicer to have x,y,lat,lon indexes:
```python
np.array(
    [(*proj(...), lat, lon) for ...],
    np.dtype([('x', 'f8'), ('y', 'f8'), ('lat', 'f8'),('lon', 'f8')])
)
```
and access then as `obs_stat_pos["x"]` rather than `obs_stat_pos[:, 0]`

pyaerocom/colocation/colocation_utils.py
```diff
@@ -971,14 +962,60 @@ def latlon_proj(lat, lon):
...
+    obs_stat_data = []
+    grid_stat_data = []
+
+    for i, tile in enumerate(tiles):
+        xrange = xranges[i]
+        yrange = yranges[i]
+
+        found_id = np.where(
+            np.logical_and(
+                np.logical_and(obs_stat_pos[:, 0] >= xrange[0], obs_stat_pos[:, 0] < xrange[1]),
```
heikoklein: why do we exclude xrange[1]? Ranges are currently not bounds but last grid-point included. Change to bounds (see construction of range) or include.

pyaerocom/colocation/colocation_utils.py
```diff
@@ -971,14 +962,60 @@ def latlon_proj(lat, lon):
...
+                np.logical_and(obs_stat_pos[:, 1] >= yrange[0], obs_stat_pos[:, 1] < yrange[1]),
+            )
+        )[0]
```
heikoklein, line 998: Is this the same as
```python
xrange[0] <= obs_stat_pos['x'] & obs_stat_pos['x']  <= xrange[1] &
yrange[0] <= obs_stat_pos['y'] & obs_stat_pos['y'] <= yrange[1]
```
`np.logical_and` is the same as &, and `np.where` does not do anything on a logical array?
```

File: pyaerocom/colocation/colocation_utils.py [hunk 18]
```diff
@@ -971,14 +962,60 @@ def latlon_proj(lat, lon):
...
+
+        found_stations = [unsorted_obs_stat_data[i] for i in found_id]
```
heikoklein, line 1000: `unsorted_obs_stat_data[found_id]` ? Or am I missing something? Please comment

pyaerocom/colocation/colocation_utils.py
```diff
@@ -971,14 +962,60 @@ def latlon_proj(lat, lon):
...
+
+        ungridded_lats = list(obs_stat_pos[found_id, 2])
```
heikoklein: why convert to python-list?
`obs_stat_pos['lat'][found_id]`

pyaerocom/colocation/colocation_utils.py
```diff
@@ -971,14 +962,60 @@ def latlon_proj(lat, lon):
...
+        ungridded_lons = list(obs_stat_pos[found_id, 3])
+
+        if len(ungridded_lats) == 0:
```
heikoklein: abort earlier, as soon as you have done the bounds-check

pyaerocom/griddeddata.py
```diff
@@ -993,6 +993,69 @@ def transpose(self, new_order):
...
+
+        xranges = []
+        yranges = []
+
+        xrange = None
+        yrange = None
+        for coord in self.cube.dim_coords:
+            if coord.var_name == self.proj_info.x_axis:
+                vals = coord.points
+                xrange = (np.min(vals), np.max(vals))
```
heikoklein: don't use min/max but [0] and [-1] (monotonic), only check if [0] is larger or smaller than [-1]

pyaerocom/griddeddata.py
```diff
@@ -993,6 +993,69 @@ def transpose(self, new_order):
...
+        -------
+        tuple[list[tuple[float, float]]]
+            Two list, one for lats and one for lons
+
+        """
+
+        latitude = self.latitude.points
+        longitude = self.longitude.points
+        lat_range = (np.min(latitude), np.max(latitude))
+        lon_range = (np.min(longitude), np.max(longitude))
```
heikoklein: if lat/lon projection/1D, use [0]/[-1] rather than whole array.

pyaerocom/griddeddata_container.py
heikoklein: Why do we have this min/max implementation so often? Should only exist for lat/lon and x/y, i.e. twice.

pyaerocom/griddeddata_container.py
heikoklein: Please remove `sort_stationdatas_by_coords()`. It's no longer needed as far as I can see.

pyaerocom/projection_information.py
```diff
@@ -110,3 +110,15 @@ def from_xarray(ds: xarray.Dataset, var: str):
...
+        if value._crs != self._crs:
+            return False
+
+        if value._x_axis != self._x_axis:
+            return False
+        if value._y_axis != self._y_axis:
+            return False
+        if value._units != self._units:
+            return False
+        return True
```
heikoklein: Slightly dangerous because projections might be equal, even if the attribute strings are now, but ok.
Why not use python boilerplate code her:
```python
    def __eq__(self, other):
        if not isinstance(other, type(self)):
            return False
        return self.__dict__ == other.__dict__
```

tests/test_griddeddata_container.py
```diff
@@ -0,0 +1,175 @@
...
+    data_id = "test_id"
+    data = GriddedDataContainer(data_id)
+    assert data.data_id == data_id
+
+
+def test_fixture(cities_data):
+    for model in ["uEMEP", "EMEP"]:
+        data = cities_data[model]
+        for city in data:
+            assert isinstance(city, GriddedData)
```
heikoklein, line 18: Can't you test against `GriddedDataContainer` here?

tests/test_griddeddata_container.py
```diff
@@ -0,0 +1,175 @@
+import pytest
+from copy import deepcopy
+
+from pyaerocom.griddeddata_container import GriddedDataContainer, GriddedDataContainerException
+from pyaerocom import GriddedData
```
heikoklein, line 5: `GriddedData` really needed?

heikoklein (changes requested):
Great PR, fast implementation and new functionality.

Some nitbits

pyaerocom/colocation/colocation_utils.py
dulte: The where is included so that the it's the indices that are returned instead of the bool array. This is done because I couldn't find a working way of having `unsorted_obs_stat_data` (see below)

pyaerocom/colocation/colocation_utils.py
dulte: It is because it is a list, and thus I can't pass in a list/array of indices. I tried passing in `unsorted_obs_stat_data` as an array, but it it did not preserve the elements correctly as StationData, so I opted for a list

tests/test_griddeddata_container.py
dulte: No, I'm checking if all children are GriddedData

tests/test_griddeddata_container.py
dulte: See above

pyaerocom/griddeddata_container.py
dulte: removed

pyaerocom/griddeddata_container.py
dulte: Changed name to just `_get_xyrange`

pyaerocom/colocation/colocation_utils.py
dulte: Implemented this

pyaerocom/griddeddata_container.py
dulte: I've changed it to `xrange = (vals[0], vals[-1])`. The axes are not guaranteed to have bounds, sadly

pyaerocom/griddeddata_container.py
dulte: see above

dulte:
I think I've addressed all the points now. For some reason the 3.10 tests fail. I think it has to do with the version of typer. I had the bug with some other project

pyaerocom/griddeddata_container.py
heikoklein: Can you document in the docstrings what you mean by xrange then, i.e. that it is the ordered frist and last value of the axes, and not the beginning and end of the data-range? Just to be clear. (Or you could check for bounds, use them if they exist, and if not, assume value is center and extrapolate the range half a cell...)

pyaerocom/griddeddata.py
heikoklein: Is this still needed? You promised to change it back...

pyaerocom/colocation/colocation_utils.py
heikoklein: I think `&` is shorter/nicer than `np.logical_and`, and writing bound-checks in natural order, e.g. `(xrange[0] <= val) & (val <= xrange[1])` is much easier to read than your code. (in non-numpy code this would be even nicer, e.g. `xrange[0] <= val <= xrange[1]` but my example is very close to that)

pyaerocom/colocation/colocation_utils.py
heikoklein: not addressed in your review. Could be moved right after construction of found_id

pyaerocom/colocation/colocation_utils.py
heikoklein: Please add a comment that xrange is inclusive.

pyaerocom/griddeddata_container.py
dulte: This code was originally in `colocation_utils.py`, and has been there the last year. (I just moved it to be a method of griddeddata_container) So if we are going to change it's behavior (i.e. use bounds or extrapolate) then we might want to have an own PR for that, not let it be part of this big PR

heikoklein: A docstring saying what we actually do would be nice.

heikoklein:
I just merged a fix from @charlienegri to get the tests working again with newer xarray and older typer. Please try merging main-dev to this branch.

dulte:
lat_res seems to work fine with uEMEP reporting data. So the test will be skipped for now

heikoklein (approved):

PR #1693, opened 2025-07-22
Country filtering for 4D ColocatedData objects

### LINKED ISSUE #1295: Implement country filtering for 4D colocated data sets
Country filtering is not implemented for 4D colocated data sets.
https://github.com/metno/pyaerocom/blob/1a3c12e21d3708e08272a4313c1facad6cdfb87e/pyaerocom/aeroval/coldatatojson_engine.py#L130-L132

4D colocated data sets are defined as 

https://github.com/metno/pyaerocom/blob/1a3c12e21d3708e08272a4313c1facad6cdfb87e/pyaerocom/colocation/colocated_data.py#L85-L88

I would like this implemented. Needed for satellite data comparisions, e.g., in SESAM.

 - Would you be able to work on this solution yourself? kanskje
 - How can the Pyaerocom development team assist you? hjelp

**Additional context**
Regions in general need an overhaul, but this is a smaller feature which would be very helpful for satellite data comparisons. It isn't obvious to me why this is an issue, but the devil is in the details. This is what such a `ColocatedData` instance looks like:

Screenshot from 2024-08-01 20-28-39

lewisblake:
## Change Summary

Implements country filtering for 4D ColocatedData.

Seems to be working.

Screenshot from 2025-07-22 12-54-46

## Related issue number

Close #1295 

## Checklist

* [x] Start with a draft-PR
* [x] Tests pass locally
* [x] Tests pass on CI
* [x] Make PR ready to review

### DIFF
--- a/pyaerocom/colocation/colocated_data.py
+++ b/pyaerocom/colocation/colocated_data.py
@@ -547,7 +547,7 @@ class ColocatedData(BaseModel):
             dictionary of unique country names (keys) and corresponding country
             codes (values)
         """
-        if "country" not in self.coords:
+        if "country" not in list(self.coords):
             raise MetaDataError(
                 "No country information available in "
                 "ColocatedData. You may run class method "
@@ -869,19 +869,25 @@ class ColocatedData(BaseModel):
 
         """
         if self.has_latlon_dims:
-            raise DataDimensionError(
-                "Countries cannot be assigned to 4D"
-                "ColocatedData with othorgonal lat / lon "
-                "dimensions. Please consider stacking "
-                "the latitude and longitude dimensions-"
-            )
+            # raise DataDimensionError(
+            #     "Countries cannot be assigned to 4D"
+            #     "ColocatedData with othorgonal lat / lon "
+            #     "dimensions. Please consider stacking "
+            #     "the latitude and longitude dimensions-"
+            # )
+            pass
         if assign_to_dim is None:
             assign_to_dim = "station_name"
 
-        if assign_to_dim not in self.dims:
-            raise DataDimensionError("No such dimension", assign_to_dim)
+        # stack lat and lon dimensions into station_name for 4D data
+        if all(_ in self.dims for _ in ("latitude", "longitude")):
+            coldata = self if inplace else self.copy()
+            coldata = coldata.stack(station_name=("latitude", "longitude"))
 
-        coldata = self if inplace else self.copy()
+        else:
+            if assign_to_dim not in self.dims:
+                raise DataDimensionError("No such dimension", assign_to_dim)
+            coldata = self if inplace else self.copy()
 
         if "country" in coldata.data.coords:
             logger.info("Country information is available")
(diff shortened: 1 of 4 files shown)


pyaerocom/colocation/colocated_data.py
```diff
@@ -856,38 +856,31 @@ def check_set_countries(self, inplace=True, assign_to_dim=None):
...
-                "dimensions. Please consider stacking "
-                "the latitude and longitude dimensions-"
-            )
         if assign_to_dim is None:
             assign_to_dim = "station_name"
 
-        if assign_to_dim not in self.dims:
-            raise DataDimensionError("No such dimension", assign_to_dim)
+        # stack lat and lon dimensions into station_name for 4D data
+        if all(_ in self.dims for _ in ("latitude", "longitude")):
```
heikoklein: I don't like a underscore as meaningful loop variable. Underscore should only be used if it isn't used at another place. It is to easy to overlook an undescore. I would use `dim` or `d` here as understandable variable name.

pyaerocom/colocation/colocated_data.py
```diff
@@ -899,8 +892,12 @@ def check_set_countries(self, inplace=True, assign_to_dim=None):
         arr = arr.assign_coords(
             country=(assign_to_dim, countries), country_code=(assign_to_dim, codes)
         )
-        coldata.data = arr
-        return coldata
+        if inplace:
+            self.data = arr
+            return self
+        else:
+            coldata = arr
```
heikoklein: This looks wrong? coldata is in most cases something like self, so I would expect it to be `coldata.data = arr` for symmetry.

pyaerocom/colocation/colocated_data.py
```diff
@@ -856,38 +856,31 @@ def check_set_countries(self, inplace=True, assign_to_dim=None):
...
         if assign_to_dim is None:
             assign_to_dim = "station_name"
 
-        if assign_to_dim not in self.dims:
-            raise DataDimensionError("No such dimension", assign_to_dim)
+        # stack lat and lon dimensions into station_name for 4D data
+        if all(_ in self.dims for _ in ("latitude", "longitude")):
+            coldata = self if inplace else self.copy()
+            logger.info("Stacking lat and lon dimensions into station_name")
+            coldata = coldata.stack(station_name=("latitude", "longitude"))
```
heikoklein: The inplace code was very clean (one line), but is now a mess, merged all over this function.

Do we need this function inplace/non-inplace functionality? Do we have tests for it (I guess it uses mostly the default==inplace)? Maybe we should remove?

heikoklein (commented):
The functionality is well implemented. For aesthetic reasons I don't like the use of the _ when it is used. I'm worried about the too complicated implementation of inplace=False, which might be unneccessary.

pyaerocom/colocation/colocated_data.py
lewisblake: Doesn't look like it's used anywhere else. We could just get rid of it, but that changes the interface. Happy to get rid of it if you want.

heikoklein: Feel free to remove it.

heikoklein (approved):
Thanks, looking good.

PR #1702, opened 2025-07-29
Adds gt to region sorting

dulte:
## Change Summary

Fixes how the region sorting is done by just adding a dunder gt to RegionNames

## Related issue number

Albania was before ALL in regions.json

## Checklist

* [x] Start with a draft-PR
* [x] Tests pass locally
* [x] Tests pass on CI
* [x] Make PR ready to review

### DIFF
--- a/pyaerocom/region.py
+++ b/pyaerocom/region.py
@@ -360,6 +360,9 @@ class RegionName(str):
         if str(self) == ALL_REGION_NAME:
             return True
 
+        if str(other) == ALL_REGION_NAME:
+            return False
+
         return str(self).lower() < str(other).lower()
 
     def __gt__(self, other) -> bool:
@@ -368,5 +371,8 @@ class RegionName(str):
 
         if str(self) == ALL_REGION_NAME:
             return False
+
+        if str(other) == ALL_REGION_NAME:
+            return True
 
         return str(self).lower() > str(other).lower()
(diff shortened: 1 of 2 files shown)


thorbjoernl:
Given the use of the `total_ordering` decorator, it shouldn't be necessary to implement \_\_gt\_\_ explicitly (https://docs.python.org/3/library/functools.html#functools.total_ordering). Something else is going on here, I suspect.

thorbjoernl:
Actually, now thinking about it, it's probably that `str` already implements the other dunders, so total_ordering doesn't override.

>This decorator makes no attempt to override methods that have been declared in the class or its superclasses. Meaning that if a superclass defines a comparison operator, total_ordering will not implement it again, even if the original method is abstract.

thorbjoernl (changes requested):
Two minor nitpics, otherwise looks good:

- Remove `@totalordering`.
- This leaves the other comparison dunders as the ones inherited from str, which is inconsistent.

charlienegri:
> Given the use of the `total_ordering` decorator, it shouldn't be necessary to implement __gt__ explicitly (https://docs.python.org/3/library/functools.html#functools.total_ordering). Something else is going on here, I suspect.

changing this line https://github.com/metno/pyaerocom/blob/main-dev/pyaerocom/region.py#L360 of the original code to `if str(other) == ALL_REGION_NAME:` works for me

dulte:
@thorbjoernl I think you are correct about gt. So if we add a small for the if statement for `str(other) == ALL_REGION_NAME`, then I think it will work. I added a test for RegionName, which I think will cover every perturbation of >, <, self and other

thorbjoernl (approved):
LGTM. Good to have a test!

PR #1712, opened 2025-08-08
Emep units patch

thorbjoernl:
## Change Summary

I hope this fixes the conversion factor issue with EBAS reading.

- EBAS only test run: https://aeroval-test.met.no/emep_evaluation/pages/evaluation/?project=emep2024-status-tl&experiment=testrun-2025-08-12
- Looks fine compared to current status run.

## Related issue number

EMEP status reporting: units issue.

## Checklist

* [x] Start with a draft-PR
* [x] Tests pass locally
* [x] Tests pass on CI
* [x] Make PR ready to review

### DIFF
--- a/pyaerocom/units/units.py
+++ b/pyaerocom/units/units.py
@@ -234,6 +234,7 @@ class Unit:
                 raise ValueError(
                     f"cfunit '{self._cfunit}' is not convertible to cfunit '{other._cfunit}'."
                 )
+            return
 
         if not compatible_element:
             raise ValueError(
@@ -245,12 +246,16 @@ class Unit:
                 raise ValueError(
                     f"cfunit '{self._cfunit}' is not convertible to cfunit '{other._cfunit}'."
                 )
+            return
 
-        if (self._element == other._element) and same_variable:
+        if (self._element == other._element) and same_variable and same_species:
             if not self._cfunit.is_convertible(other._cfunit):
                 raise ValueError(
                     f"cfunit '{self._cfunit}' is not convertible to cfunit '{other._cfunit}'."
                 )
+            return
+
+        raise ValueError()
 
     def is_convertible(self, other: str | Unit) -> bool:
         """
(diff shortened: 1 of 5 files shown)


pyaerocom/units/units.py
```diff
@@ -374,7 +381,7 @@ def convert(
             assert isinstance(other, Unit)
```
heikoklein: Since you check already with if/else I would have 
```python

        if isinstance(other, Unit):
            to_unit = other
        elif isinstance(other, str):
            to_unit = Unit(str(other), **kwargs)
        else:
            throw AssertionError(f"other must be str|Unit, but is {type(other)}")

File: pyaerocom/units/units.py [hunk 2]
```diff
@@ -234,6 +234,7 @@ def _validate_convertible(self, other: str | Unit) -> None:
                 raise ValueError(
                     f"cfunit '{self._cfunit}' is not convertible to cfunit '{other._cfunit}'."
                 )
+            return
```
heikoklein: `positive if`, then `negative if` and then a `return` statement are hard to read and very fragile.

`same_species` and `compatble_element` is calculated even if it is `cf_units only`. Maybe those should be functions?

Remove the return statements but put everything into `if/elif` statements?

heikoklein (approved):
Good you found the problem. I suggest some code-cleanup to make the code easier to read. Otherwise approved.

heikoklein (approved):

pyaerocom/units/units.py
```diff
@@ -234,26 +234,20 @@ def _validate_convertible(self, other: str | Unit) -> None:
...
                 )
-            return
-
-        if (self._element == other._element) and same_variable:
+        elif (self._element == other._element) and same_variable:
             if not self._cfunit.is_convertible(other._cfunit):
                 raise ValueError(
                     f"cfunit '{self._cfunit}' is not convertible to cfunit '{other._cfunit}'."
                 )
-            return
```
heikoklein: shouldn't there be a `else` before the final `raise valueError`?

PR #1744, opened 2025-10-21
Cams282 cli weprofiles

dulte:
## Change Summary

Adds different tools for cams2_82, such as an CLI, as well as fixes to eprofiles that makes it work with GriddedDataContainer

## Checklist

* [x] Start with a draft-PR
* [x] Tests pass locally
* [x] Tests pass on CI
* [x] Make PR ready to review


pyaerocom/data/paths.ini
```diff
@@ -90,6 +90,7 @@ AIR_NOW = ${BASEDIR}/aerocom/aerocom1/AEROCOM_OBSDATA/MACC_INSITU_AirNow
 MARCO_POLO = ${BASEDIR}/aerocom/aerocom1/AEROCOM_OBSDATA/CHINA_MP_NRT
 CNEMC = ${BASEDIR}/aerocom/aerocom1/AEROCOM_OBSDATA/MEP/aggregated/
 ICOS = ${BASEDIR}/aerocom/aerocom1/AEROCOM_OBSDATA/ICOS/aggregated/
+; ICOS = ${BASEDIR}/aerocom/aerocom1/AEROCOM_OBSDATA/ICOS/NRT/
```
heikoklein: Why is that comment here? Maybe remove, since it can't be used together with the line above?

pyaerocom/io/cams2_82/reader.py
```diff
@@ -0,0 +1,556 @@
...
+        "molmass": get_molmass("no2"),
+        "fromunit": "kg kg**-1",
+    },
+    "go3": {
+        "molmass": get_molmass("o3"),
+        "fromunit": "kg kg**-1",
+    },
+}
+
+TO_VMR = {
```
heikoklein, line 150: can't you use the units module for the conversions, e.g. `28.9644 = get_molmass("air_dry")`, and there is a function
```
get_mmr_to_vmr_fac()
```

pyaerocom/aux_var_helpers.py
```diff
@@ -780,6 +780,45 @@ def vmrx_to_concx(data, p_pascal, T_kelvin, vmr_unit, mmol_var, mmol_air=None, t
...
+        kg m-3. The default is None.
+
+    Returns
+    -------
+    float or ndarray
+        input data converted to mass concentration
+
+    """
+
+    Rspecific = 287.058  # J kg-1 K-1
```
heikoklein: 3rd time this constant is specified in this file, 8th time in pyaerocom. Can you put that value into a constant?

pyaerocom/scripts/cams2_82/cli.py
```diff
@@ -0,0 +1,239 @@
...
+import multiprocessing as mp
+from copy import deepcopy
+from collections.abc import Iterator
+from datetime import date, datetime, timedelta
+from pathlib import Path
+from typing import Optional
+
+import typer
+
+from pyaerocom import change_verbosity, const
```
heikoklein: const is deprecated
```
Use of pyaerocom.const is deprecated. Please use Config.get_instance() or pyaerocom.config instead.
```
(I know, we are still using it everywhere else, but we shouldn't add it in new places.)

pyaerocom/io/read_eprofile.py
```diff
@@ -1,28 +1,24 @@
...
+from collections.abc import Iterator
+from glob import glob
+from pathlib import Path
 
 import numpy as np
 import xarray
-from glob import glob
 from tqdm import tqdm
 
 from pyaerocom import const
```
heikoklein: deprecated, see below

pyaerocom/scripts/cams2_82/evaluation.py
```diff
@@ -0,0 +1,145 @@
+from __future__ import annotations
+
+import logging
+from concurrent.futures import ProcessPoolExecutor, as_completed
+from datetime import date, timedelta
+from enum import Enum
+from pathlib import Path
+from pprint import pformat
+
+from pyaerocom import const
```
heikoklein: deprecated

pyaerocom/aux_var_helpers.py
heikoklein: maybe to `pyaerocom/units/constants.py`

pyaerocom/aux_var_helpers.py
```diff
@@ -780,6 +780,45 @@ def vmrx_to_concx(data, p_pascal, T_kelvin, vmr_unit, mmol_var, mmol_air=None, t
     return conc
 
 
+def mmrx_to_concx(data, p_pascal, T_kelvin, mmr_unit, to_unit=None):
```
heikoklein: typing info, like
data: float|np.typing.ArrayLike
then you don't need to specify it in the docstring any longer

heikoklein (approved):
Looks good. Please have reuse constants as much as possible. And write typing information in the signature rather than in the docstring.

And `from pyaerocom import const` is forbidden in any file you touch. (Maybe somebody should remove it for all files so it is no longer there...)

pyaerocom/__init__.py
```diff
@@ -29,9 +29,7 @@
 try:
     # Opt in to new defaults introduced by xarray 2025.08.0
     # https://docs.xarray.dev/en/stable/whats-new.html#v2025-08-0-august-14-2025
-    xarray.set_options(
-        use_new_combine_kwarg_defaults=True, netcdf_engine_order=["netcdf4", "scipy", "h5netcdf"]
-    )
+    xarray.set_options(use_new_combine_kwarg_defaults=True)
```
heikoklein: I think it is safer to keep the engine-order fixed. We have not excluded the v2025-8 release, and we haven't changed the comment.

heikoklein (approved):

pyaerocom/__init__.py
charlienegri: but this is how I got deployment to conda to pass...

heikoklein: Maybe we need a switch to check for xarray versions v2025-8? I guess they have removed `netcdf_engine_order`?

charlienegri: the version that needed the extra argument was the one from [september](https://docs.xarray.dev/en/stable/whats-new.html#v2025-09-1-september-29-2025), the comment about 2025-08 is from Thorbjørn and pertains a previous [commit](https://github.com/metno/pyaerocom/commit/83907c362a8d0037ed3ce8f027b2b9fb3611a206#diff-bfd3bde85cffe24de1fa3d3252abbb68022779325933193a1f839f04c23ebf5cR28-R29), introducing the try/except block entirely   
in the [latest](https://docs.xarray.dev/en/stable/whats-new.html#v2025-10-0-october-6-2025) version just released the change in default engine is reversed but the `netcdf_engine_order`  argument is still usable, so I am not sure what exactly breaks the conda deployment with having `netcdf_engine_order=["netcdf4", "scipy", "h5netcdf"]` ... 
or maybe I just hit a totally unrelated glitch when the CI ran? I will try to reinstate the argument and see if conda passes or fails, if it passes it was a glitch

charlienegri: yep, it was just a one-time failure :man_shrugging:   
this is now ready to go then :rocket:

charlienegri:
addressed https://github.com/metno/pyaerocom/issues/1747 as per discussion in pyaerocom meeting

PR #1766, opened 2025-11-18
Adds possibility for descriptions for obs and mod

dulte:
## Change Summary
Adds option to model and obs entries where the user can add a description of that entry. This description will be passed all the way to the outputted config, where the frontend can pick it up

## Checklist

* [x] Start with a draft-PR
* [x] Tests pass locally
* [x] Tests pass on CI
* [x] Make PR ready to review

### DIFF
--- a/pyaerocom/aeroval/experiment_output.py
+++ b/pyaerocom/aeroval/experiment_output.py
@@ -836,7 +836,7 @@ class ExperimentOutput(ProjectOutput):
                     f"Invalid entry: model {mod_name} ({mod_var}), obs {obs_name} ({obs_var}) ⚠️"
                 )
 
-        return Menu.model_validate(new).model_dump()
+        return Menu.return_validated(new)
 
     def _sort_menu_entries(self, avail: dict) -> dict:
         """
--- a/pyaerocom/aeroval/experiment_output.py
+++ b/pyaerocom/aeroval/experiment_output.py
@@ -801,6 +801,8 @@ class ExperimentOutput(ProjectOutput):
                     ocfg = self.cfg.obs_cfg.get_entry(src_name)
                     longname = ocfg.longname
                     webname = ocfg.web_interface_name
+                    if webname is None:
+                        continue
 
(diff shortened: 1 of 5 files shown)


pyaerocom/aeroval/experiment_output.py
```diff
@@ -794,9 +794,25 @@ def _create_menu_dict(self) -> dict:
...
+                            raise ValueError(
+                                f"Different longnames given for same web interface name: {longname} != {obs_longname[webname]}"
+                            )
+                    else:
+                        obs_longname[webname] = longname
+
                 var = mcfg.get_varname_web(mod_var, obs_var)
                 if var not in new:
                     new[var] = self._init_menu_entry(var)
+                    new[var]["obs_longnames"] = obs_longname
```
heikoklein: Why is this plural while all the others are singular? And it doesn't seem to be a list, so "obs_longname" seems to be right?

heikoklein (approved):
Thanks. It would be good if we could get a documentation of the expected output somewhere. And maybe a test? Or full validated / pydantic output... It's Chrismas soon, so time for wishes :-)

pyaerocom/aeroval/experiment_output.py
dulte: It is a dict with the possibility of many obs networks.

dulte:
> Thanks. It would be good if we could get a documentation of the expected output somewhere. And maybe a test? Or full validated / pydantic output... It's Chrismas soon, so time for wishes :-)

I will try my hand on making a pydantic validation for this output :santa: Then we can (hopefully) start to define the rest of the output as well in other PRs

dulte:
@heikoklein I've added some pydantic models for menu.json if you want to have a look

heikoklein:
Thanks, nicely implemented validation in an external class. 

I would have added a method 'return_validated()' to the output-model, just to hide any pydantic-interna - at some point (in 10 years?) we might want to switch to a different validation-framework.

PR #1773, opened 2025-12-18
Fix 1770

### LINKED ISSUE #1770: O3mda8 model maps
In the CAMS policy support quarterly reports, the AeroVal maps from the CAMS2_83 project (regional evaluation) are being used. 

- User request: add O3mda8 variable to the list of computed maps
- Challenge: O3mda8 is computed at the stations only, and not at the gridded data level
- Suggested solution: use nco/cdo tools to precompute O3mda8 as a new model variable, and add to current evaluation

charlienegri:
## Change Summary

conco3mda calculation in the cams283 reader to be consumed by the map engine 

doing the calculation of conco3mda8 directly in modelmaps_engine.py, say after conco3 model data has been read https://github.com/metno/pyaerocom/blob/main-dev/pyaerocom/aeroval/modelmaps_engine.py#L173, poses the problem that the returned object is a GriddedDataContainer that is based on an iris cube for cams2_83 but might be different for different cases

this branch is an attempt to put the conco3mda8 calculation logic into the (cams2-83) reader, and in the map engine just open up for the possibility that the returned object from _read_model_data might be a list of GriddedDataContainers

this is ugly and very slow

the calculation of conco3mda8 is otherwise supposed to be coherent with the one baked in the colocation

the timestamps of the geojson files are brute-force shifted by 1h which is necessary because aeroval expects them to be identical to the timestamps in the ts files

## Related issue number

fixes #1770 

## Checklist

* [x] Start with a draft-PR
* [ ] The PR title is a good summary of the changes
* [ ] Tests pass locally
* [ ] Tests pass on CI
* [ ] Make PR ready to review

### DIFF
--- a/pyaerocom/aeroval/experiment_output.py
+++ b/pyaerocom/aeroval/experiment_output.py
@@ -780,7 +780,7 @@ class ExperimentOutput(ProjectOutput):
                     )
                     if not first_with_mod_name:  # should already be taken care of in new
                         continue
-                    obs_name = first_with_mod_name[0]
+                    obs_name = self.cfg.obs_cfg.get_entry(first_with_mod_name[0]).obs_name
                     all_combinations.remove(first_with_mod_name)
                 else:
                     logger.warning(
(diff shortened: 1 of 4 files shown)


charlienegri:
approach dropped in favor of #1774

PR #1774, opened 2025-12-18
Fix 1770, O3mda8 model maps, selectively for countour maps

### LINKED ISSUE #1770: O3mda8 model maps

charlienegri:
## Change Summary

calculation of countour model maps for conco3mda8 in the map engine

## Related issue number

fixes #1770 

## Checklist

* [x] Start with a draft-PR
* [x] Tests pass locally
* [x] Tests pass on CI
* [x] Make PR ready to review

### DIFF
--- a/pyaerocom/aeroval/modelmaps_engine.py
+++ b/pyaerocom/aeroval/modelmaps_engine.py
@@ -52,7 +52,7 @@ def calco3mda8(data: GriddedData) -> list[GriddedData]:
         .mean("time")
         .resample(time="24h", origin="start_day", label="left", offset="1h")
         .reduce(lambda x, axis: np.apply_along_axis(min_periods_max, 0, x, min_periods=18))
-        .dropna("time")
+        # .dropna("time") # this is too slow
         .rename("conco3mda8")
         .assign_attrs(
             long_name="conco3mda8",
--- a/pyaerocom/aeroval/modelmaps_engine.py
+++ b/pyaerocom/aeroval/modelmaps_engine.py
@@ -45,11 +45,7 @@ MODELREADERS_USE_MAP_FREQ = ["ReadMscwCtm"]  # , "ReadCAMS2_83"]
 
 
 def calco3mda8(data: GriddedData) -> list[GriddedData]:
-    try:
-        data2 = data.to_xarray()
-    except Exception as e:
-        logger.warning(f"Could not convert data to xarray: {e}")
-        return [data]
+    data2 = data.to_xarray()
 
     o3mda8 = (
         data2.rolling(time=8, center=False, min_periods=6)
@@ -226,7 +222,12 @@ class ModelMapsEngine(ProcessingEngine, DataImporter):
         var_ranges_defaults = self.cfg.var_scale_colmap
         freq = self._get_maps_freq()
 
-        if var == "conco3" and read_data.ts_type == "hourly" and compute_conco3mda8_contours:
+        if (
+            var == "conco3"
+            and isinstance(read_data, GriddedData)
+            and read_data.ts_type == "hourly"
+            and compute_conco3mda8_contours
+        ):
             datalist = calco3mda8(read_data)
         else:
             datalist = [read_data]
--- a/pyaerocom/aeroval/modelmaps_engine.py
+++ b/pyaerocom/aeroval/modelmaps_engine.py
@@ -151,7 +151,12 @@ class ModelMapsEngine(ProcessingEngine, DataImporter):
                         model_name, False
                     )
                 if CONTOUR in self.cfg.modelmaps_opts.plot_types or make_contour:
-                    self._process_contour_map_var(model_name, var, self.reanalyse_existing)
+                    self._process_contour_map_var(
+                        model_name,
+                        var,
+                        self.reanalyse_existing,
+                        self.cfg.processing_opts.compute_conco3mda8_contours,
+                    )
 
                 if OVERLAY in self.cfg.modelmaps_opts.plot_types or make_overlay:
                     # create overlay (pixel) plots
@@ -179,7 +184,9 @@ class ModelMapsEngine(ProcessingEngine, DataImporter):
             data = data.extract_surface_level()
         return data
 
-    def _process_contour_map_var(self, model_name, var, reanalyse_existing):  # pragma: no cover
+    def _process_contour_map_var(
+        self, model_name, var, reanalyse_existing, compute_conco3mda8_contours
+    ):  # pragma: no cover
         """
         Process model data to create map geojson files
 
@@ -203,7 +210,7 @@ class ModelMapsEngine(ProcessingEngine, DataImporter):
             If model/var data cannot be read
         """
         try:
-            data = self._read_model_data(model_name, var)
+            read_data = self._read_model_data(model_name, var)
         except Exception as e:
             raise ModelVarNotAvailable(
                 f"Cannot read data for model {model_name} (variable {var}): {e}"
@@ -212,10 +219,10 @@ class ModelMapsEngine(ProcessingEngine, DataImporter):
         var_ranges_defaults = self.cfg.var_scale_colmap
         freq = self._get_maps_freq()
 
-        if var == "conco3" and data.ts_type == "hourly":
-            datalist = calco3mda8(data)
+        if var == "conco3" and read_data.ts_type == "hourly" and compute_conco3mda8_contours:
+            datalist = calco3mda8(read_data)
         else:
-            datalist = [data]
+            datalist = [read_data]
 
         idx = 0
         for data in datalist:
--- a/pyaerocom/aeroval/modelmaps_engine.py
+++ b/pyaerocom/aeroval/modelmaps_engine.py
@@ -63,7 +63,7 @@ def calco3mda8(data: GriddedData) -> list[GriddedData]:
     griddedo3mda8 = GriddedData(
         cubeo3mda8,
         var_name="conco3mda8",
-        ts_type=data.ts_type,
+        ts_type="daily",
         check_unit=True,
         convert_unit_on_init=True,
     )
--- a/pyaerocom/aeroval/modelmaps_engine.py
+++ b/pyaerocom/aeroval/modelmaps_engine.py
@@ -1,8 +1,10 @@
 import glob
 import logging
+from datetime import datetime, timedelta
 
 import aerovaldb
 import xarray as xr
+import numpy as np
 
 from pyaerocom import ColocatedData, GriddedData, GriddedDataContainer, TsType, __version__, const
 from pyaerocom.aeroval._processing_base import DataImporter, ProcessingEngine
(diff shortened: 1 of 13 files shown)


charlienegri:
Screenshot from 2026-01-08 09-13-16

charlienegri:
@dulte @heikoklein this approach makes it possible to compute the maps for a single full season, which is what Svetlana needs if I understood correctly, but for 11 seasons it takes close to 40h so I would not use the flag in that case... 

if you see a smarter way of implementing it let me know

pyaerocom/aeroval/modelmaps_engine.py
```diff
@@ -27,14 +36,41 @@
...
+        .assign_attrs(
+            long_name="conco3mda8",
+            species="O3 MDA8",
+        )
+    )
+    cubeo3mda8 = o3mda8.to_iris()
+    griddedo3mda8 = GriddedData(
+        cubeo3mda8,
+        var_name="conco3mda8",
+        ts_type=data.ts_type,
```
heikoklein: why is ts_type = data.ts_type? I would expect hourly input and daily output?

pyaerocom/aeroval/modelmaps_engine.py
```diff
@@ -27,14 +36,41 @@
...
 
 
+def calco3mda8(data: GriddedData) -> list[GriddedData]:
+    data2 = data.to_xarray()
+
+    o3mda8 = (
+        data2.rolling(time=8, center=False, min_periods=6)
+        .mean("time")
+        .resample(time="24h", origin="start_day", label="left", offset="1h")
+        .reduce(lambda x, axis: np.apply_along_axis(min_periods_max, 0, x, min_periods=18))
```
heikoklein: I find the code in stats/mda8/mda8.py easier readable:
```mda8 = _daily_max(_rolling_average_8hr(data))```

Maybe you can reuse that part? (and/or put it it there)?

**Review by heikoklein — COMMENTED** [review]
I don't have any better ideas, but I don't understand why the mda8 calculation of 3d data should take 40hours for a year? Maybe you have a simple test-case where we can check performance?

Possible pitfalls might be that the dataset is computed several times? An hourly dataset will be too much for memory, while a daily dataset should work fine, so maybe try to compute before moving back to iris?

Maybe try some dask parallellization? (also easier to test with a simpler test-case)

**Comment by charlienegri** [comment]
> I don't have any better ideas, but I don't understand why the mda8 calculation of 3d data should take 40hours for a year? Maybe you have a simple test-case where we can check performance?
> 
> Possible pitfalls might be that the dataset is computed several times? An hourly dataset will be too much for memory, while a daily dataset should work fine, so maybe try to compute before moving back to iris?
> 
> Maybe try some dask parallellization? (also easier to test with a simpler test-case)

40h is the total runtime for all models and 11 season which is almost 3 years... 
conco3mda calculation for a single model and 11 seasons (33 months) is something like a little under 2h if I remember correctly from that test..
I think I have leveraged the lazy loading as possible but maybe I am not seeing some easy optimization
I can further test performance with a simple case.. 
but anyway for 1 season is doable and for a single model and 1 year might also be doable... the code is stiil ugly tho,

File: pyaerocom/aeroval/modelmaps_engine.py (hunk 1 above)
Comment (charlienegri) [reply pyaerocom/aeroval/modelmaps_engine.py]: yes but the the guard on input ts_type is here https://github.com/metno/pyaerocom/blob/4c3bda678d53e7c74e8c0954ad586cd5c6eb5cf5/pyaerocom/aeroval/modelmaps_engine.py#L228

File: pyaerocom/aeroval/modelmaps_engine.py (hunk 2 above)
Comment (charlienegri) [reply pyaerocom/aeroval/modelmaps_engine.py]: cannot be reused directly, at least the _daily_max part because of different structure of the object handled, and so the axis to apply the lambda function to is different (if I got it right...)

**Comment by charlienegri** [comment]
> > I don't have any better ideas, but I don't understand why the mda8 calculation of 3d data should take 40hours for a year? Maybe you have a simple test-case where we can check performance?
> > Possible pitfalls might be that the dataset is computed several times? An hourly dataset will be too much for memory, while a daily dataset should work fine, so maybe try to compute before moving back to iris?
> > Maybe try some dask parallellization? (also easier to test with a simpler test-case)
> 
> 40h is the total runtime for all models and 11 season which is almost 3 years... conco3mda calculation for a single model and 11 seasons (33 months) is something like a little under 2h if I remember correctly from that test.. I think I have leveraged the lazy loading as possible but maybe I am not seeing some easy optimization I can further test performance with a simple case.. but anyway for 1 season is doable and for a single model and 1 year might also be doable... the code is stiil ugly tho,

Screenshot from 2026-01-15 20-27-29

**Comment by heikoklein** [comment]
> > 
> > 40h is the total runtime for all models and 11 season which is almost 3 years... conco3mda calculation for a single model and 11 seasons (33 months) is something like a little under 2h if I remember correctly from that test.. I think I have leveraged the lazy loading as possible but maybe I am not seeing some easy optimization I can further test performance with a simple case.. but anyway for 1 season is doable and for a single model and 1 year might also be doable... the code is stiil ugly tho,
> 
> Screenshot from 2026-01-15 20-27-29

From this I see that the maps for a normal component takes ~20min, while mda8/conc03 takes ~115min. Considering 24x the amount of data which needs to be read, this looks pretty good. Or are all variables read hourly?

**Comment by charlienegri** [comment]
> > > 40h is the total runtime for all models and 11 season which is almost 3 years... conco3mda calculation for a single model and 11 seasons (33 months) is something like a little under 2h if I remember correctly from that test.. I think I have leveraged the lazy loading as possible but maybe I am not seeing some easy optimization I can further test performance with a simple case.. but anyway for 1 season is doable and for a single model and 1 year might also be doable... the code is stiil ugly tho,
> > 
> > 
> > Screenshot from 2026-01-15 20-27-29
> 
> From this I see that the maps for a normal component takes ~20min, while mda8/conc03 takes ~115min. Considering 24x the amount of data which needs to be read, this looks pretty good. Or are all variables read hourly?

all data is read hourly so there is no extra reading time at all, it's all purely computation alas..

File: pyaerocom/aeroval/modelmaps_engine.py (hunk 2 above)
Comment (heikoklein) [reply pyaerocom/aeroval/modelmaps_engine.py]: Ah, I see, `mean("time")` <-> `mean()` and `np.apply_along_axis(min_periods_max, 0, x, min_periods=18)` <-> `np.apply_along_axis(min_periods_max, 1, x, min_periods=18)`

What I like in https://github.com/metno/pyaerocom/blob/4c3bda678d53e7c74e8c0954ad586cd5c6eb5cf5/pyaerocom/stats/mda8/mda8.py#L91 is that the 2 map-reduce operations are split and somehow documented by function names. Putting your code in the same module as e.g. `mda8_3d` and `rolling_average_8h_3d` would make the code clearer and would show, that you tried to implement exactly the same mda8 as we do for collocated data.

File: pyaerocom/aeroval/modelmaps_engine.py (hunk 1 above)
Comment (heikoklein) [reply pyaerocom/aeroval/modelmaps_engine.py]: I don't understand this. The guard is called on the conco3 data, and first afterwards the griddedo3mda8 data is generated. griddedo3mda8 is "daily" data, while data=read_data=conco3 is "hourly".

File: pyaerocom/aeroval/modelmaps_engine.py (hunk 2 above)
Comment (charlienegri) [reply pyaerocom/aeroval/modelmaps_engine.py]: sure, this we can do , I will refactor the code

**Comment by heikoklein** [comment]
> > 
> > From this I see that the maps for a normal component takes ~20min, while mda8/conc03 takes ~115min. Considering 24x the amount of data which needs to be read, this looks pretty good. Or are all variables read hourly?
> 
> all data is read hourly so there is no extra reading time at all, it's all purely computation alas..

If all data is read hourly, and all data is resampled to daily plots, then 6 times the time just for a rolling average resampling sounds a lot (pure gutt-feeling, no proof yet). Or is conco3 also plottet as other things (o3dailymax/o3dailymean)? Do you have an example input-file?

File: pyaerocom/aeroval/modelmaps_engine.py (hunk 1 above)
Comment (charlienegri) [reply pyaerocom/aeroval/modelmaps_engine.py]: I misread your question sorry, you are right, the ts_type there should be hard-coded as 'daily'

**Comment by charlienegri** [comment]
> > > From this I see that the maps for a normal component takes ~20min, while mda8/conc03 takes ~115min. Considering 24x the amount of data which needs to be read, this looks pretty good. Or are all variables read hourly?
> > 
> > 
> > all data is read hourly so there is no extra reading time at all, it's all purely computation alas..
> 
> If all data is read hourly, and all data is resampled to daily plots, then 6 times the time just for a rolling average resampling sounds a lot (pure gutt-feeling, no proof yet). Or is conco3 also plottet as other things (o3dailymax/o3dailymean)? Do you have an example input-file?

the rolling average alone is quite fast, based on my tests, it's the rest that is slow... 
a way to run a test is to use the cli with type season and whatever time window you want, example 
```
cams2_83 forecast season 2025-11-01 2025-11-30 --model-path /lustre/storeB/project/fou/kl/CAMS2_83/model --obs-path /lustre/storeB/project/fou/kl/CAMS2_83/obs --data-path /lustre/storeB/users/heikok/something/data --coldata-path /lustre/storeB/users/heikok/something/coldata --cache /lustre/storeB/users/heikok/something/_cache_test --name 'TEST' --id test-conco3mda-contours --description 'test'  -p 2 --onlymap --conco3mda8contours
 ```
 (note that this will not produce an experiment output valid for aeroval)

charlienegri:
I am testing the latest code with as much overlap as possible with the mda8 calculation already existing and it's even slower.... I think the reason is that the time dimension shift + filtering baked into _calc_mda8 are too expensive, even if they make sense also for the mapengine calculation... 
Screenshot from 2026-01-16 15-01-50
for 1 season is ~ 9 min per model   
I am not sure how much slower, may also be very little, I'll do some more testing and then refactor again if significant

pyaerocom/stats/mda8/mda8.py
```diff
@@ -109,6 +113,11 @@ def _rolling_average_8hr(arr: xr.DataArray) -> xr.DataArray:
...
-        lambda x, axis: np.apply_along_axis(min_periods_max, 1, x, min_periods=18)
-    )
+    if "data_source" in arr.dims:  # colocated data object
+        return arr.resample(time="24h", origin="start_day", label="left", offset="1h").reduce(
+            lambda x, axis: np.apply_along_axis(min_periods_max, 1, x, min_periods=18)
+        )
+    else:  # calculation of mda8 in the map engine, griddeddata
+        return arr.resample(time="24h", origin="start_day", label="left", offset="1h").reduce(
+            lambda x, axis: np.apply_along_axis(min_periods_max, 0, x, min_periods=18)
+        )
```
heikoklein: Maybe nicer like this?

```suggestion
    if "data_source" in arr.dims:
        # colocated data object has time-axis at axis 1
        t_axis = 1
    else:
        # usual data has time-axis at axis 0
        t_axis = 0
    
    return arr.resample(time="24h", origin="start_day", label="left", offset="1h").reduce(
        lambda x, axis: np.apply_along_axis(min_periods_max, t_axis, x, min_periods=18)
    )
```

pyaerocom/aeroval/modelmaps_engine.py
```diff
@@ -27,14 +35,36 @@
     VariableDefinitionError,
     VarNotAvailableError,
 )
+from pyaerocom.stats.mda8.mda8 import _calc_mda8
```
heikoklein: In theory, functions starting with _ should not be considered private and not be imported (except in tests). Consider turning `_calc_mda8` to `calc_mda8` to make it officially public.

heikoklein (approved):
Nice improvements in readability.

charlienegri:
timing with latest code for 11 seasons for 1 model is ~ 1.5h
Screenshot from 2026-01-18 18-00-28

heikoklein:
> timing with latest code for 11 seasons for 1 model is ~ 1.5h
This is better than the 115min you had before, but also the concno2 numbers are reduced from 9min to 6min. Unless you have made considerable code-changes (last changes were more about readability), I would rather say the performance improvement come from a faster node?

charlienegri:
> > timing with latest code for 11 seasons for 1 model is ~ 1.5h
> > This is better than the 115min you had before, but also the concno2 numbers are reduced from 9min to 6min. Unless you have made considerable code-changes (last changes were more about readability), I would rather say the performance improvement come from a faster node?

yes indeed, so far we have just moved things around, performance has not been improved (was actually made worse before https://github.com/metno/pyaerocom/pull/1774/commits/164e8f3bbe80517337c5b07b5008cac93f12330a )

charlienegri:
https://aeroval-test.met.no/charlien/pages/maps/?project=cams2-83&experiment=forecast-SON2025&parameter=conco3mda8

note for cams2_83: menu.json is now written compatibly with a non-only-map experiment.. it will need to be rsynced from the only-map experiments too on the top of the contour folder

charlienegri:
@heikoklein if no objections I will merge this, we can consider performance improvement if there is a need at some point..

PR #1792, opened 2026-02-27
Adapt cams2_83 reader to mixed vars files

charlienegri:
## Change Summary

This extends the cams2_83 reader so that having on disk files that contain the extra variables will not blow up the memory, even though the variables are, at this stage, not read in. 
The extra variables are kept instead of being dropped in preparation for having to actually use them... 

## Checklist

* [x] Start with a draft-PR
* [x] Tests pass locally
* [x] Tests pass on CI
* [x] Make PR ready to review


jgriesfeller (approved):
looks good

PR #1829, opened 2026-07-17
Only try somo30/mda8 on hourly data

heikoklein:
## Change Summary

When no hourly data is given, but O3 data available, SOMO and MDA8 results cannot be calculated. This raises currently an exception, but just is expected and should be ignored.

In addition, colocated objects are now saved using netcdf-compress-level 1.

## Checklist

* [x] Start with a draft-PR
* [x] Tests pass locally
* [x] Tests pass on CI
* [x] Make PR ready to review

### DIFF
--- a/pyaerocom/colocation/colocated_data.py
+++ b/pyaerocom/colocation/colocated_data.py
@@ -1230,12 +1230,13 @@ class ColocatedData(BaseModel):
     @staticmethod
     def _build_netcdf_encoding(ds: xr.Dataset) -> dict:
         encoding = {}
-        for name, da in ds.data_vars.items():
+        nc_vars = list(ds.data_vars.keys()) + list(ds.coords.keys())
+        for nc_var in nc_vars:
             var_encoding = {
                 "zlib": True,
                 "complevel": 1,
             }
-            encoding[name] = var_encoding
+            encoding[nc_var] = var_encoding
         return encoding
 
     def to_netcdf(self, out_dir, savename=None, **kwargs):
--- a/pyaerocom/colocation/colocated_data.py
+++ b/pyaerocom/colocation/colocated_data.py
@@ -1227,6 +1227,17 @@ class ColocatedData(BaseModel):
                 meta_out[key] = val
         return meta_out
 
+    @staticmethod
+    def _build_netcdf_encoding(ds: xr.Dataset) -> dict:
+        encoding = {}
+        for name, da in ds.data_vars.items():
+            var_encoding = {
+                "zlib": True,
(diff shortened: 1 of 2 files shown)


pyaerocom/colocation/colocated_data.py
```diff
@@ -1256,8 +1267,11 @@ def to_netcdf(self, out_dir, savename=None, **kwargs):
             savename = f"{savename}.nc"
         arr = self.data.copy()
         arr.attrs = self._prepare_meta_to_netcdf()
+        ds = arr.to_dataset()
+        if "encoding" not in kwargs:
+            kwargs["encoding"] = self._build_netcdf_encoding(ds)
```
charlienegri, line 1273: why is it necessary to switch to dataset and loop over variables to build the dictionary?  
wouldn't a `arr.encoding = {"zlib": True,"complevel": 1}` suffice?

charlienegri (approved):

pyaerocom/colocation/colocated_data.py
heikoklein: Not according to the documentation: https://docs.xarray.dev/en/v0.10.6/generated/xarray.DataArray.to_netcdf.html xarray.DataArray.to_netcdf internally just does a arr.to_dataset().to_netcdf(kwargs), and xArray.Datasets need encoding of form: Nested dictionary with variable names as keys and dictionaries of variable specific encodings as values.

(I tried with something simple as you wrote, but ended up with the _build_netcdf_encoding. If you know of a working simpler way, I would like that.)

charlienegri: I see, if the conversion happens anyway at the to_netcdf call then no point in changing anything... 
given that it's supposedly a mono-variable dataset tho, to construct the encoding dictionary the array alone would be enough without looping 
```
        if "encoding" not in kwargs:
            kwargs["encoding"] = {arr.name: {"zlib": True,"complevel": 1}}
        fp = os.path.join(out_dir, savename)
        arr.to_netcdf(path=fp, **kwargs)
        return fp
```

heikoklein: The dimensions/coordinates should also be compressed.

charlienegri: bu they are not in ds.data_vars.items() either

heikoklein: You are right, I missed them out. I will fix. The coordinates for collocated data are bigger (at least for yearly files) than the data.

PR #1831, opened 2026-08-05
add dummy variables to irregular multi-files

heikoklein:
## Change Summary

emep model-files for trends/multiple year can contain additional variable which are not really needed, but xarray.open_mfdataset was struggling with the complete file. This PR adds dummy variables which are not in the first file, or removes additional files not found in the first dataset.

## Checklist

* [x] Start with a draft-PR
* [ ] Tests pass locally
* [x] Tests pass on CI
* [x] Make PR ready to review

### DIFF
--- a/pyaerocom/io/mscw_ctm/reader.py
+++ b/pyaerocom/io/mscw_ctm/reader.py
@@ -657,7 +657,7 @@ class ReadMscwCtm(GriddedReader):
             chunks={"time": 24},
             decode_timedelta=True,
             combine="by_coords",
-            join="exact",
+            join="outer",
         )
 
         self._private.filedata = ds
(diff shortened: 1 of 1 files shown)


charlienegri (approved):

charlienegri:
is there a risk that one might be very unlucky with the first file and have important variables left out?

heikoklein:
> is there a risk that one might be very unlucky with the first file and have important variables left out?

It is, indeed. But at least it is logged now. In the original version, files without the exact identical variables as the first one were left out.

One could read twice, and in the first round just generate a total number of variables. But that is for another time...

michaelgau:
The risk that important variables are left out in some of the trend years but not in others is very low and can only happen if the person running the trend runs is not aware of it. At least I have always rerun trend years when important variables had been left out and I was made aware of it.
In any case, I will write in the README file for trends that one should make sure that all output files have exactly the same output (even if bugs are found while the trends are running - in that case: rerun!).

PR #1840, opened 2026-08-26
cams282 tweaks to accomodate for hourly data 

charlienegri:
## Change Summary

Hourly data is requested. We get it for the surface fields only (too costly to get it for ml fields and the evaluation that uses vertical levels is only daily, not hourly). All files are read together by the reader in the same dataset tho, so open_mfdataset needs to do more work.

## Checklist

* [x] Start with a draft-PR
* [x] Tests pass locally
* [x] Tests pass on CI
* [x] Make PR ready to review

### DIFF
--- a/pyaerocom/io/cams2_82/reader.py
+++ b/pyaerocom/io/cams2_82/reader.py
@@ -287,7 +287,7 @@ def read_dataset(paths: list[Path]) -> xr.Dataset:
         return ds.pipe(only_first_day).pipe(drop_vars).pipe(interpolate_hourly)
 
     ds = xr.open_mfdataset(
-        paths, preprocess=preprocess, parallel=False, join="outer", chunks={"level": 10, "time": 1}
+        paths, preprocess=preprocess, parallel=False, join="outer", chunks={"level": 10, "time": 2}
     )
     return ds.pipe(fix_missing_vars).pipe(convert_units).pipe(fix_names)
 
--- a/pyaerocom/io/cams2_82/reader.py
+++ b/pyaerocom/io/cams2_82/reader.py
@@ -282,7 +282,7 @@ def read_dataset(paths: list[Path]) -> xr.Dataset:
         return ds.pipe(only_first_day).pipe(drop_vars)
 
     ds = xr.open_mfdataset(
-        paths, preprocess=preprocess, parallel=False, chunks={"level": 10, "time": 1}
+        paths, preprocess=preprocess, parallel=False, join="outer", chunks={"level": 10, "time": 1}
     )
     return ds.pipe(fix_missing_vars).pipe(convert_units).pipe(fix_names)
 
--- a/pyaerocom/io/cams2_82/reader.py
+++ b/pyaerocom/io/cams2_82/reader.py
@@ -276,7 +276,8 @@ def only_first_day(ds: xr.Dataset) -> xr.Dataset:
 
 
 def interpolate_hourly(ds: xr.Dataset) -> xr.Dataset:
-    ds = ds.resample(time="1h").interpolate()
+    if len(ds.time) != 24:
+        ds = ds.resample(time="1h").interpolate(kind="nearest")
     return ds
 
 
--- a/pyaerocom/io/cams2_82/reader.py
+++ b/pyaerocom/io/cams2_82/reader.py
@@ -275,11 +275,16 @@ def only_first_day(ds: xr.Dataset) -> xr.Dataset:
     return ds.sel(time=ds.time.dt.day == first_day)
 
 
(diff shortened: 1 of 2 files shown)


pyaerocom/scripts/cams2_82/config.py
```diff
@@ -78,7 +78,7 @@
     min_num_obs=dict(
         # yearly=dict(monthly=9),
         # monthly=dict(daily=21, weekly=3), # not used
-        daily=dict(hourly=6), # 3-hourly data
+        daily=dict(hourly=18), # 3-hourly data
```
heikoklein:
```suggestion
        daily=dict(hourly=18), # 1-hourly data
```

pyaerocom/io/cams2_82/reader.py
```diff
@@ -275,14 +275,19 @@ def only_first_day(ds: xr.Dataset) -> xr.Dataset:
...
 def read_dataset(paths: list[Path]) -> xr.Dataset:
     paths = check_files(paths)
 
     def preprocess(ds: xr.Dataset) -> xr.Dataset:
-        return ds.pipe(only_first_day).pipe(drop_vars)
+        return ds.pipe(only_first_day).pipe(drop_vars).pipe(interpolate_hourly)
 
     ds = xr.open_mfdataset(
-        paths, preprocess=preprocess, parallel=False, chunks={"level": 10, "time": 1}
+        paths, preprocess=preprocess, parallel=False, join="outer", chunks={"level": 10, "time": 1}
```
heikoklein: I'm not really sure how xarray works under the hood. For the time-interpolation, at least two slices are needed, so you might end up reading the whole dataset twice. Maybe this is cached within netcdf-cache if it fits, but I think you should specify a full day, `"time": 24` without consuming too much memory reducing the slices which need re-reads?
If you have tested the performance then disregard my comment here.

heikoklein (approved):

pyaerocom/io/cams2_82/reader.py
charlienegri: the only thing that I tested is that without adding the resampling in preprocess the memory does blow up a bit now that we have different resolutions for different variables and the reader aggregates that in a single dataset... I do not think that tweaking chunking alone can spare us from having to do this   
the performance is ok with this preprocessing added, but I think open_mfdataset actually just re-chunks in the way you suggest, under the carpet, I will test with "time": 24 explicitly and change to that

charlienegri: so with chunking  time 1 it logs about re-chunking for performance reason but it works,  
with time 24 and the same memory I used for the previous test, it does not complain about chunking but I get a core dumped 🤔
```
numpy._core._exceptions._ArrayMemoryError: Unable to allocate 681. MiB for an array with shape (22, 4059000) and data type float64
 ❌ 

  0%|                                                                                                                                                                | 0/273 [00:00<?, ?it/s]NCbytes failure
Aborted (core dumped)
```
seems like 24 is wrong for chunking but the timesteps are 24 indeed

charlienegri: I repeated a run with time 1 chunking from command line to capture the warning: 
```
dask/array/core.py:5201: PerformanceWarning: Increasing number of chunks by factor of 22
```
this factor does not make sense to me

charlienegri: just to test different cases:  
"time": 24 -> memory error
"time": 23 -> memory error  
"time": 22 -> memory error
the 2nd and 3rd as expected based on the timesteps we have in theory  
no PerformanceWarning in any of these

heikoklein: Maybe time-chunk should be 8 to cover a complete day in case of 3h data, and it only matters there? In fact, this is anyway just a rough speedup to avoid re-reading data, so anything > 1 should be good. 10 is also fine, then only 10% of the data will need to get read twice.

charlienegri: the data should not be 3-hourly anymore with the preprocessing pipe  
I think that having too big time chunks does not fit into memory when dealing with multi-level data,  
  
I still cannot make sense of the numbers given in the performance warnings btw 
with "time": 2 --> `PerformanceWarning: Increasing number of chunks by factor of 15` 🤷‍♂️  

~I see no difference in speed, whatever xarray is doing under the carpet, it seems to be doing it no matter what chunking size, like 1 or 2 does not make a difference~

charlienegri: I am not sure anymore of the latter.. the run I compared was doing less work because of a (different) obs filtering issue, "time": 4 already uses too much memory, I will settle for "time": 2

heikoklein (approved):

